PR #3, opened 2016-06-01
Fixed a null pointer exception in Pinger code

### LINKED ISSUE #2: Cross-extension communication is not torn down correctly when pinging times out
When the message channel pinger detects a timeout and decides to dispose the message channel, the disposal process is interrupted with JavaScript SyntaxError.

As the result, the message channel may be left opened afterwards (though all requests coming through it will fail).

ivansandrk:
dispose() was called on the pinger before calling messageChannel_.dispose(),
causing the messageChannel_ to be set to null

this commit fixes #2


emaxx-google:
Nice work, thanks!

PR #4, opened 2016-06-01
Dispose message channel when pinger detects a failure

emaxx-google:
When the message channel pinger detects a failure (like the wrong format of the "pong" message or the change of remote channel id), the message channel should be closed and disposed.

The previous code didn't do this really.


ivansandrk:
lgtm

PR #11, opened 2016-06-29
Added optional non-fatal default service callback for message channels

### LINKED ISSUE #10: Sending a message with unexpected type raises a fatal error in the Connector App
Currently, a fatal failure happens when a message with unhandled type is received in the Connector App.

Given that the fatal failure makes the whole Connector App to "crash" and restart when in Release mode, this is not an acceptable behavior.

Instead, a warning/error message should be emitted, and the message channel to the bad client should be destroyed.

/cc @dskaram

ivansandrk:
Fixes #10

### DIFF
--- a/common/js/src/messaging/message-channel-common.js
+++ b/common/js/src/messaging/message-channel-common.js
@@ -1,0 +1,46 @@
+/** @license
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview This file containts common functionality for the provided
+ * message channels.
+ */
+
+goog.provide('GoogleSmartCard.MessageChannelCommon');
+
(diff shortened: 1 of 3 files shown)


common/js/src/messaging/message-channel-common.js
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview This file containts common functionality for the provided
+ * message channels.
+ */
+
+goog.provide('GoogleSmartCard.MessageChannelCommon');
```
emaxx-google: Maybe name the file just "common.js", and the namespace "GoogleSmartCard.MessagingCommon"?

common/js/src/messaging/message-channel-common.js
```diff
@@ -0,0 +1,46 @@
...
+};
+
+function nonFatalDefaultServiceCallback(channel, serviceName, payload) {
+  channel.logger.warning(
+      'Unhandled message received: serviceName="' + serviceName +
+      '", payload=' + GSC.DebugDump.debugDump(payload));
+  channel.dispose();
+}
+
+})  // goog.scope
```
emaxx-google: nit: add ";"

common/js/src/messaging/message-channel-common.js
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview This file containts common functionality for the provided
```
emaxx-google: nit: s/containts/contains/

emaxx-google:
lgtm, except probably think about simpler naming

PR #12, opened 2016-06-29
Fix Connector App fatal failure on trying to parse an invalid message

emaxx-google:
Instead of failing (and reloading the whole app) when the message received from a client contains invalid data, the message channel should be disposed.


ivansandrk:
lgtm

PR #15, opened 2016-07-04
Introduce JavaScript unit tests

emaxx-google:
Ivan, PTAL.

This change contains some build tools changes and, as the example, the first unit test for the "common/js" library.

/cc @ivansandrk


common/js/src/fixed-size-integer-unittest.js
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+goog.require('GoogleSmartCard.FixedSizeInteger');
+goog.require('GoogleSmartCard.Logging');
+goog.require('goog.testing.jsunit');
+
+goog.setTestOnly();
```
emaxx-google: BTW, this gives at least some protection against accidental inclusion of test sources into a normal build.

common/js/src/fixed-size-integer-unittest.js
```diff
@@ -0,0 +1,41 @@
...
+var castToInt32 = GSC.FixedSizeInteger.castToInt32;
+
+goog.exportSymbol('testCastToInt32', function() {
+  assertEquals(0, castToInt32(0));
+  assertEquals(1, castToInt32(1));
+  assertEquals(-1, castToInt32(-1));
+  assertEquals(-1, castToInt32(0xFFFFFFFF));
+  assertEquals(0x7FFFFFFF, castToInt32(0x7FFFFFFF));
+  assertEquals(-0x80000000, castToInt32(0x80000000));
+  assertEquals(-0x7FFFFFFF, castToInt32(0x80000001));
```
ivansandrk: I'd add the following (if not just for kicks):

assertEquals(-castToInt32(0x80000000), castToInt32(0x80000000));

ivansandrk:
lgtm

common/js/src/fixed-size-integer-unittest.js
emaxx-google: What about this?
assertEquals(-0x80000000, castToInt32(-0x80000000))
(If I understood your idea correctly)

PR #18, opened 2016-07-18
Fix possible crash in pcsc-lite-related debug code

emaxx-google:
This fixes the potential memory access violation that could occur in the debug printing code in the PC/SC-Lite client NaCl port.

The crash could happen when the SCARD_AUTOALLOCATE feature is used and the debug tracing of the executed PC/SC-Lite functions was enabled. The buggy code would try to dump the buffer contents from the address storing the buffer address (instead of dereferencing the pointer and then dumping the bytes).

BTW, this code path is used only on the client side, not on the server side - so there were no security implications for the Connector app.

The bug was discovered because of a hint from clang-tidy (the local variable containing the correct pointer was assigned, but never used).


ivansandrk:
lgtm

PR #19, opened 2016-07-19
Fixed Chromevox accessibility issues

ivansandrk:
* Switch translation logic from CSS::after to custom javascript
  (using chrome.i18n.getMessage(textId))

### DIFF
--- a/common/js/src/i18n.js
+++ b/common/js/src/i18n.js
@@ -19,7 +19,6 @@
  * internationalization.
  */
 
-
 goog.provide('GoogleSmartCard.I18n');
 
 goog.require('GoogleSmartCard.Logging');
@@ -34,7 +33,7 @@ var GSC = GoogleSmartCard;
  * @type {!goog.log.Logger}
  * @const
  */
-var logger = GSC.Logging.getScopedLogger('GoogleSmartCard.I18n');
+var logger = GSC.Logging.getScopedLogger('I18n');
 
 /** @const */
 var I18N_DATA_ATTRIBUTE = 'data-i18n';
@@ -46,22 +45,20 @@ var I18N_DATA_ARIA_LABEL_ATTRIBUTE = 'data-i18n-aria-label';
  * @param {!Element} element
  * @param {string} attribute
  * @param {function(!Element,string)} transformFunction
- * @private
  */
 function transformElement(element, attribute, transformFunction) {
   var i18nId = element.getAttribute(attribute);
-  if (!i18nId) {
-    GSC.Logging.failWithLogger(
-        logger,
-        'Failed to get element attribute: ' + element.outerHTML);
-  }
+  GSC.Logging.checkWithLogger(
+      logger,
+      i18nId,
+      'Failed to get attribute [' + attribute +
+      '] for element: ' + element.outerHTML);
 
   var translatedText = chrome.i18n.getMessage(i18nId);
-  if (!translatedText) {
-    GSC.Logging.failWithLogger(
-        logger,
-        'Failed to get translation for text with id: ' + i18nId);
-  }
+  GSC.Logging.checkWithLogger(
+      logger,
+      translatedText,
+      'Failed to get translation for text with id: ' + i18nId);
 
   transformFunction(element, translatedText);
 }
@@ -69,7 +66,6 @@ function transformElement(element, attribute, transformFunction) {
 /**
  * @param {string} attribute
  * @param {function(!Element,string)} transformFunction
- * @private
  */
 function transformAllElements(attribute, transformFunction) {
   var selector = '[' + attribute + ']';
@@ -88,13 +84,17 @@ function transformAllElements(attribute, transformFunction) {
 GSC.I18n.adjustElementsTranslation = function() {
   transformAllElements(
       I18N_DATA_ATTRIBUTE, function(element, translatedText) {
+    logger.fine('Translating element.textContent [' + element.outerHTML +
+                '] to "' + translatedText + '"');
     element.textContent = translatedText;
   });
 
   transformAllElements(
       I18N_DATA_ARIA_LABEL_ATTRIBUTE, function(element, translatedText) {
+    logger.fine('Translating element.aria-label [' + element.outerHTML +
+                '] to "' + translatedText + '"');
     element.setAttribute('aria-label', translatedText);
   });
-}
+};
 
 });  // goog.scope
--- a/common/js/src/i18n.js
+++ b/common/js/src/i18n.js
@@ -1,0 +1,100 @@
+/** @license
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview This file contains common functionality for
+ * internationalization.
+ */
+
+
+goog.provide('GoogleSmartCard.I18n');
+
+goog.require('GoogleSmartCard.Logging');
+goog.require('goog.log.Logger');
+
+goog.scope(function() {
+
+/** @const */
+var GSC = GoogleSmartCard;
+
+/**
+ * @type {!goog.log.Logger}
+ * @const
+ */
+var logger = GSC.Logging.getScopedLogger('GoogleSmartCard.I18n');
+
+/** @const */
+var I18N_DATA_ATTRIBUTE = 'data-i18n';
+
+/** @const */
+var I18N_DATA_ARIA_LABEL_ATTRIBUTE = 'data-i18n-aria-label';
+
+/**
+ * @param {!Element} element
+ * @param {string} attribute
+ * @param {function(!Element,string)} transformFunction
+ * @private
+ */
+function transformElement(element, attribute, transformFunction) {
+  var i18nId = element.getAttribute(attribute);
+  if (!i18nId) {
+    GSC.Logging.failWithLogger(
+        logger,
+        'Failed to get element attribute: ' + element.outerHTML);
+  }
+
+  var translatedText = chrome.i18n.getMessage(i18nId);
+  if (!translatedText) {
+    GSC.Logging.failWithLogger(
+        logger,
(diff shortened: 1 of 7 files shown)


smart_card_connector_app/src/_locales/en/messages.json
```diff
@@ -19,6 +19,10 @@
     "message": "Smart Card Connector",
     "description": "The header displayed in the main window"
   },
+  "mainWindowClosebutton": {
```
emaxx-google: nit: s/Closebutton/CloseButton/?

smart_card_connector_app/src/window-logs-exporting.js
```diff
@@ -66,8 +69,8 @@ function exportLogsClickListener(e) {
   if (!isExportLogsAvailable)
     return;
 
-  goog.dom.classlist.add(
-      exportLogsElement, EXPORT_LOGS_ELEMENT_EXPORTING_CLASS);
+  exportLogsElement.innerText =
```
emaxx-google: "innerText" is a non-standard feature, basically - IE legacy.
Better to use "textContent" or goog.dom.setTextContent.

smart_card_connector_app/src/window-logs-exporting.js
```diff
@@ -81,20 +84,20 @@ function exportLogs() {
       ' log messages from the log buffer');
   var copyingSuccess = GSC.Clipboard.copyToClipboard(dumpedLogs);
 
-  goog.dom.classlist.remove(
-      exportLogsElement, EXPORT_LOGS_ELEMENT_EXPORTING_CLASS);
+  exportLogsElement.innerText =
```
emaxx-google: nit: Code would look a bit cleaner if this would be put inside "else" branch of the condition below.

smart_card_connector_app/src/window-main.js
```diff
@@ -58,7 +58,34 @@ GSC.ConnectorApp.Window.AppsDisplaying.initialize();
 GSC.ConnectorApp.Window.DevicesDisplaying.initialize();
 GSC.ConnectorApp.Window.LogsExporting.initialize();
 
-GSC.PopupWindow.Client.setExplicitWindowTitle();
 GSC.PopupWindow.Client.showWindow();
 
+/** @const */
```
emaxx-google: I think it makes sense to put this into common/, so that the same functionality could be easily used within other windows.
E.g. GoogleSmartCard.I18n.ProcessAllElementI18nAttributes (that's just an example, maybe you come up with a better name).

smart_card_connector_app/src/window-main.js
```diff
@@ -58,7 +58,34 @@ GSC.ConnectorApp.Window.AppsDisplaying.initialize();
...
+var DATA_I18N_STR = 'data-i18n';
```
emaxx-google: "_STR" doesn't give much information here, and it's not clear that this constant specifies attribute name.
Maybe "I18_DATA_ATTRIBUTE"?

smart_card_connector_app/src/window-main.js
```diff
@@ -58,7 +58,34 @@ GSC.ConnectorApp.Window.AppsDisplaying.initialize();
...
+
+/** @const */
+var DATA_I18N_ARIA_LABEL_STR = 'data-i18n-aria-label';
+
+function setElementI18nText(element) {
+  var i18nId = element.getAttribute(DATA_I18N_STR);
+  var translatedText = chrome.i18n.getMessage(i18nId);
```
emaxx-google: Please add assertions that the returned value is not undefined.
It's a bad error if somebody mistypes the message name; and the consequence is that the "undefined" string would be rendered, which would confuse the user.

smart_card_connector_app/src/window-main.js
```diff
@@ -58,7 +58,34 @@ GSC.ConnectorApp.Window.AppsDisplaying.initialize();
...
+  element.innerText = translatedText;
+}
+
+function setElementI18nAriaLabel(element) {
+  var i18nId = element.getAttribute(DATA_I18N_ARIA_LABEL_STR);
+  var translatedText = chrome.i18n.getMessage(i18nId);
+  element.setAttribute('aria-label', translatedText);
+}
+
+for (let element of /** @type {?} */ (goog.dom.getElementsByTagNameAndClass(
```
emaxx-google: Have you found anywhere in the docs that goog.dom.getElementsByTagNameAndClass may accept such selectors. I think it's just a coincidence that it works that way (actually, it contains several branches, and it was just a luck that the picked branch uses Element.querySelectorAll instead of any other way).

So, instead of relying on this behavior (which may change at any time), some cleaner way should be used - e.g. using Document.querySelectorAll.

smart_card_connector_app/src/window.css
```diff
@@ -120,8 +112,8 @@ body {
   margin: 12px;
 }
 
-#add-device::after {
-  content: '__MSG_addDevice__';
+#app-list-header {
+  margin: 0px 0 0;
```
emaxx-google: Why did this become necessary?
I think the whole layout was already approved with the UI team.

(And, if this line is actually needed, then please change it to the shortest format: "margin: 0".)

smart_card_connector_app/src/window.html
```diff
@@ -16,27 +16,28 @@
...
   <link rel="stylesheet" href="window.css">
   <meta charset="UTF-8">
 </head>
 <body>
-  <h1 id="header"></h1>
+  <div id="navbar"></div>
+  <h1 id="header" data-i18n="mainWindowHeader"></h1>
 
-  <button id="close-window">×</button>
+  <button id="close-window" data-i18n-aria-label="mainWindowClosebutton">×</button>
```
emaxx-google: I'd suggest removing the word "main" here. The close button would probably behave in the same way regardless of the window it's in.

smart_card_connector_app/src/window.css
```diff
@@ -26,26 +26,22 @@ button {
...
-  content: '__MSG_mainWindowTitle__';
-}
-
 body {
   margin: 0;
   padding: 0 0 0 16px;
 }
 
-#header {
+#navbar {
```
emaxx-google: OK, but please make sure that neither the header nor the close button moved after this change.

smart_card_connector_app/src/window.html
ivansandrk: Renamed to "windowClose".

smart_card_connector_app/src/window.css
ivansandrk: Right, removed it as it doesn't change anything.

smart_card_connector_app/src/window.css
ivansandrk: Took screenshots and compared them - they are equal down to the pixel.

common/js/src/i18n.js
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * @param {!Element} element
+ * @param {string} attribute
+ * @param {function(!Element,string)} transformFunction
+ * @private
+ */
+function transformElement(element, attribute, transformFunction) {
+  var i18nId = element.getAttribute(attribute);
+  if (!i18nId) {
+    GSC.Logging.failWithLogger(
```
emaxx-google: Better to use the shorter and more explicit "checkWithLogger" function.

common/js/src/i18n.js
```diff
@@ -0,0 +1,100 @@
...
+        logger,
+        'Failed to get element attribute: ' + element.outerHTML);
```
emaxx-google: If you decided to put a meaningful error message, then it makes sense to print the requested attribute name too.

common/js/src/i18n.js
```diff
@@ -0,0 +1,100 @@
...
+var I18N_DATA_ATTRIBUTE = 'data-i18n';
+
+/** @const */
+var I18N_DATA_ARIA_LABEL_ATTRIBUTE = 'data-i18n-aria-label';
+
+/**
+ * @param {!Element} element
+ * @param {string} attribute
+ * @param {function(!Element,string)} transformFunction
+ * @private
```
emaxx-google: No need to do this, all functions are private by default, unless exported explicitly.

common/js/src/i18n.js
```diff
@@ -0,0 +1,100 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview This file contains common functionality for
+ * internationalization.
+ */
+
+
```
emaxx-google: nit: Please remove the extra blank line.

common/js/src/i18n.js
```diff
@@ -0,0 +1,100 @@
...
+goog.scope(function() {
+
+/** @const */
+var GSC = GoogleSmartCard;
+
+/**
+ * @type {!goog.log.Logger}
+ * @const
+ */
+var logger = GSC.Logging.getScopedLogger('GoogleSmartCard.I18n');
```
emaxx-google: There should be no "GoogleSmartCard." prefix in the passed string, because it is appended by getScopedLogger automatically.

common/js/src/i18n.js
```diff
@@ -0,0 +1,100 @@
...
+ * Goes through all the HTML elements in the current window, replacing
+ * element.textContent with translation for elements containing attribute
+ * I18N_DATA_ATTRIBUTE, and setting attribute aria-label to translation for
+ * elements containing attribute I18N_DATA_ARIA_LABEL_ATTRIBUTE. Used for
+ * translation that plays nice with Chromevox (accessibility tool).
+ */
+GSC.I18n.adjustElementsTranslation = function() {
+  transformAllElements(
+      I18N_DATA_ATTRIBUTE, function(element, translatedText) {
+    element.textContent = translatedText;
```
emaxx-google: (this is optional) Could be useful to emit log messages describing each transformation with the DOM (probably, at FINE level, so that by default they go to the console only in Debug mode).

common/js/src/i18n.js
```diff
@@ -0,0 +1,100 @@
...
+  });
+
+  transformAllElements(
+      I18N_DATA_ARIA_LABEL_ATTRIBUTE, function(element, translatedText) {
+    element.setAttribute('aria-label', translatedText);
+  });
+}
```
emaxx-google: nit: add ";"

smart_card_connector_app/src/window-main.js
```diff
@@ -58,7 +59,9 @@ GSC.ConnectorApp.Window.AppsDisplaying.initialize();
 GSC.ConnectorApp.Window.DevicesDisplaying.initialize();
 GSC.ConnectorApp.Window.LogsExporting.initialize();
 
-GSC.PopupWindow.Client.setExplicitWindowTitle();
 GSC.PopupWindow.Client.showWindow();
 
+// This call does all of the translation work.
```
emaxx-google: This comment doesn't provide much useful information. I guess, the function name and its JSDoc already provide with all the necessary context.

emaxx-google:
lgtm

PR #21, opened 2016-07-20
Fix PC/SC-Lite client handling error when the server is destroyed

emaxx-google:
This fixes JS fatal errors that could happen when a client tried to
connect to the Connector app after the server message channel was destroyed
(e.g. because of the NaCl module crash).

The problem was that the ClientHandler class constructor could be called
with an undefined value instead of the server message channel (where the
undefined value originates from accessing a property of already disposed
NaclModule object).


emaxx-google:
Ivan, PTAL.

ivansandrk:
lgtm

PR #22, opened 2016-07-20
Emit a fatal error when the PC/SC-Lite server NaCl module destroys

emaxx-google:
This fixes the behavior when the Connector app continued to work even
after the NaCl module was disposed of (which could result in various errors
and, generally, makes no sense).


emaxx-google:
Ivan, PTAL.

ivansandrk:
lgtm

PR #23, opened 2016-07-20
Clean up logic in the fatal error emitting code

emaxx-google:
Make the code more straightforward and also handle the (potential) case
when the goog.asserts.ENABLE_ASSERTS constant is changed from its default
value.

### DIFF
--- a/common/js/src/logging/logging.js
+++ b/common/js/src/logging/logging.js
@@ -155,8 +155,7 @@ GSC.Logging.getChildLogger = function(
  * Checks if the condition evaluates to true.
  *
  * In contrast to goog.asserts.assert method, this method works in non-Debug
- * builds too (provided that the goog.asserts.ENABLE_ASSERTS definition is
- * true).
+ * builds too.
  * @template T
  * @param {T} condition The condition to check.
  * @param {string=} opt_message Error message in case of failure.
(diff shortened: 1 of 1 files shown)


emaxx-google:
Ivan, PTAL.

common/js/src/logging/logging.js
```diff
@@ -195,14 +194,13 @@ GSC.Logging.checkWithLogger = function(
  * @param {...*} var_args The items to substitute into the failure message.
  */
 GSC.Logging.fail = function(opt_message, var_args) {
-  var messageAndArgs = Array.prototype.slice.call(arguments);
   if (goog.DEBUG) {
-    goog.asserts.fail.apply(goog.asserts, messageAndArgs);
+    throwAssertionError(opt_message, var_args);
   } else {
+    var messageAndArgs = Array.prototype.slice.call(arguments);
```
ivansandrk: What does this line do? Ie. I'm unfamiliar with the Array.proto... syntax.

ivansandrk:
lgtm (+ a question from me)

common/js/src/logging/logging.js
emaxx-google: This is a trick to transform arguments variable (which is for some reason not an array) into an array. I believe it's a pretty common way of doing this, e.g. see the sample code in the Array.slice docs in MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice#Array-like_objects

PR #26, opened 2016-08-04
Reader status

ivansandrk:
Hey Max, this is the preliminary version. I'm having some doubts about the design decisions, could use input there (so focus more on the design than the "nits" in the review).

I've decided to create a new class (reader-tracker.js) to track the status of the readers. It's being passed inside the third argument of GSC.PopupWindow.Server.createWindow to the foreground window. I've commented out DebugDump for now because it doesn't play nice with this passed argument (server.js:69 - GSC.DebugDump.debugDump(opt_data) - is causing a stack overflow), I need to figure this out later.

There are some TODO's left in the code with things that need to be done, or questions I have.

Here are some screenshots of how it looks like:

screenshot from 2016-08-04 20 07 59

screenshot from 2016-08-04 20 08 12

screenshot from 2016-08-04 20 08 42

screenshot from 2016-08-04 20 08 49


third_party/pcsc-lite/naclport/server/src/readerfactory_nacl.cc
```diff
@@ -0,0 +1,110 @@
...
+// NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+// THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+// This file contains a replacement function for the original readerfactory.c
+// PC/SC-Lite internal implementation.
+
+// TODO: These includes need to come before readerfactory.h otherwise compiler
```
emaxx-google: I think that the readerfactory.h header from PC/SC-Lite is written a bit inaccurately, so that it actually requires including some other PC/SC-Lite headers before it.
Or you could also have problems if you tried to include C++ headers (like "ppapi/...") from inside "extern "C"", which won't work.

You can look up a working example here: https://github.com/GoogleChrome/chromeos_smart_card_connector/blob/master/third_party/pcsc-lite/naclport/server/src/google_smart_card_pcsc_lite_server/global.cc

third_party/pcsc-lite/naclport/server/src/readerfactory_nacl.cc
```diff
@@ -0,0 +1,110 @@
...
+const char kDeviceMessageKey[] = "device";
+const char kReturnCodeMessageKey[] = "returnCode";
+
+void post_message(const char* type, const pp::VarDictionary& message_data)
+{
+  pp::VarDictionary message;
+  message.Set(kTypeMessageKey, type);
+  message.Set(kDataMessageKey, message_data);
+
+  const pp::Module* const pp_module = pp::Module::Get();
```
emaxx-google: Mark this block as TODO, as this is not a clean solution.
First, this is untidy as it introduces a hidden dependency on the global context.
Second, it's not thread-safe.

third_party/pcsc-lite/naclport/server/src/readerfactory_nacl.cc
```diff
@@ -0,0 +1,110 @@
...
+
+  // TODO: Testing code, to be removed at a later time (failing reader).
+  if (std::string("SCM Microsystems Inc. SCR 3310") == readerNameLong) {
+    post_message(kAddReaderMessageType, VarDictBuilder()
+        .Add(kNameMessageKey, readerNameLong).Add(kPortMessageKey, port)
+        .Add(kDeviceMessageKey, device).Add(kReturnCodeMessageKey, SCARD_E_INVALID_VALUE).Result());
+    return ret;
+  }
+
+  post_message(kAddReaderMessageType, VarDictBuilder()
```
emaxx-google: I would say that it's cleaner to use here a different message type than in the real "add" even that happens in the line 79. Here it's actually an "update" of the reader.

third_party/pcsc-lite/naclport/server/src/readerfactory_nacl.cc
```diff
@@ -0,0 +1,110 @@
...
+
+  return ret;
+}
+
+LONG RFRemoveReader(const char *readerName, int port)
+{
+  LONG ret = RFRemoveReaderServer(readerName, port);
+
+  if (ret == SCARD_S_SUCCESS) {
+    post_message(kRemoveReaderMessageType, VarDictBuilder()
```
emaxx-google: Looks like this branch won't be executed when the _failed_ reader is disconnected. If that's true, then the items standing for the failed readers will be displayed forever, which is not good.

third_party/pcsc-lite/naclport/server/src/readerfactory_nacl.cc
```diff
@@ -0,0 +1,110 @@
...
+      .Add(kDeviceMessageKey, device).Add(kReturnCodeMessageKey, ret).Result());
+
+  return ret;
+}
+
+LONG RFRemoveReader(const char *readerName, int port)
+{
+  LONG ret = RFRemoveReaderServer(readerName, port);
+
+  if (ret == SCARD_S_SUCCESS) {
```
emaxx-google: Looking at how RFRemoveReader function is used (in file hotplug_libusb.c), seems that the reader is removed from the internal lists regardless of the return code. So this condition is probably extra.

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,149 @@
...
+
+  var name = json['readerName'];
+  var port = json['port'];
+  var device = json['device'];
+  var returnCode = json['returnCode'];
+
+  var value = {'name': name};
+
+  if (returnCode === undefined) {
+    value.color = 'yellow';
```
emaxx-google: Closure Compiler may rename properties in such constructs. You would need to access the properties using the quoted strings in order to avoid the renaming.
And also it makes sense to introduce a small class that would define which exactly fields does this "value" have.

common/js/src/logging/debug-dump.js
```diff
@@ -179,27 +179,28 @@ function dumpObject(value) {
  * @return {string}
  */
 GSC.DebugDump.dump = function(value) {
-  if (!goog.isDef(value))
```
emaxx-google: I think the problem is that GSC.PopupWindow.Server.createWindow makes the debug dump of all passed arguments, which really is a bad thing once we start passing the complex arguments (like pointing to DOM elements, which contain circular references) to it.
So just removing these debug dumps from createWindow should be enough to fix the issue.

emaxx-google:
I think the overall approach is good.
But, regarding the NaCl side, my opinion still is that finally we need to organize it in a cleaner fashion: with a new singleton class owned by the PpInstance class.
It won't be hard to do, but will allow to structure the code much better. And you can even take a look at the existing similar pieces, e.g.: https://github.com/GoogleChrome/chromeos_smart_card_connector/blob/master/third_party/pcsc-lite/naclport/cpp_client/src/google_smart_card_pcsc_lite_client/global.h

third_party/pcsc-lite/naclport/server/src/readerfactory_nacl.cc
ivansandrk: Did some more code digging, made some tests, removed the condition completely.

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * TODO: I need to figure out the param because the closure compiler complains
+ *       about mismatched types, will do it at a later time.
+ * param {Object|string} message
+ * @private
+ */
+ReaderTracker.prototype.readerAddedListener_ = function(message) {
+  this.logger_.warning('readerAddedListener_ ' + message);
+
+  // TODO: needs a promise? (check json.js)
```
ivansandrk: > this.logger_.warning(typeof message);
----> [  1.627s] [GoogleSmartCard.NaclModule<"nacl_module.nmf">] string

For some reason I'm receiving a string instead of an object, maybe I did something wrong, I'm not sure at this point. Anyway, it works with the json parsing, doesn't without it.

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -20,6 +20,7 @@
  * NaCl (check readerfactory_nacl.cc for more information).
  */
 
+goog.provide('GoogleSmartCard.ReaderInfo')
```
ivansandrk: How do I make ReaderInfo structure local? (closure compiler forces me to put up this provide line)

ivansandrk:
I realised that RFRemoveReader can be called in the middle of RFAddReader, and I'm not sure what to do in that case. I need to think about it more. Option 1 would be to remove the reader completely (and log the actual error you get when RFAddReader finishes). Option 2 would be to keep the reader in the list together with the error indicator and the error message, but then I have to figure out a way of removing the reader from the list when it actually gets disconnected from the computer. Maybe keep listening to libusb and correlate somehow the device indicators.

emaxx-google:
> I realised that RFRemoveReader can be called in the middle of RFAddReader, and I'm not sure what to do in that case.

Are you sure that this can happen? Track how both of these functions are called. My quick investigation said that both of them are called from a loop inside HPRescanUsbBus, which means that they cannot be called simultaneously.

smart_card_connector_app/src/reader-tracker.js
emaxx-google: The only difference of "public" vs. "local" definitions is that the former ones are assigned to the properties of the global objects like "GoogleSmartCard", while the latter ones are just defined in the file.
Example of a local definition:
`/** @constructor */ function A() {}`

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * Structure used to store information about the reader.
+ * @param {string} name
+ * @param {string=} status
+ * @param {string=} error
+ * @constructor
+ * @struct
+ */
+GSC.ReaderInfo = function(name, status, error) {
+  this.name = name;
```
emaxx-google: You have to assign also properties using strings, because when passing this object into a different compilation unit (into window.js), the compiler may rename the properties in a different way.
E.g., combining both approaches:
`this['name'] = this.name = name;`

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,181 @@
...
+ * @param {!goog.messaging.AbstractChannel} messageChannel
+ * @param {!goog.log.Logger} parentLogger
+ * @constructor
+ */
+GSC.ReaderTracker = function(messageChannel, parentLogger) {
+  /** @private */
+  this.logger_ = parentLogger;
+
+  messageChannel.registerService(
+      'reader_init_add', this.readerInitAddListener_.bind(this));
```
emaxx-google: You need to pass "true" as the third argument in order to receive the data as objects. Without it, the default behavior is to JSON-encode all data, which makes no sense in our case (and yes, I agree that it's a weird API provided by the Closure Library).

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,181 @@
...
+var ReaderTracker = GSC.ReaderTracker;
+
+/**
+ * TODO: I need to figure out the param because the closure compiler complains
+ *       about mismatched types, will do it at a later time.
+ * param {Object|string} message
+ * @private
+ */
+ReaderTracker.prototype.readerInitAddListener_ = function(message) {
+  this.logger_.warning('readerInitAddListener_ ' + message);
```
emaxx-google: When working on the final implementation, please make sure to produce meaningful logs with the "info" level. Because that's very useful information for investigating the issues post-factum.

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,181 @@
...
+  var port = json['port'];
+
+  this.readers_.remove(port);
+  this.fireOnUpdateListeners_();
+};
+
+/**
+ * param {function()} listener
+ */
+ReaderTracker.prototype.addOnUpdateListener = function(listener) {
```
emaxx-google: I think you also need to add a way to remove a listener.
Because, first, each time when the app window is opened, a new listener would be added without removing the old one. Second, all old listeners from the closed windows would be still called on each update, which sounds quite weird (won't they crash while trying to update the contents of the already destroyed window?).

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,181 @@
...
+  this.onUpdateListeners_.push(listener);
+};
+
+/**
+ * @private
+ */
+ReaderTracker.prototype.fireOnUpdateListeners_ = function() {
+  this.logger_.fine('Firing readers updated listeners');
+  for (let listener of this.onUpdateListeners_) {
+    listener();
```
emaxx-google: Why not pass the readers data to the listeners?

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -148,9 +176,12 @@ function getUserSelectedDevicesCallback(devices) {
 }
 
 GSC.ConnectorApp.Window.DevicesDisplaying.initialize = function() {
-  loadDeviceList();
```
emaxx-google: I think it's still useful to log the data supplied by the chrome.usb API, in case something in PC/SC or CCID went wrong.

emaxx-google: (But, probably, it would make much more sense to move this logging into the background script - it's not useful when closing the window stops this information from logging.)

smart_card_connector_app/src/reader-tracker.js
ivansandrk: According to the official docs, this should be safe https://github.com/google/closure-compiler/wiki/@struct-and-@dict-Annotations

> @struct is used to create objects that have a fixed number of properties, ie, objects that behave similarly to Java objects.

> By using @struct, you know that the compiler will rename all properties safely, because you can't use bracket access.

ivansandrk:
>Are you sure that this can happen? Track how both of these functions are called. My quick investigation said that both of them are called from a loop inside HPRescanUsbBus, which means that they cannot be called simultaneously.

Yes, RFRemoveReader is called from RFAddReader function. But I've investigated this further and all is fine in the end. Too long to write about it here, I will tell you live.

smart_card_connector_app/src/reader-tracker.js
ivansandrk: Consistency and design questions. I uploaded the alternate version and you see if that one looks better. [Pass the readers data to the listeners in the callback - d6169aa]

smart_card_connector_app/src/window-devices-displaying.js
ivansandrk: Added a TODO for a later time.

smart_card_connector_app/src/reader-tracker.js
emaxx-google: I don't see why it is safe.
Look at this example, where the compiler renames properties "foo" and "bar" into "b" and "a":
https://closure-compiler.appspot.com/home#code%3D%252F%252F%2520%253D%253DClosureCompiler%253D%253D%250A%252F%252F%2520%2540compilation_level%2520ADVANCED_OPTIMIZATIONS%250A%252F%252F%2520%2540output_file_name%2520default.js%250A%252F%252F%2520%2540formatting%2520pretty_print%250A%252F%252F%2520%253D%253D%252FClosureCompiler%253D%253D%250A%250A%252F**%2520%2540struct%2520%2540constructor%2520*%252F%250Afunction%2520A()%2520%257B%250A%2520%2520this.foo%2520%253D%25201%253B%250A%2520%2520this.bar%2520%253D%25202%253B%250A%257D%250A%250Aa.foo%2520%253D%25203%253B%250Aa.bar%2520%253D%25204%253B%250A%250Aconsole.log(a)%253B
There is no guarantee that the compiler will do the renaming in the same way while compiling several files. So when you'll pass this object from one window to another, there's a chance that the properties would be named in a different manner than it was expected in the target script.

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,189 @@
...
+ * This class tracks readers, provides methods to retrieve the list of readers
+ * and get updates on change, and listens to messages received from the hook
+ * living in the NaCl space.
+ * @param {!goog.messaging.AbstractChannel} messageChannel
+ * @param {!goog.log.Logger} parentLogger
+ * @constructor
+ */
+GSC.ReaderTracker = function(messageChannel, parentLogger) {
+  /** @private */
+  this.logger_ = parentLogger;
```
emaxx-google: You should get child logger here instead of just logging the messages into the parent's logger.

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,189 @@
...
+
+/** @const */
+var ReaderTracker = GSC.ReaderTracker;
+
+/**
+ * @param {!Object|string} message
+ * @private
+ */
+ReaderTracker.prototype.readerInitAddListener_ = function(message) {
+  var name = message['readerName'];
```
emaxx-google: It's better to specify explicit types here in order to help the compiler with doing appropriate checks in the code that follows this.
For example, currently the "port" variable is actually a number, while you have defined this.readers_ to have string keys - and the compiler wasn't able to find out this inconsistency.

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,189 @@
...
+
+goog.scope(function() {
+
+/** @const */
+var GSC = GoogleSmartCard;
+
+/**
+ * Structure used to store information about the reader.
+ * @param {string} name
+ * @param {string} status
```
emaxx-google: Looks like the enum type would be more appropriate here.

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,189 @@
...
+/**
+ * @param {!Object|string} message
+ * @private
+ */
+ReaderTracker.prototype.readerFinishAddListener_ = function(message) {
+  var name = message['readerName'];
+  var port = message['port'];
+  var device = message['device'];
+  var returnCode = message['returnCode'];
+  var returnCodeHex = '0x' + (returnCode >>> 0).toString(16);
```
emaxx-google: GSC.DebugDump.dump would do the same; this would remove the tricky conversions here and would also apply the better formatting (e.g. the usual capital hexadecimal letters, proper length, treating negative numbers in the more expected way, etc.).

smart_card_connector_app/src/window.css
```diff
@@ -127,3 +127,28 @@ body {
   margin-right: 15px;
   text-decoration: none;
 }
+
+#readers-list > li {
+  display: flex;
+  align-items: center;
+}
+
+.circle {
```
emaxx-google: CSS class names should explain the meaning, but not describe the actual visual styling.
(This comment relates to "circle", "yellow", "red", "green".)

third_party/pcsc-lite/naclport/server/build/Makefile
```diff
@@ -144,6 +144,19 @@ PCSC_LITE_SERVER_SVC_CPPFLAGS := \
...
 
+# Special group just for readerfactory.c
+PCSC_LITE_SERVER_READERFACTORY_SOURCES := \
+	$(PCSC_LITE_SOURCES_PATH)/readerfactory.c \
+
+# Calls to RFAddReader and RFRemoveReader (readerfactory.c) are hooked to get
+# better information on reader status.
+PCSC_LITE_SERVER_READERFACTORY_CPPFLAGS := \
+	$(PCSC_LITE_SERVER_CPPFLAGS) \
+	-DRFAddReader=RFAddReaderServer \
```
emaxx-google: "Server" is not very informative suffix. In the existing code, it was used to distinguish between server-side and client-side functions. This is not the case in your code - your new code is also conceptually part of the server side.

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -55,11 +56,36 @@ var readersListElement = goog.dom.getElement('readers-list');
...
+ * @type {GSC.ReaderTracker}
+ */
+var readerTracker;
+
 function loadDeviceList() {
   logger.fine('Requesting available USB devices list...');
   chrome.usb.getDevices({'filters': USB_DEVICE_FILTERS}, getDevicesCallback);
 }
 
+function loadReaderList(readers) {
```
emaxx-google: s/load/display/ (or something like this)

smart_card_connector_app/src/reader-tracker.js
ivansandrk: True. I've added the types, but still the compiler doesn't complain at all. I would guess it doesn't do any checks for goog.structs.Map. Unless I did something wrong, I'll keep the types out. I've left a TODO in to discuss this with you live.

smart_card_connector_app/src/background.js
```diff
@@ -93,7 +99,8 @@ function launchedListener() {
 
 function openWindow() {
   var data = {'clientAppListUpdateSubscriber':
-      messageChannelPool.addOnUpdateListener.bind(messageChannelPool)};
+      messageChannelPool.addOnUpdateListener.bind(messageChannelPool)
+      ,'readerTracker': readerTracker};
```
emaxx-google: nit: move the comma to the previous line.

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,212 @@
...
+var ReaderStatus = {
+  INIT: 'init',
+  SUCCESS: 'success',
+  FAILURE: 'failure'
+};
+
+/**
+ * Structure used to store information about the reader.
+ * @param {string} name
+ * @param {ReaderStatus} status
```
emaxx-google: Looks like the "!ReaderStatus" type definition would be more exact.

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,212 @@
+/** @license
```
emaxx-google: I would suggest to move this JavaScript file under the third_party/pcsc-lite/naclport/server/src directory, because it provides a piece that is required for the PC/SC-Lite server to function correctly (because without this class, the messages sent from your new code would be left unhandled on the JS side).

And together with moving the file, add also something like "PcscLiteServer" into the provided symbols name, for better scoping.

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,212 @@
...
+      'reader_remove', this.readerRemoveListener_.bind(this), true);
+
+  /**
+   * @type {!goog.structs.Map.<number, !ReaderInfo>}
+   * @private
+   */
+  this.readers_ = new goog.structs.Map;
+
+  /**
+   * @type {!Array.<!function(!Array.<!ReaderInfo>)>}
```
emaxx-google: nit: "!" before "function" is not required.

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,212 @@
...
+ReaderTracker.prototype.readerInitAddListener_ = function(message) {
+  /** @type {string} */
+  var name = message['readerName'];
+  /** @type {number} */
+  var port = message['port'];
+  /** @type {string} */
+  var device = message['device'];
+
+  this.logger_.info(
+      'readerInitAddListener_ called for ' + name + ' (port ' + port +
```
emaxx-google: nit: Instead of including the private member names (like "readerInitAddListener_" in this case) into the log messages, better print some human-readable information (like "Adding a new reader "<name>" with port <port>, device "<device>" and status "initializing"" in this case).

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,212 @@
...
+      ', device ' + device + ')');
+
+  this.readers_.set(port, new ReaderInfo(name, ReaderStatus.INIT));
```
emaxx-google: Can we add an assertion here that no overwriting occurs in the map?

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,212 @@
...
+ReaderTracker.prototype.readerFinishAddListener_ = function(message) {
+  /** @type {string} */
+  var name = message['readerName'];
+  /** @type {number} */
+  var port = message['port'];
+  /** @type {string} */
+  var device = message['device'];
+  /** @type {number} */
+  var returnCode = message['returnCode'];
+  /** @type {string} */
```
emaxx-google: Here the type definition is not required.

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,212 @@
...
+/** @const */
+var ReaderTracker = GSC.ReaderTracker;
+
+/**
+ * @param {!Object|string} message
+ * @private
+ */
+ReaderTracker.prototype.readerInitAddListener_ = function(message) {
+  /** @type {string} */
+  var name = message['readerName'];
```
emaxx-google: Generally, I would much prefer to add assertions that all fields that were expected were extracted from the message.
(Maybe, you can find a way to code this nicely without repeating lots of assertion calls.)

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,212 @@
...
+  this.logger_.fine('Added an OnUpdateListener');
+  this.onUpdateListeners_.push(listener);
+};
+
+/**
+ * @param {function(!Array.<!ReaderInfo>)} listener
+ */
+ReaderTracker.prototype.removeOnUpdateListener = function(listener) {
+  this.logger_.fine('Removed an OnUpdateListener');
+  var index = this.onUpdateListeners_.indexOf(listener);
```
emaxx-google: In my opinion, using goog.array.remove would be simpler and more readable.

smart_card_connector_app/src/reader-tracker.js
```diff
@@ -0,0 +1,212 @@
...
+};
+
+/**
+ * Structure used to store information about the reader.
+ * @param {string} name
+ * @param {ReaderStatus} status
+ * @param {string=} opt_error
+ * @constructor
+ */
+var ReaderInfo = function(name, status, opt_error) {
```
emaxx-google: Are you not exporting this symbol intentionally?
It's a bit strange to keep this definition "private", given that values of this type are returned through public methods.

third_party/pcsc-lite/naclport/server/src/readerfactory_nacl.cc
```diff
@@ -0,0 +1,94 @@
...
+const char kReaderInitAddMessageType[] = "reader_init_add";
+const char kReaderFinishAddMessageType[] = "reader_finish_add";
+const char kReaderRemoveMessageType[] = "reader_remove";
+const char kNameMessageKey[] = "readerName";
+const char kPortMessageKey[] = "port";
+const char kDeviceMessageKey[] = "device";
+const char kReturnCodeMessageKey[] = "returnCode";
+
+void post_message(const char* type, const pp::VarDictionary& message_data)
+{
```
emaxx-google: nit: According to style guide, the opening bracket should be on the same line with the function declaration line.

third_party/pcsc-lite/naclport/server/src/readerfactory_nacl.cc
```diff
@@ -0,0 +1,94 @@
...
+  pp::VarDictionary message;
+  message.Set(kTypeMessageKey, type);
```
emaxx-google: It's better to use VarDictBuilder all the time, because it adds some assertions (first, because pp::VarDictionary::Set may return error; second, because trying to overwrite a key is usually an error), and because it ensures the proper value conversion (though in this exact case the standard pp::VarDictionary API is enough).

third_party/pcsc-lite/naclport/server/src/readerfactory_nacl.cc
```diff
@@ -0,0 +1,94 @@
...
+  message.Set(kDataMessageKey, message_data);
+
+  const google_smart_card::PPInstanceHolder* pp_instance_holder =
+      google_smart_card::GetGlobalPPInstanceHolder();
+  pp::Instance* const instance = pp_instance_holder->GetPPInstance();
+  instance->PostMessage(message);
```
emaxx-google: Looking at this code and at how RFAddReader/RFRemoveReader is used, it seems that some synchronization primitive would be required somewhere.
RFAddReader/RFRemoveReader are called on a background thread that is never finished, and it's not trivial to stop it before pp::Instance destruction begins.
So we would have to ensure that the code here works correctly with a possible concurrent pp::Instance destruction.

The similar problem was already solved in the past: see the LibusbOverChromeUsbGlobal class and how its lifetime is managed in the PpInstance class (in file smart_card_connector_app/src/pp_module.cc).
In short, the solution is:

1. Continue to have the singleton as a unique_ptr member of the PpInstance class.
2. Suppress destroying of the singleton in PpInstance destructor, so that any code that operates with it in a background thread still doesn't explode.
3. Add a method to the singleton class that "detaches" it from the pp::Instance object (essentially, the method should clear the pointer to pp::Instance in a thread-safe manner). This method should be called from the PpInstance destructor.

This solution ensures that, on one side, the code that uses the singleton won't explode during shutdown, and, on the other side, that it won't use the invalid pp::Instance* pointer.

third_party/pcsc-lite/naclport/server/src/readerfactory_nacl.cc
```diff
@@ -0,0 +1,94 @@
...
+}
+
+}  // namespace
+
+LONG RFAddReader(const char *readerNameLong, int port, const char *library,
```
emaxx-google: nit: Variables, according to style guide, should be named_like_this.

emaxx-google: nit: The style of positioning the "*" character in the existing code in this project was with space _after_, not _before_. E.g.: "const char* library".

common/js/src/messaging/common.js
```diff
@@ -56,4 +56,22 @@ function nonFatalDefaultServiceCallback(channel, serviceName, payload) {
...
+ * @param {string} key
+ */
+GSC.MessagingCommon.extractKey = function(message, key) {
+  GSC.Logging.checkWithLogger(LOGGER, goog.isObject(message));
+  goog.asserts.assert(goog.isObject(message));
+  GSC.Logging.checkWithLogger(
+      LOGGER,
+      goog.object.containsKey(message, key),
+      'Key "' + key + '" not present in message ' +
+      GSC.DebugDump.dump(message));
```
emaxx-google: It's safer to use debugDump instead of dump here, because GSC.MessagingCommon.extractKey is a generic function that may be called with a security/privacy-sensitive data.

common/js/src/messaging/common.js
```diff
@@ -56,4 +56,22 @@ function nonFatalDefaultServiceCallback(channel, serviceName, payload) {
...
+ * is actually present).
+ * @param {!Object|string} message
+ * @param {string} key
+ */
+GSC.MessagingCommon.extractKey = function(message, key) {
+  GSC.Logging.checkWithLogger(LOGGER, goog.isObject(message));
+  goog.asserts.assert(goog.isObject(message));
+  GSC.Logging.checkWithLogger(
+      LOGGER,
+      goog.object.containsKey(message, key),
```
emaxx-google: nit: Require goog.asserts and goog.object.
(I wonder why compiler sometimes allows missing requires.)

common/js/src/messaging/common.js
```diff
@@ -56,4 +56,22 @@ function nonFatalDefaultServiceCallback(channel, serviceName, payload) {
...
+      'Key "' + key + '" not present in message ' +
+      GSC.DebugDump.dump(message));
+
+  return message[key];
+}
```
emaxx-google: nit: ;

smart_card_connector_app/src/usb-devices-logging.js
```diff
@@ -0,0 +1,104 @@
...
+
+goog.require('GoogleSmartCard.DebugDump');
+goog.require('GoogleSmartCard.Logging');
+goog.require('goog.array');
+goog.require('goog.log.Logger');
+
+goog.scope(function() {
+
+/** @const */
+var USB_DEVICE_FILTERS = [{'interfaceClass': 0x0B}];
```
emaxx-google: This duplicates the constant in window-devices-displaying.js. In order to rid of copy-pasting, the constant may be removed from this file - it's not crucially important for the logging stuff.

smart_card_connector_app/src/usb-devices-logging.js
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * @type {!goog.log.Logger}
+ * @const
+ */
+var logger = GSC.Logging.getScopedLogger('ConnectorApp.UsbDevicesLogging');
+
+function loadDeviceList() {
+  logger.fine('Requesting available USB devices list...');
+  chrome.usb.getDevices({'filters': USB_DEVICE_FILTERS}, getDevicesCallback);
```
emaxx-google: Just pass {} instead of {'filters': USB_DEVICE_FILTERS} here in order to rid of USB_DEVICE_FILTERS.

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -55,26 +55,32 @@ var readersListElement = goog.dom.getElement('readers-list');
...
 }
 
-/**
- * @param {!chrome.usb.Device} device
- */
-function usbDeviceRemovedListener(device) {
-  logger.fine(
-      'A USB device was removed: ' + makeReaderDescriptionString(device));
-  loadDeviceList();
+function readerUpdatedListener(readers) {
```
emaxx-google: nit: You may just remove this function, it's basically a synonym to displayReaderList.

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -55,26 +55,32 @@ var readersListElement = goog.dom.getElement('readers-list');
...
 /**
- * @param {!chrome.usb.Device} device
+ * @type {GSC.PcscLiteServer.ReaderTracker}
  */
-function usbDeviceAddedListener(device) {
-  logger.fine('A USB device was added: ' + makeReaderDescriptionString(device));
-  loadDeviceList();
+var readerTracker;
+
+function displayReaderList(readers) {
```
emaxx-google: nit: add JSDoc containing the function parameter definition.

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -92,65 +98,18 @@ function addDeviceClickListener(e) {
...
   logger.fine('USB selection dialog finished, ' + devices.length + ' devices ' +
               'were chosen');
-  loadDeviceList();
 }
 
 GSC.ConnectorApp.Window.DevicesDisplaying.initialize = function() {
-  loadDeviceList();
-  chrome.usb.onDeviceAdded.addListener(usbDeviceAddedListener);
-  chrome.usb.onDeviceRemoved.addListener(usbDeviceRemovedListener);
+  readerTracker = GSC.PopupWindow.Client.getData()['readerTracker'];
```
emaxx-google: nit: Add assertion that the requested key was present in the data.

common/js/src/messaging/common.js
ivansandrk: Forgot about the difference, fixed.

smart_card_connector_app/src/window-devices-displaying.js
ivansandrk: Maybe this is a place where GSC.MessagingCommon.extractKey can be used? I think this is specifically the use case we made it for.

emaxx-google: It's similar, but not the same. Here, the word "messaging" is irrelevant. Also, allowing to pass argument as a string instead of object is strange outside the "messaging" context.

So I would agree if that would be a separate common function.

common/js/src/object-helpers.js
```diff
@@ -0,0 +1,56 @@
...
+ * @const
+ */
+var LOGGER = GSC.Logging.getScopedLogger('ObjectHelpers');
+
+/**
+ * Helper function which extracts key from object (first it makes sure the key
+ * is actually present).
+ * @param {!Object} object
+ * @param {string} key
+ */
```
emaxx-google: nit: add @return {?}

emaxx-google:
lgtm for the JS part, now only the singleton part in C++ is left

common/js/src/messaging/common.js
```diff
@@ -56,4 +58,17 @@ function nonFatalDefaultServiceCallback(channel, serviceName, payload) {
   channel.dispose();
 }
 
+/**
+ * Helper function which extracts key from object (first it makes sure the key
+ * is actually present).
+ * @param {!Object|string} object
+ * @param {string} key
+ */
```
emaxx-google: nit: add @return {?}

ivansandrk:
I think it's all finished now, I'm just a bit unsure of indentation in some places.

third_party/pcsc-lite/naclport/server/src/readerfactory_nacl.cc
```diff
@@ -0,0 +1,62 @@
...
+// This file contains a replacement function for the original readerfactory.c
+// PC/SC-Lite internal implementation.
+
+#include <google_smart_card_pcsc_lite_server/global.h>
+
+extern "C" {
+#include "readerfactory.h"
+
+LONG RFAddReaderOriginal(const char* reader_name, int port,
+    const char* library, const char* device);
```
emaxx-google: nit: I think that, according to style guide, either additional line with the arguments should be aligned to match the position of the first argument, or the first argument should be also on a separate line with 4 spaces indentation.

third_party/pcsc-lite/naclport/server/src/readerfactory_nacl.cc
```diff
@@ -0,0 +1,62 @@
...
+
+// This file contains a replacement function for the original readerfactory.c
+// PC/SC-Lite internal implementation.
+
+#include <google_smart_card_pcsc_lite_server/global.h>
+
+extern "C" {
+#include "readerfactory.h"
+
+LONG RFAddReaderOriginal(const char* reader_name, int port,
```
emaxx-google: nit: either include "wintypes.h" or use long here.

third_party/pcsc-lite/naclport/server/src/readerfactory_nacl.cc
```diff
@@ -0,0 +1,62 @@
...
+// OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+// IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT,
+// INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT
+// NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+// THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+// This file contains a replacement function for the original readerfactory.c
```
emaxx-google: nit: s/a replacement function/replacement functions/

third_party/pcsc-lite/naclport/server/src/google_smart_card_pcsc_lite_server/global.h
```diff
@@ -38,6 +43,30 @@ namespace google_smart_card {
 // already been initialized.
 void InitializeAndRunPcscLiteServer();
```
emaxx-google: The global declaration should be removed, and its comment now should be placed right before the corresponding method declaration.

third_party/pcsc-lite/naclport/server/src/google_smart_card_pcsc_lite_server/global.h
```diff
@@ -38,6 +43,30 @@ namespace google_smart_card {
...
+  void PostReaderFinishAddMessage(const char* reader_name, int port,
+                                  const char* device, long return_code) const;
+  void PostReaderRemoveMessage(const char* reader_name, int port) const;
+
+ private:
+  pp::Instance* GetPPInstance() const;
+  void PostMessage(const char* type, const pp::VarDictionary& message_data) const;
+
+  pp::Instance* pp_instance_;
+  mutable std::mutex mutex_;
```
emaxx-google: I'd recommend to place the mutex member before pp_instance, in order to give a hint that the pp_instance_ field is protected by the mutex.

third_party/pcsc-lite/naclport/server/src/google_smart_card_pcsc_lite_server/global.h
```diff
@@ -38,6 +43,30 @@ namespace google_smart_card {
...
+  void InitializeAndRunPcscLiteServer();
+
+  void PostReaderInitAddMessage(const char* reader_name, int port,
+                                const char* device) const;
+  void PostReaderFinishAddMessage(const char* reader_name, int port,
+                                  const char* device, long return_code) const;
+  void PostReaderRemoveMessage(const char* reader_name, int port) const;
+
+ private:
+  pp::Instance* GetPPInstance() const;
```
emaxx-google: nit: s/PP/Pp/
(the existing code in this project capitalizes all abbreviations, and also I think I met this recommendation in some of the Chromium style guides.)

third_party/pcsc-lite/naclport/server/src/google_smart_card_pcsc_lite_server/global.h
```diff
@@ -38,6 +43,30 @@ namespace google_smart_card {
 // already been initialized.
 void InitializeAndRunPcscLiteServer();
 
+class PcscLiteServerGlobal final {
+ public:
+  explicit PcscLiteServerGlobal(pp::Instance* pp_instance);
+  ~PcscLiteServerGlobal();
+  void Detach();
```
emaxx-google: Please write a comment for this function, because this is a nontrivial functionality.
(You may take as an example the existing docs for the Detach methods in other classes.)

third_party/pcsc-lite/naclport/server/src/google_smart_card_pcsc_lite_server/global.cc
```diff
@@ -68,9 +71,35 @@ void PcscLiteServerDaemonThreadMain() {
   }
 }
 
+PcscLiteServerGlobal* g_pcsc_lite_server = nullptr;
+
+const char kReaderInitAddMessageType[] = "reader_init_add";
```
emaxx-google: nit: Either move this constants block next to the kLoggingPrefix constant definition, or move the kLoggingPrefix constant definition here.

third_party/pcsc-lite/naclport/server/src/google_smart_card_pcsc_lite_server/global.h
```diff
@@ -38,6 +43,30 @@ namespace google_smart_card {
 // already been initialized.
 void InitializeAndRunPcscLiteServer();
 
+class PcscLiteServerGlobal final {
```
emaxx-google: Please write, at least short, comment with the class description.
Also it makes sense to mention the thread safety principles used in this class (e.g. that most methods are generally thread-safe, except GetInstance which is not protected against concurrent class construction or destruction).

third_party/pcsc-lite/naclport/server/src/google_smart_card_pcsc_lite_server/global.h
```diff
@@ -38,6 +43,30 @@ namespace google_smart_card {
...
+ public:
+  explicit PcscLiteServerGlobal(pp::Instance* pp_instance);
+  ~PcscLiteServerGlobal();
+  void Detach();
+
+  static const PcscLiteServerGlobal* GetInstance();
+
+  void InitializeAndRunPcscLiteServer();
```
emaxx-google: "PcscLiteServer" part duplicates the part of the class name. Maybe rename it to "StartDaemonThread"?

third_party/pcsc-lite/naclport/server/src/google_smart_card_pcsc_lite_server/global.cc
```diff
@@ -68,9 +71,35 @@ void PcscLiteServerDaemonThreadMain() {
...
+  GOOGLE_SMART_CARD_CHECK(!g_pcsc_lite_server);
+  g_pcsc_lite_server = this;
+}
+
+PcscLiteServerGlobal::~PcscLiteServerGlobal() {
+  GOOGLE_SMART_CARD_CHECK(g_pcsc_lite_server == this);
+  g_pcsc_lite_server = nullptr;
+}
+
+void PcscLiteServerGlobal::Detach () {
```
emaxx-google: nit: remove space before ().

third_party/pcsc-lite/naclport/server/src/google_smart_card_pcsc_lite_server/global.cc
```diff
@@ -153,4 +182,37 @@ void InitializeAndRunPcscLiteServer() {
       "successfully finished.";
 }
 
+const PcscLiteServerGlobal* PcscLiteServerGlobal::GetInstance() {
+  GOOGLE_SMART_CARD_CHECK(g_pcsc_lite_server);
+  return g_pcsc_lite_server;
+}
+
+void PcscLiteServerGlobal::PostMessage(
```
emaxx-google: nit: Reorder the method definition order to match the order of declaration in the header file.

third_party/pcsc-lite/naclport/server/src/readerfactory_nacl.cc
ivansandrk: I always actually wondered this... doesn't matter that "readerfactory.h" includes "wintypes.h", good practice says include it anyway?

third_party/pcsc-lite/naclport/server/src/google_smart_card_pcsc_lite_server/global.h
ivansandrk: Missed that one, thanks!

third_party/pcsc-lite/naclport/server/src/google_smart_card_pcsc_lite_server/global.h
ivansandrk: Got it!

third_party/pcsc-lite/naclport/server/src/google_smart_card_pcsc_lite_server/global.h
ivansandrk: Forgot about that one, also dead code.

third_party/pcsc-lite/naclport/server/src/readerfactory_nacl.cc
emaxx-google: Yes:

> You should include all the headers that define the symbols you rely upon, except in the unusual case of forward declaration. If you rely on symbols from bar.h, don't count on the fact that you included foo.h which (currently) includes bar.h: include bar.h yourself, unless foo.h explicitly demonstrates its intent to provide you the symbols of bar.h. However, any includes present in the related header do not need to be included again in the related cc (i.e., foo.cc can rely on foo.h's includes).

https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

emaxx-google:
lgtm

PR #30, opened 2016-08-18
Bullet-proof debug dumping code

emaxx-google:
Ivan, PTAL.

This ensures that debug dumping never results in throwing an exception when in Release mode.

### DIFF
--- a/common/js/src/logging/debug-dump.js
+++ b/common/js/src/logging/debug-dump.js
@@ -219,45 +219,40 @@ function dump(value) {
   return encodeJson(value);
 }
 
-if (goog.DEBUG) {
-  /**
-   * Generates a debug textual representation of the passed value.
-   *
-   * Note that in the cases when the passed value may contain privacy-sensitive
-   * data, the GoogleSmartCard.DebugDump.debugDump method should be used instead.
-   * @param {?} value
-   * @return {string}
-   */
-  GSC.DebugDump.dump = function(value) {
+/**
+ * Generates a debug textual representation of the passed value.
+ *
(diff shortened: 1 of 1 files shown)


common/js/src/logging/debug-dump.js
```diff
@@ -220,7 +217,32 @@ GSC.DebugDump.dump = function(value) {
...
+if (goog.DEBUG) {
+  /**
+   * Generates a debug textual representation of the passed value.
+   *
+   * Note that in the cases when the passed value may contain privacy-sensitive
+   * data, the GoogleSmartCard.DebugDump.debugDump method should be used instead.
+   * @param {?} value
+   * @return {string}
+   */
+  GSC.DebugDump.dump = function(value) {
```
ivansandrk: Is it possible to write "GSC.DebugDump.dump = dump;" instead?

ivansandrk:
lgtm

common/js/src/logging/debug-dump.js
emaxx-google: Initially I thought to agree with you, but then I realized that this may affect the debugging abilities a bit: when compiling in Debug mode, the Closure Compiler is asked to produce a debug name for each function (like "$GoogleSmartCard$DebugDump$dump$"), which makes exception messages and stack traces informative.

However, it's good that this question arose. After looking into the resulting compiled file, it turned out that compiler for some reason doesn't generate function names in the constructs like:
`if (goog.DEBUG) Foo.bar = function() {...}; else Foo.bar = function() {...};`

So I'll change the code to use this construct instead:
`Foo.bar = function() { if (goog.DEBUG) ... else ... };`

PR #31, opened 2016-08-18
Protect debug dumping code against undefined classes

emaxx-google:
Ivan, PTAL.

This hopefully fixes a intermittent problem that I observed when debug dump function failed with TypeError saying that right-hand argument of "instanceof" operator cannot be undefined. This is a pretty mysterious error - my guess was that some standard classes are not initialized by the time when the extension code starts.

### DIFF
--- a/common/js/src/logging/debug-dump.js
+++ b/common/js/src/logging/debug-dump.js
@@ -187,13 +187,13 @@ GSC.DebugDump.dump = function(value) {
   // produce thorny false positives.
   //
   // TODO(emaxx): Think about a proper solution that deals with cyclic references.
-  if (value instanceof Document)
+  if (Document && value instanceof Document)
     return '<Document>';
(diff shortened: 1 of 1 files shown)


common/js/src/logging/debug-dump.js
```diff
@@ -187,13 +187,13 @@ GSC.DebugDump.dump = function(value) {
   // produce thorny false positives.
   //
   // TODO(emaxx): Think about a proper solution that deals with cyclic references.
-  if (value instanceof Document)
+  if (Document && value instanceof Document)
```
ivansandrk: Haha interesting behavior. I was about to write "maybe put a if (!value) return 'undefined';" when I realized the right side is what's undefined. Strange indeed.

ivansandrk:
lgtm

PR #35, opened 2016-08-25
Display a warning when running Connector on non-Chrome OS system

### LINKED ISSUE #27: Display proper message when app is opened on non-Chrome OS operating systems
1. Open connector on Windows/Mac/Linux Chrome.

Expected
Runs properly

Actual
?

We need to document the behavior on other platforms and show the proper signal to users. In the beginning we had blocked installation on other platforms but that got in the way of developers testing things out in their dev envrionments before publishing to the store (it's hard to develop on Chrome OS as middleware is a very low-level software that needs particular tools for development not found on Chrome OS).

emaxx-google:
This commit fixes #27.

### DIFF
--- a/smart_card_connector_app/src/window-main.js
+++ b/smart_card_connector_app/src/window-main.js
@@ -28,6 +28,7 @@ goog.require('GoogleSmartCard.I18n');
 goog.require('GoogleSmartCard.Logging');
 goog.require('GoogleSmartCard.PopupWindow.Client');
 goog.require('goog.dom');
+goog.require('goog.dom.classlist');
 goog.require('goog.events');
 goog.require('goog.events.EventType');
 goog.require('goog.log.Logger');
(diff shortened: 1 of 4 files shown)


emaxx-google:
Iva, PTAL.

smart_card_connector_app/src/window-main.js
```diff
@@ -59,8 +56,27 @@ GSC.ConnectorApp.Window.AppsDisplaying.initialize();
 GSC.ConnectorApp.Window.DevicesDisplaying.initialize();
 GSC.ConnectorApp.Window.LogsExporting.initialize();
 
+GSC.I18n.adjustElementsTranslation();
```
emaxx-google: Moved this line before showWindow call, so that the window is displayed when all titles are already in place (so that no flickering may occur).

ivansandrk:
lgtm

PR #42, opened 2016-09-22
Initial commit

ivansandrk:
- rudimentary functionality, app permission removed by clicking on a x button

### DIFF
--- a/smart_card_connector_app/src/background.js
+++ b/smart_card_connector_app/src/background.js
@@ -109,8 +109,26 @@ function openWindow() {
       'readerTrackerSubscriber':
           readerTracker.addOnUpdateListener.bind(readerTracker),
       'readerTrackerUnsubscriber':
-          readerTracker.removeOnUpdateListener.bind(readerTracker)};
+          readerTracker.removeOnUpdateListener.bind(readerTracker),
+      'removeApp':
+          removeApp};
   GSC.PopupWindow.Server.createWindow(WINDOW_URL, WINDOW_OPTIONS, data);
+}
+
+/**
+ * Remove connected app (disconnect it and remove stored permission).
+ * @param {string} clientExtensionId
+ */
+function removeApp(clientExtensionId) {
+  // TODO(isandrk): Make the deeper code do the bulk of work and handling of
+  // dialog stuff, then return the result as promise and remove channel if
+  // neccessary.
+  var checkerInstance =
+      GSC.PcscLiteServerClientsManagement.ClientHandler.getPermissionsChecker();
+  checkerInstance.removeAppPermission(clientExtensionId);
+  // TODO(isandrk): Argument referring to the extensionId isn't consistent
+  // across source files (its name).
+  messageChannelPool.disposeChannel(clientExtensionId);
 }
 
 /**
--- a/smart_card_connector_app/src/window-apps-displaying.js
+++ b/smart_card_connector_app/src/window-apps-displaying.js
@@ -23,11 +23,15 @@ goog.provide('GoogleSmartCard.ConnectorApp.Window.AppsDisplaying');
 
 goog.require('GoogleSmartCard.DebugDump');
 goog.require('GoogleSmartCard.Logging');
+goog.require('GoogleSmartCard.ObjectHelpers');
 goog.require('GoogleSmartCard.PcscLiteServerClientsManagement.PermissionsChecking.KnownApp');
 goog.require('GoogleSmartCard.PcscLiteServerClientsManagement.PermissionsChecking.KnownAppsRegistry');
+goog.require('GoogleSmartCard.PcscLiteServerClientsManagement.PermissionsChecking.Checker');
 goog.require('goog.array');
 goog.require('goog.asserts');
 goog.require('goog.dom');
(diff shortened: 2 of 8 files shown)


ivansandrk:
Hey Max please take a stab at this! Thanks

smart_card_connector_app/src/background.js
```diff
@@ -109,11 +109,29 @@ function openWindow() {
...
+ * Remove connected app (disconnect it and remove stored permission).
+ * @param {string} clientExtensionId
+ */
+function removeApp(clientExtensionId) {
+  // TODO(isandrk): Make the deeper code do the bulk of work and handling of
+  // dialog stuff, then return the result as promise and remove channel if
+  // neccessary.
+  var checkerInstance =
+      GSC.PcscLiteServerClientsManagement.ClientHandler.getPermissionsChecker();
+  checkerInstance.removeAppPermission(clientExtensionId);
```
emaxx-google: If the permission was forced by an admin policy (see the PermissionsChecking.ManagedRegistry class), then removal of the permission shouldn't be allowed.

First of all, this situation should be probably guarded from the UI side - i.e. by displaying your "x" button as grayed out.

But, second, this should also be checked here in the code to protect from the possible race conditions.

smart_card_connector_app/src/window-apps-displaying.js
```diff
@@ -81,7 +96,24 @@ function updateAppView(knownAppsPromise, appIds, knownApps) {
...
+          removePermission,
+          goog.events.EventType.CLICK,
+          removeApp.bind(undefined, appIds[i]));
+          // TODO(isandrk): ^ Is there some problem with this (renaming)? Possibly
+          // do the binding inside getPermissionsChecker (and rename it obviously).
+
+      var newElement = goog.dom.createDom(
+          'li', undefined, text, removePermission);
       goog.dom.append(appListElement, newElement);
     }
```
emaxx-google: It'd be also useful to render the denied items in the end of the list too (e.g. as grayed out or in some other way).
With that, the user would have an opportunity to undo her erroneous "Deny" selections.

(Note that we currently "workaround" this by not storing the denied decisions of the user in the storage - i.e. it's currently enough to restart the Chrome OS user session in order to flush all denied decisions. But after your change lands, we could also start preserving the denied decisions too, for better UX.)

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker.js
```diff
@@ -320,6 +320,35 @@ UserPromptingChecker.prototype.storeUserSelection_ =
 };
 
 /**
+ * Removed stored permission for App with given id.
+ * @param {string} clientAppId
+ */
+UserPromptingChecker.prototype.removeAppPermission = function(clientAppId) {
```
emaxx-google: I'd suggest to change this method to something like "revokeUserSelection", as it would be better to be able to undo both kinds of the user selections: both "allow" and "deny".

And note that we should also do something with the corresponding method in the PermissionsChecking.Checker class - because, on one side, it's a more generic class, but, on the other side, it's impossible to revoke the policy-granted permissions.

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker.js
```diff
@@ -320,6 +320,35 @@ UserPromptingChecker.prototype.storeUserSelection_ =
...
+                          GSC.DebugDump.dump(dumpedValue));
+        chrome.storage.local.set(goog.object.create(
+            LOCAL_STORAGE_KEY, dumpedValue));
+      },
+      function() {
+        this.logger.warning(
+            'Failed to remove app permission from the local storage');
+      },
+      this);
+};
```
emaxx-google: That's not enough: you have to deal with the checkPromiseMap_ member which would otherwise keep the previous user selection cached.
That's probably why you observed the bug that the client was still displayed in the UI after you revoked its access.

(You may consider the option of refactoring the class internals to make the access revocation be a bit more natural.)

smart_card_connector_app/src/window-apps-displaying.js
```diff
@@ -112,14 +144,17 @@ function onUpdateListener(appListArg) {
...
 
 GSC.ConnectorApp.Window.AppsDisplaying.initialize = function() {
+  var passedData = GSC.PopupWindow.Client.getData();
+
   logger.fine('Registering listener on connected apps update');
   // FIXME(emaxx): Do unsubscription too.
-  // FIXME(emaxx): Use GSC.ObjectHelpers.extractKey to ensure that the expected
-  // object is passed to the window.
-  var data = GSC.PopupWindow.Client.getData()['clientAppListUpdateSubscriber'];
+  var data = extractKey(passedData, 'clientAppListUpdateSubscriber');
```
emaxx-google: nit: Collapse this with the following line (as the "data" temporary variable doesn't bring much information).

emaxx-google:
Closing this, so that it doesn't hang around in the pull requests list. If/when we decide to restart this effort, we can always revive this branch.

PR #59, opened 2018-05-16
Fix incorrect storage of user selection data

### LINKED ISSUE #57: App crash-restart loop caused by bad data stored for user selection
Steps to reproduce:
1. Install the Connector app.
2. Install at least two middleware apps that talk to the Connector app. Press "Allow" in both prompts about granting permissions to them.

Expected:
This works.

Actual:
The Connector app crashes and fails to start due to a fatal error during initialization. After several crash-restart loops, the browser disables the App.

Workaround:
Uninstall the Connector and install it back.

/cc @dskaram

emaxx-google:
The Connector app was incorrectly serializing the user selection data
when saving it in the local storage, in cases when the stored data
contained more than one item.

Besides corrupting the stored user selections, this bug was leading to
the app crashes and crash-restart loops, as the error was unrecoverable.

This fixes #57.

### DIFF
--- a/common/js/src/container-helpers.js
+++ b/common/js/src/container-helpers.js
@@ -17,8 +17,6 @@
 goog.provide('GoogleSmartCard.ContainerHelpers');
 
 goog.require('GoogleSmartCard.Logging');
-goog.require('goog.array');
-goog.require('goog.object');
 
 goog.scope(function() {
 
@@ -27,17 +25,19 @@ var GSC = GoogleSmartCard;
 
 /**
  * Converts the given Map to an Object.
+ *
+ * The Map must have only string or numeric keys.
  * @param {!Map} map
  * @return {!Object}
  */
 GSC.ContainerHelpers.buildObjectFromMap = function(map) {
-  var keyValuePairArray = Array.from(map.entries());
-  GSC.Logging.check(keyValuePairArray.length == map.size);
-  var flatKeysAndValuesArray = goog.array.flatten(keyValuePairArray);
-  GSC.Logging.check(flatKeysAndValuesArray.length == map.size * 2);
-  var object = goog.object.create(flatKeysAndValuesArray);
-  GSC.Logging.check(goog.object.getCount(object) == map.size);
-  return object;
+  let obj = {};
+  for (let [key, value] of map) {
+    GSC.Logging.check(goog.isString(key) || goog.isNumber(key),
+                      'Invalid type for object key');
+    obj[key] = value;
+  }
+  return obj;
 };
 
 });  // goog.scope
--- a/common/js/src/container-helpers.js
+++ b/common/js/src/container-helpers.js
@@ -1,0 +1,43 @@
+/** @license
+ * Copyright 2018 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+goog.provide('GoogleSmartCard.ContainerHelpers');
+
+goog.require('GoogleSmartCard.Logging');
+goog.require('goog.array');
+goog.require('goog.object');
+
+goog.scope(function() {
+
+/** @const */
+var GSC = GoogleSmartCard;
+
+/**
+ * Converts the given Map to an Object.
+ * @param {!Map} map
+ * @return {!Object}
+ */
+GSC.ContainerHelpers.buildObjectFromMap = function(map) {
+  var keyValuePairArray = Array.from(map.entries());
+  GSC.Logging.check(keyValuePairArray.length == map.size);
+  var flatKeysAndValuesArray = goog.array.flatten(keyValuePairArray);
+  GSC.Logging.check(flatKeysAndValuesArray.length == map.size * 2);
+  var object = goog.object.create(flatKeysAndValuesArray);
+  GSC.Logging.check(goog.object.getCount(object) == map.size);
+  return object;
+};
+
+});  // goog.scope
--- a/third_party/pcsc-lite/naclport/server_clients_management/build/js_unittests/Makefile
+++ b/third_party/pcsc-lite/naclport/server_clients_management/build/js_unittests/Makefile
@@ -13,6 +13,12 @@
 # limitations under the License.
 
 
+#
+# This makefile builds and runs JavaScript unit tests for the
+# "server_clients_management" component of the PC/SC-Lite naclport.
+#
+
+
 TARGET := js_pcsc_lite_server_clients_management_unittests
 
 include ../../../../../../common/make/common.mk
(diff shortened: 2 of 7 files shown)


common/js/src/container-helpers.js
```diff
@@ -0,0 +1,43 @@
...
+
+/** @const */
+var GSC = GoogleSmartCard;
+
+/**
+ * Converts the given Map to an Object.
+ * @param {!Map} map
+ * @return {!Object}
+ */
+GSC.ContainerHelpers.buildObjectFromMap = function(map) {
```
pavolmarko: I guess an alternative way to put it would be:
```
function (map) {
    let obj = {};
    for (let [k,v] of map) {
        obj[k] = v;
    }
    return obj;
}
```

Would that maybe be clearer than going from `map` -> `[[key1, value1], [key2, value2], ..]` -> flattened array of `[key1, value1, key2, value2, ..]` -> `goog.object.create`? Or is there a reason why we _must_ use goog.object.create?

pavolmarko: What (should happen or) happens if the Map has a key that is not a string? Should we test that?

third_party/pcsc-lite/naclport/server_clients_management/build/js_unittests/Makefile
```diff
@@ -0,0 +1,31 @@
...
+include ../../../../../../common/make/common.mk
+
+include $(COMMON_DIR_PATH)/make/js_building_common.mk
+
+include $(COMMON_DIR_PATH)/js/include.mk
+include ../../include.mk
+
+
+JS_COMPILER_INPUT_PATHS := \
+	$(PCSC_LITE_SERVER_CLIENTS_MANAGEMENT_JS_COMPILER_INPUT_DIR_PATHS) \
```
pavolmarko: Where does this come from? I don't understand this (but my Makefile knowledge is limited :) )

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker-unittests.js
```diff
@@ -0,0 +1,288 @@
...
+ */
+function setUpChromeStorageMock(
+    mockControl, propertyReplacer, fakeInitialData, expectedDataToBeWritten) {
+  propertyReplacer.set(
+      chrome, 'storage',
+      {local: {
+         get: mockControl.createFunctionMock('chrome.storage.local.get'),
+         set: mockControl.createFunctionMock('chrome.storage.local.set')}});
+
+  /** @type {?} */ chrome.storage.local.get;
```
pavolmarko: Why do we need this (and the same for .set)?
Is it to tell the closure compiler that these exist?

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker-unittests.js
```diff
@@ -0,0 +1,288 @@
...
+  chrome.storage.local.get(STORAGE_KEY, ignoreArgument).$does(
+      function(key, callback) {
+        callback(fakeInitialData);
+      });
+
+  /** @type {?} */ chrome.storage.local.set;
+  if (!goog.isNull(expectedDataToBeWritten))
+    chrome.storage.local.set(expectedDataToBeWritten);
```
pavolmarko: Does this already set up an expectation that .set will be called with the passed parameter ( and not called with anything else ), or do we need to call something like $does / $return / .. on it?

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker-unittests.js
```diff
@@ -0,0 +1,288 @@
...
+ * corresponding result.
+ */
+function setUpDialogMock(mockControl, mockedBehavior) {
+  /** @type {?} */
+  var mockedFunction = mockControl.createMethodMock(
+      GSC.PopupWindow.Server, 'runModalDialog');
+  var mockAction;
+  switch (mockedBehavior) {
+    case MockedDialogBehavior.NOT_RUN:
+      return;
```
pavolmarko: Should this fail somehow if the dialog is invoked when not expected?

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker-unittests.js
```diff
@@ -0,0 +1,288 @@
...
+ * @param {!goog.Promise} promise
+ * @return {!goog.Promise.<undefined>} A promise which is rejected when
+ * |promise| gets resolved, and resolved (with no data) when |promise| is
+ * rejected.
+ */
+function negatePromise(promise) {
+  return promise.then(function() {
+    fail('Unexpected successful response');
+  }, function() {
+    // Do not pass on the rejection.
```
pavolmarko: I'd have expected a resolve() call here -- does this not just ignore the rejection?

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker.js
```diff
@@ -293,8 +294,8 @@ UserPromptingChecker.prototype.storeUserSelection_ =
   this.localStoragePromiseResolver_.promise.then(
```
pavolmarko: (I'm too stupid to add the comment where it actually belongs)
Do we need to explicitly ignore garbage values in 
|contents| in localStorageLoadedCallback_ now (if written by a buggy previous version)?

common/js/src/container-helpers.js
emaxx-google: Great point! Indeed, my original attempt in the functional style is failing compared to an ad-hoc approach.

emaxx-google: Good point, added an assertion (I think it's reasonable to assume only strings and numbers to be passed here by a valid code) and tests.

third_party/pcsc-lite/naclport/server_clients_management/build/js_unittests/Makefile
emaxx-google: It comes from ../../include.mk. I've added a short comment for this piece.

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker-unittests.js
emaxx-google: The problem is that compiler actually knows them and verifies against the real signatures, which leads to errors as there's no "$does" property in the function's return value.
Added a comment on that.

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker-unittests.js
emaxx-google: Nope, the expectation is already set up. For instance, the tests fail if I repeat this line.

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker-unittests.js
emaxx-google: It will - by default, a function mock expects to be not called at all.

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker-unittests.js
emaxx-google: Maybe I misunderstood the question, but there's no resolve() method on the promise; and returning undefined from this function should resolve the resulting promise - according to the docs:

"If the Promise is rejected, the onRejected callback will be invoked with the rejection reason as argument, and the child Promise will be resolved with the return value or rejected with the thrown value of the callback."

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker.js
emaxx-google: Yes, that's a good suggestion, we should not crash due to anything in the externally stored data. I was actually planning to do this, but in a follow-up change.

emaxx-google:
PTAL.

emaxx-google:
Thanks!

PR #61, opened 2018-06-05
Make local storage data not crash the app

### LINKED ISSUE #57: App crash-restart loop caused by bad data stored for user selection

emaxx-google:
The Connector app shouldn't crash due to invalid data
loaded from the local storage. Errors are logged, but
invalid entries are just ignored.

This fixes #57.

### DIFF
--- a/third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker.js
+++ b/third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker.js
@@ -168,18 +168,7 @@ UserPromptingChecker.prototype.loadLocalStorage_ = function() {
 UserPromptingChecker.prototype.localStorageLoadedCallback_ = function(items) {
   this.logger.finer('Loaded the following data from the local storage: ' +
                     GSC.DebugDump.dump(items));
-
-  /** @type {!Map.<string, boolean>} */
-  var storedUserSelections = new Map;
-  if (items[LOCAL_STORAGE_KEY]) {
-    var contents = items[LOCAL_STORAGE_KEY];
-    GSC.Logging.checkWithLogger(this.logger, goog.isObject(contents));
-    goog.object.forEach(contents, function(userSelection, appId) {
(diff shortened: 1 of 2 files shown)


third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker.js
```diff
@@ -192,6 +181,36 @@ UserPromptingChecker.prototype.localStorageLoadedCallback_ = function(items) {
...
+  var contents = localStorageItems[LOCAL_STORAGE_KEY];
+  if (!goog.isObject(contents)) {
+    this.logger.warning(
+        'Corrupted local storage data - expected an object, received: ' +
+        GSC.DebugDump.dump(contents));
+    return storedUserSelections;
+  }
+  goog.object.forEach(contents, function(userSelection, appId) {
+    if (!goog.isBoolean(userSelection)) {
+      this.logger.warning(
```
pavolmarko: Should we make some attempt at removing these from the storage?
(+) We would not spam the logs with warnings forever
(-) Harder to diagnose issues with this if we just remove stuff and warn only once.

I can't decide :-)

emaxx-google: Thanks for raising this question.

My primary intention was to avoid complicating the implementation because of that cleanup which wouldn't affect anything user-visible.

Also note that the bad data would be overwritten anyway by the next write operation, which will happen after the user grants a permission for some app. I'm expecting this to happen soon after our update, as, presumable, many users keep their existing set of installed extensions, which implies that a permission prompt will be shown again for the previously corrupted entries.

So I'm inclined to keep having no explicit cleanup in the code.

PR #62, opened 2018-06-14
Fix dead links

yoe:
Debian has decomissioned alioth, replacing it by salsa.debian.org, a gitlab instance. As a result, all the CCID and PC/SC-Lite URLs are now dead.

Replace them by their apdu.fr counterparts (or the git repository in one case).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,7 +39,7 @@ The high-level overview of the architecture is:
     section below.
 
     The App is bundled with the **free USB CCID driver** (see
-    <https://pcsclite.alioth.debian.org/ccid.html>). This driver is
+    <https://ccid.apdu.fr/>). This driver is
     patched appropriately in order to be able to work through the
     **chrome.usb** Chrome API.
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -49,7 +49,7 @@ The high-level overview of the architecture is:
     technically (each driver can be created as a separate App talking to
     the Connector App through some defined interface). A list of readers
     supported by CCID can be found on their website
-    <https://pcsclite.alioth.debian.org/ccid/section.html>.
+    <https://ccid.apdu.fr/#readers>.
```
emaxx-google: The exact equivalent of the old link is "https://ccid.apdu.fr/ccid/section.html", let's use this.

yoe:
ping. Is this waiting for something from me? If so, please tell me what. If not, just ignore ;-)

README.rst
emaxx-google: Ping - I don't see a reason to switch from a direct link to the devices information.

emaxx-google:
yoe - please see my comment above about README.rst.

yoe:
oh, sorry, missed that -- this should be better.

emaxx-google:
Thanks!

PR #68, opened 2019-06-12
Update crxmake.sh to produce CRX3 format

emaxx-google:
Since M73, Chrome only allows installation of extension packages in the
CRX3 format. The previously used CRX2 format is now deprecated.

This CL changes the crxmake.sh script to produce the .crx file in the
new format. Unfortunately, the original distribution of this shell
script - the developer.chrome.com article - doesn't provide the new
version of the crxmake.sh file, so I had to fix it manually to produce
the file in the new format.


emaxx-google:
This is superseded by #69 - that one uses the approach that you proposed: namely, use Chrome capabilities instead of having a dedicated shell script.

PR #69, opened 2019-06-12
Build CRX files using Chrome

emaxx-google:
Use Chrome's --pack-extension command-line switch in order to produce
.crx files, rather than manually crafting the .crx files via a magic
shell script.

This also resolves the problem with the transition onto the new CRX3
format for the .crx files: the old shell script was producing the CRX2
files which aren't accepted by Chrome starting from version M73, and
fixing the script was non-trivial.

This change also deletes the crxmake.sh third-party dependency
altogether.


emaxx-google:
@alex292 - could you PTAL?

emaxx-google:
Sorry - I've just updated the CL in order to:
* remove the useless .p8 files for the example apps;
* update the gitignore files.

emaxx-google:
Thanks!

PR #70, opened 2019-06-15
Suppress compiler warnings in PC/SC-Lite code

emaxx-google:
Compile winscard_svc.c with "-Wno-return-type" and debuglog.c with
"-Wno-format", so that the PC/SC-Lite code is able to build by the
current NaCl toolchain without warnings (any warnings are fatal).

This fixes the recent regression that prevented the code from
being compiled using the current pepper_47 SDK. Apparently, this
old SDK was updated at some point by Chromium authors without
public announcements, and the new toolchain produces compiler
warnings that were not present before.

In particular, this commit fixes the following fatal build errors:

  third_party/pcsc-lite/src/src/winscard_svc.c:809:1: error:
    control reaches end of non-void function
    [-Werror,-Wreturn-type]
  }
  ^

  third_party/pcsc-lite/src/src/debuglog.c:196:32: error:
    format specifies type 'long' but the argument has type
    'pthread_t' (aka 'struct __nc_basic_thread_data *')
    [-Werror,-Wformat]
  time_pfx, delta, time_sfx, thread_id,
                             ^~~~~~~~~

### DIFF
--- a/third_party/pcsc-lite/naclport/server/build/Makefile
+++ b/third_party/pcsc-lite/naclport/server/build/Makefile
@@ -91,10 +91,10 @@ PCSC_LITE_COMMON_CPPFLAGS := \
 
 # Part of the list of the original PC/SC-Lite daemon source files to be
 # compiled (see also PCSC_LITE_SERVER_SVC_SOURCES,
-# PCSC_LITE_SERVER_READERFACTORY_SOURCES variables)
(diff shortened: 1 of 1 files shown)


emaxx-google:
@alex292, could you PTAL?
Admittedly, you're unfamiliar with the building system in this project, but this change is pretty much following the existing pattern in this file - see, for example, PCSC_LITE_SERVER_READERFACTORY_SOURCES.

P.S. Sorry, I unintentionally already landed this commit into master. But I'll surely create a follow-up commit to address the feedback, should there be any.

PR #72, opened 2019-06-17
Use bundled NaCl SDK instead of fetching it

### LINKED ISSUE #71: CERTIFICATE_VERIFY_FAILED when running initialize.sh
The env/initialize.sh script is currently broken:

  third_party.fancy_urllib.InvalidCertificateException: Host storage.googleapis.com returned an invalid certificate ([SSL: CERTIFICATE_VERIFY_FAILED] cer
tificate verify failed (_ssl.c:727))

The problem seems to be with the CA certificates validation in the current version of the NaCl SDK.

emaxx-google:
Change the initialize.sh script to use the pinned and patched
version of the NaCl SDK instead of fetching it from
storage.googleapis.com and auto-updating it.

This fixes #71 (see also the previous commits associated with
that issue).

This also includes these required fixes:

Update the cacerts.txt file in the NaCl SDK to contain an
up-to-date certificate that is required in order to talk to
storage.googleapis.com - the "GlobalSign_Root_CA_-_R2"
certificate.
This is needed for addressing #71, since currently the NaCl SDK
cannot fetch the Pepper SDK due to TLS error when doing network
requests.

Fix the fancy_urllib script in order to be compatible with
OpenSSL 1.1.1+.
For this, enforce the PROTOCOL_TLSv1 protocol when making the
ssl.wrap_socket() call, since otherwise the call fails due to
the host validation.


emaxx-google:
@alex292, PTAL.

P.S. I was rushing to land these fixes into master in order to unblock external developers, but will be happy to address any feedback in follow-up CLs.

third_party/nacl_sdk/nacl_sdk/sdk_tools/cacerts.txt
```diff
@@ -1,77 +1,22 @@
-# Copyright (c) 2012 The Chromium Authors. All rights reserved.
-# Use of this source code is governed by a BSD-style license that can be
-# found in the LICENSE file.
-
-# The following SSL certificates are used to validate the SSL connection
```
alex292: nit: I guess this comment here is still valid? Also the cert name/identifier, although it's just one now.

emaxx-google: Done, good point.

PR #75, opened 2019-06-17
Import Closure Compiler 20190528

emaxx-google:


emaxx-google:
@alex292 - PTAL, I've updated the change to reduce the number of awkward type casts in the test.

emaxx-google:
Thanks. Your feedback about the initial patchset (where I had a lot of hacky casts) helped a lot.

PR #77, opened 2019-06-21
Fix error in cert provider bridge under Linux

emaxx-google:
Fix the TypeError exception during the
CertificateProviderBridge.Backend disposal when the
chrome.certificateProvider API is unavailable (e.g., when the
extension is run on Linux) :

  TypeError: Cannot read property 'onSignDigestRequested' of undefined

This just is a minor cleanup, since normally the Backend instance is
never disposed of - this only happens when another fatal error occurs
(e.g., when the NaCl module crashes). So this fix only avoids
emitting an additional spurious error which could hamper the
investigation of the original root cause.


emaxx-google:
Thanks!

PR #78, opened 2019-06-21
Best-effort fix for NaCl crash log message delivery

emaxx-google:
Wait a little bit between emitting the FATAL log message and
crashing, so that the emitted log message has a chance to be
delivered to the JavaScript side where it's easier to be
discovered by the developer.

This fix is imperfect and works only with some probability
(depending on the system load and, likely, on whether the fatal
error happened on the main thread or on a background thread). But
that's still an improvement, and also the error logging in this
case is best-effort anyway.


common/cpp/src/google_smart_card_common/logging/logging.cc
```diff
@@ -156,8 +158,14 @@ LogMessage::LogMessage(LogSeverity severity)
 LogMessage::~LogMessage() {
   if (ShouldLogWithSeverity(severity_)) {
     EmitLogMessage(severity_, stream_.str());
-    if (severity_ == LogSeverity::kFatal)
+    if (severity_ == LogSeverity::kFatal) {
+      // Wait for some time before crashing, to leave a chance for the log
+      // message with the crash reason to be delivered the JavaScript side. This
```
alex292: nit: delivered *to* the Javascript

emaxx-google: Done

PR #79, opened 2019-07-05
C++ example: More verbose logging of PIN requests

emaxx-google:
Log every invocation of requestPin/stopPinRequest functions even
in Release mode, together will their input parameters and whether
the operation succeeds.

This should help in investigating bug reports in the third-party
extensions in the future.

The only piece of information that is intentionally NOT logged is
the actual PIN.

### DIFF
--- a/example_cpp_smart_card_client_app/src/chrome_certificate_provider/api_bridge.cc
+++ b/example_cpp_smart_card_client_app/src/chrome_certificate_provider/api_bridge.cc
@@ -17,8 +17,10 @@
 #include <thread>
 #include <utility>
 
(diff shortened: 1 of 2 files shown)


emaxx-google:
Example of the logs that will be printed:

```
[INFO] chrome.certificateProvider.requestPin(options={"signRequestId": 0x7B}): called...                                                                         
[INFO] chrome.certificateProvider.requestPin: returning true (success)
[INFO] chrome.certificateProvider.stopPinRequest(options={"signRequestId": 0x7B}): called...                                                                     
[INFO] chrome.certificateProvider.stopPinRequest: returning success
```

emaxx-google:
Thanks!

PR #81, opened 2019-07-25
Allow logging the PC/SC-Lite API call results

emaxx-google:
Add the implementation of logging the results of the PC/SC-Lite API
calls that are made from a NaCl module. This should aid in the
future in the investigation of bugs in third-party middleware
extensions.

The implementation is disabled by default, since even under normal
circumstances a typical smart card middleware makes dozens of
PC/SC-Lite requests per single certificates listing / signature
operation.

This additional logging can be enabled by running these snippets in
the JavaScript console of the extension's background page:

* enabling the logging of return codes:

  $jscomp.scope.pcscLiteNaclClientBackend.logger.setLevel(goog.log.Level.FINE)

* additionally, enabling the dumping of all values and buffers (use
  with care - this may include privacy/security-sensitive data) :

  goog.DEBUG=true

### DIFF
--- a/third_party/pcsc-lite/naclport/js_client/src/api.js
+++ b/third_party/pcsc-lite/naclport/js_client/src/api.js
@@ -2211,16 +2211,22 @@ goog.exportProperty(
  * structure.
  *
  * This function is safe to be used in Release builds, because all potentially
- * privacy-sensitive data is stripped away from the resulting text.
+ * security/privacy-sensitive data is stripped away from the resulting text.
  *
  * @return {string}
  */
 API.ResultOrErrorCode.prototype.getDebugRepresentation = function() {
   if (this.errorCode == API.SCARD_S_SUCCESS) {
+    // Apply debugDump() to all result values except the error code, since they
+    // can contain sensitive information in them (like the contents of the
+    // digital signature). The debugDump() function dumps the values only when
+    // goog.DEBUG is true, and otherwise shows only the number of values.
     const dumpedItems = this.resultItems && this.resultItems.length ?
         ' ' + GSC.DebugDump.debugDump(this.resultItems) : '';
     return 'SCARD_S_SUCCESS' + dumpedItems;
   }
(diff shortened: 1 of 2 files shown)


third_party/pcsc-lite/naclport/js_client/src/api.js
```diff
@@ -2205,6 +2206,29 @@ goog.exportProperty(
     'getResult',
     API.ResultOrErrorCode.prototype.getResult);
 
+/**
+ * Generates a debug textual representation of the PC/SC-Lite call result
+ * structure.
+ *
+ * This function is safe to be used in Release builds, because all potentially
+ * privacy-sensitive data is stripped away from the resulting text.
```
alex292: I guess dump() just returns the content, but |errorCode| doesn't contain any privacy-sensitive information and debugDump() is responsible for removing the potential privacy-sensitive information?
optional: I would explicitely mention the use of debugDump() in this comment then

alex292:
follow-up: the change has no reference to anything related to "goog.DEBUG=true", I guess that was just mentioned for documentation purpose?

third_party/pcsc-lite/naclport/js_client/src/api.js
emaxx-google: Done - added inline comments into the implementation.

emaxx-google:
> follow-up: the change has no reference to anything related to "goog.DEBUG=true", I guess that was just mentioned for documentation purpose?

Yes, this is simply a documentation, and it's correct: the goog.DEBUG is a standard Closure variable that is used in a number of places, in particular in the debugDump() function that I'm calling in this CL.

emaxx-google:
> > follow-up: the change has no reference to anything related to "goog.DEBUG=true", I guess that was just mentioned for documentation purpose?
> 
> Yes, this is simply a documentation, and it's correct: the goog.DEBUG is a standard Closure variable that is used in a number of places, in particular in the debugDump() function that I'm calling in this CL.

I've added a mention of this variable into the inline comments.

PR #82, opened 2019-07-25
C++ example: Reload app on NaCl module crash

emaxx-google:
Add code into the example C++ client app that triggers a self
reload when a fatal error with the NaCl module happens (e.g.,
when it crashes).

### DIFF
--- a/example_cpp_smart_card_client_app/src/background.js
+++ b/example_cpp_smart_card_client_app/src/background.js
@@ -89,6 +89,22 @@ var logger = GSC.Logging.getLogger(
 var naclModule = new GSC.NaclModule(
     'nacl_module.nmf', GSC.NaclModule.Type.PNACL);
 
+// Reload the self app when the NaCl module is disposed of (e.g., due to a
+// crash).
+
+/**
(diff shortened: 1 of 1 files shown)


emaxx-google:
BTW, here's the internal bug for the context: http://b/138099922 (for Googlers only).

example_cpp_smart_card_client_app/src/background.js
```diff
@@ -89,6 +89,22 @@ var logger = GSC.Logging.getLogger(
 var naclModule = new GSC.NaclModule(
     'nacl_module.nmf', GSC.NaclModule.Type.PNACL);
 
+// Reload the self app when the NaCl module is disposed of (e.g., due to a
```
alex292: 1) "self app" ?
2) any reason why the // and /**...*/ comment are separated?
3) The listener doesn't actually reload the app, but triggers sth that will cause a reload somewhere else
4) This only happens in release mode (-> mention in comment?)

emaxx-google: Done with all 1..4. Thanks.

PR #84, opened 2019-08-02
C++ example: Log versions and time on startup

emaxx-google:
Add a bit more logging into the C++ example middleware app, to
make it slightly easier to investigate bug reports from the users
(since many users forget to specify some important information).

### DIFF
--- a/example_cpp_smart_card_client_app/src/background.js
+++ b/example_cpp_smart_card_client_app/src/background.js
@@ -83,7 +83,7 @@ var logger = GSC.Logging.getLogger(
     goog.DEBUG ? goog.log.Level.FINE : goog.log.Level.INFO);
 
 logger.info(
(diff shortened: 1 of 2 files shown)


emaxx-google:
PTAL. This is analogous to the logging that was previously added into the Smart Card Connector app: see commit 265b3c750541900b04a5a6cb9fa779ab6a64cf93 and commit 45b1a1e45943c8b6692b5b9fa97238d1e0364a48.

emaxx-google:
> LGTM
> optional: I tend to use ES6 template strings over concatenating strings since it's usually easier to read: https://developers.google.com/web/updates/2015/01/ES6-Template-Strings

Good point, done. I have never used them in the past, but I see how in this specific example they greatly improve the readability.

emaxx-google:
Thanks for the review!

PR #85, opened 2019-08-18
Remove smart-ssh from the list of known clients

FabianHenneke:
The functionality of smart-ssh has meanwhile been integrated into the Secure Shell app. As a result, smart-ssh is now deprecated and no longer listed on the web store. I think that it is time to remove it from the list.


emaxx-google:
Thanks for doing this cleanup.

PR #86, opened 2019-09-16
download-google-smart-card-client-library: port to Python 3

vapier:
The current code only works with Python 2.  Port it to support both
Python 2 & Python 3.

Basically we have to handle the urllib2->urllib.request rename, and
the fact that urlopen returns bytes, not strings.  This is true for
both Python versions.


emaxx-google:
LGTM
Thanks a lot, Mike!

PR #88, opened 2019-09-26
Trusted clients registry unit tests

emaxx-google:
Add unit tests for the TrustedClientsRegistry
class which is responsible for querying information about the trusted client
application origins from the bundled JSON config.

### DIFF
--- a/third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/trusted-clients-registry-unittest.js
+++ b/third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/trusted-clients-registry-unittest.js
@@ -21,7 +21,6 @@ goog.require('goog.Promise');
 goog.require('goog.json');
 goog.require('goog.net.HttpStatus');
 goog.require('goog.testing');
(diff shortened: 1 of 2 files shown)


emaxx-google:
Time to unveil another 4-year-old pull request :)

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/trusted-clients-registry-unittest.js
```diff
@@ -0,0 +1,101 @@
+/**
+ * @license
+ * Copyright 2019 Google Inc. All Rights Reserved.
```
Fabian-Sommer, line 3: Should we update the year?

Fabian-Sommer (approved):
LGTM

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/trusted-clients-registry-unittest.js
emaxx-google: done :laughing:

PR #89, opened 2019-09-26
Handle more Xhrio failures

emaxx-google:
Detect cases in KnownAppsRegistry when goog.net.XhrIo reports
isSuccess()==true, but at the same time getResponseText()==''. As
it turned out, this is also a possible way of signaling an error from
XhrIo.

This is a generic cleanup, since in production we don't expect any
errors happening there (because the fetched URL belong to our
own CRX).


emaxx-google:
Time to review it, 4 years since I drafted out this PR :D

Fabian-Sommer (approved):
LGTM - I was starting to think this CL was for decorative purposes :)

emaxx-google:
> LGTM - I was starting to think this CL was for decorative purposes :)

One can see others' drafts? I didn't realize that!

PR #91, opened 2019-11-29
Get rid of dependency on Boost

emaxx-google:
Remove all usages of the Boost C++ library from the project.

This allows to resolve the build issues caused by some problems
with the NaCl webport of Boost:

  /usr/bin/git-secrets: line 116: /usr/lib/git-core/git: Argument list too long

### DIFF
--- a/common/cpp/src/google_smart_card_common/numeric_conversions.h
+++ b/common/cpp/src/google_smart_card_common/numeric_conversions.h
@@ -95,11 +95,11 @@ inline bool CastInt64ToInteger(
   }
   *error_message = FormatPrintfTemplate(
       "The integer value is outside the range of type \"%s\": %" PRId64 " not "
-      "in [%" PRId64 "; %" PRIu64 "] range",
+      "in [%" PRIdMAX "; %" PRIuMAX "] range",
       type_name.c_str(),
       value,
-      static_cast<int64_t>(std::numeric_limits<T>::min()),
-      static_cast<uint64_t>(std::numeric_limits<T>::max()));
+      static_cast<intmax_t>(std::numeric_limits<T>::min()),
+      static_cast<uintmax_t>(std::numeric_limits<T>::max()));
   return false;
 }
 
@@ -117,10 +117,10 @@ inline bool CastIntegerToDouble(
   std::string formatted_value;
   if (value >= 0) {
     formatted_value = FormatPrintfTemplate(
-        "%" PRIu64, static_cast<uint64_t>(value));
+        "%" PRIuMAX, static_cast<uintmax_t>(value));
   } else {
     formatted_value = FormatPrintfTemplate(
-        "%" PRId64, static_cast<int64_t>(value));
+        "%" PRIdMAX, static_cast<intmax_t>(value));
   }
   *error_message = FormatPrintfTemplate(
       "The integer %s cannot be converted into a floating-point double value "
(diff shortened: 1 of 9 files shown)


common/cpp/src/google_smart_card_common/pp_var_utils/extraction.cc
```diff
@@ -59,8 +35,9 @@ bool VarAsInteger(
     if (!CastDoubleToInt64(var.AsDouble(), &integer, error_message))
       return false;
   } else {
-    *error_message = FormatBoostFormatTemplate(
-        kErrorWrongType, kIntegerJsTypeTitle, DebugDumpVar(var));
+    *error_message = FormatPrintfTemplate(
+        "Expected a value of type \"%s\", instead got: %s",
+        kIntegerJsTypeTitle, DebugDumpVar(var).c_str());
```
alex292: nit: formatting inconsistent with similar patterns below

common/cpp/src/google_smart_card_common/pp_var_utils/extraction.cc
```diff
@@ -18,32 +18,8 @@
 
 #include <google_smart_card_common/numeric_conversions.h>
 
-// Error messages and related bits of strings
-const char kErrorWrongType[] =
```
alex292: this string is used multiple times below, so I would keep it defined as variable here

common/cpp/src/google_smart_card_common/pp_var_utils/extraction.h
```diff
@@ -264,8 +257,10 @@ inline bool IsVarArraySizeValid(
     size_t expected_size,
     std::string* error_message) {
   if (var.GetLength() != expected_size) {
-    *error_message = FormatBoostFormatTemplate(
-        kErrorWrongArraySize, expected_size, var.GetLength());
+    *error_message = FormatPrintfTemplate(
+        "Expected an array of size %zu, instead got: %" PRIu32,
```
alex292: %zu -> %PR... ?

emaxx-google: Ack - the first parameter has the |size_t| type, and for it, I believe, the "zu" specifier should be used. The "PR..." thing is mainly for types with fixed size.

common/cpp/src/google_smart_card_common/pp_var_utils/extraction.cc
emaxx-google: Done

common/cpp/src/google_smart_card_common/pp_var_utils/extraction.cc
emaxx-google: Done

emaxx-google: Done (I was afraid that compiler won't be able to verify the printf format specifiers in that case, but looks like it's able to understand the constants defined in the same file as well)

common/cpp/src/google_smart_card_common/numeric_conversions.h
```diff
@@ -96,12 +93,13 @@ inline bool CastInt64ToInteger(
...
-      type_name,
+  *error_message = FormatPrintfTemplate(
+      "The integer value is outside the range of type \"%s\": %" PRId64 " not "
+      "in [%" PRId64 "; %" PRIu64 "] range",
+      type_name.c_str(),
       value,
-      std::numeric_limits<T>::min(),
-      std::numeric_limits<T>::max());
+      static_cast<int64_t>(std::numeric_limits<T>::min()),
+      static_cast<uint64_t>(std::numeric_limits<T>::max()));
```
Fabian-Sommer: nit: I think intmax_t and uintmax_t would be better here

emaxx-google: Done.
Thanks for the tip! Didn't know about this thing.

emaxx-google:
Thanks all!

PR #93, opened 2019-11-29
Import ccid 1.4.31

emaxx-google:


emaxx-google:
Fabian, PTAL - you don't need to review most of these changes, since it's mostly a copy of the third-party code (namely, everything under third_party/ccid/src). It's mostly to show you how we're doing the uprevs in this project.

A couple of notes for posterity:
* I downloaded the sources of the 1.4.31 version from https://ccid.apdu.fr and replaced with them everything under third_party/ccid/src.
* Then I manually reapplied the patch third_party/ccid/patches/0001-Patch-ccid-tweak-ifdDriverOptions-option.patch.
* I updated the version numbers and URLs in third_party/ccid/README.google and third_party/ccid/naclport/include.mk.

Issue #100, opened 2019-12-05
Stop USB communication when the Chrome OS user session is locked


emaxx-google:
The Smart Card Connector app should monitor the Chrome OS session state and, if it's running in-session and the session becomes locked, it should stop talking to the smart card readers. If we don't do this, then the users that use smart cards for logging into Chrome OS won't be able to unlock their session, since the Lock Screen implementation talks to another instance of the app that runs in the sign-in profile context, but only one app can talk to a single USB device at a time.

The chrome.loginState API can be used for getting all the needed information: https://cs.chromium.org/chromium/src/chrome/common/extensions/api/login_state.idl

emaxx-google:
For the reference, the chrome.loginState API was enabled for the Smart Card Connector app on the Chrome side in the issue https://crbug.com/1034927. The minimum Chrome version is 81.0.4028.0.

PR #101, opened 2019-12-11
Unlock from lock screen

Fabian-Sommer:
Solves issue #100 

Previously, smartcard users could not unlock the lock screen since
the user space smartcard connector app would still own the smartcard reader.
This makes the app show no available devices when the user is locking the
screen. PCSC automatically releases devices that it used to own but no longer
show as present. This check happens once every second.

This requires the additional loginState permission.

### DIFF
--- a/third_party/closure-compiler/src_additional/chrome_extensions.js
+++ b/third_party/closure-compiler/src_additional/chrome_extensions.js
@@ -152,14 +152,12 @@ chrome.loginState.SessionState = {
 };
 
 /**
- * @param {function(!chrome.loginState.ProfileType)} callback Callback function.
- * @return {undefined}
+ * @param {function(!chrome.loginState.ProfileType)} callback
  */
 chrome.loginState.getProfileType = function(callback) {};
 
 /**
- * @param {function(!chrome.loginState.SessionState)} callback Callback function.
- * @return {undefined}
+ * @param {function(!chrome.loginState.SessionState)} callback
  */
 chrome.loginState.getSessionState = function(callback) {};
 
--- a/third_party/closure-compiler/src_additional/chrome_extensions.js
+++ b/third_party/closure-compiler/src_additional/chrome_extensions.js
@@ -136,19 +136,19 @@ chrome.loginState = {};
  * @enum {string}
  */
 chrome.loginState.ProfileType = {
-  SIGNIN_PROFILE: '',
-  USER_PROFILE: '',
+  SIGNIN_PROFILE: 'SIGNIN_PROFILE',
+  USER_PROFILE: 'USER_PROFILE',
 };
 
 /**
  * @enum {string}
  */
 chrome.loginState.SessionState = {
-  UNKNOWN: '',
-  IN_OOBE_SCREEN: '',
-  IN_LOGIN_SCREEN: '',
-  IN_SESSION: '',
-  IN_LOCK_SCREEN: '',
+  UNKNOWN: 'UNKNOWN',
+  IN_OOBE_SCREEN: 'IN_OOBE_SCREEN',
+  IN_LOGIN_SCREEN: 'IN_LOGIN_SCREEN',
+  IN_SESSION: 'IN_SESSION',
+  IN_LOCK_SCREEN: 'IN_LOCK_SCREEN',
 };
 
 /**
--- a/third_party/closure-compiler/src_additional/chrome_extensions.js
+++ b/third_party/closure-compiler/src_additional/chrome_extensions.js
@@ -126,3 +126,51 @@ chrome.certificateProvider.SignDigestRequestEvent.prototype.removeListener = fun
  * @type {!chrome.certificateProvider.SignDigestRequestEvent}
  */
 chrome.certificateProvider.onSignDigestRequested;
+
+/**
+ * @const
+ */
+chrome.loginState = {};
+
+/**
+ * @enum {string}
+ */
+chrome.loginState.ProfileType = {
+  SIGNIN_PROFILE: '',
+  USER_PROFILE: '',
+};
+
+/**
+ * @enum {string}
+ */
+chrome.loginState.SessionState = {
+  UNKNOWN: '',
+  IN_OOBE_SCREEN: '',
+  IN_LOGIN_SCREEN: '',
+  IN_SESSION: '',
+  IN_LOCK_SCREEN: '',
+};
+
+/**
+ * @param {function(!chrome.loginState.ProfileType)} callback Callback function.
+ * @return {undefined}
+ */
+chrome.loginState.getProfileType = function(callback) {};
+
+/**
+ * @param {function(!chrome.loginState.SessionState)} callback Callback function.
+ * @return {undefined}
+ */
+chrome.loginState.getSessionState = function(callback) {};
+
+/**
+ * Event that triggers when the session state changes.
+ * @interface
+ * @extends {ChromeBaseEvent<function(!chrome.loginState.SessionState)>}
+ */
+chrome.loginState.SessionStateEvent = function() {};
+
+/**
+ * @type {!chrome.loginState.SessionStateEvent}
+ */
+chrome.loginState.onSessionStateChanged;
--- a/third_party/libusb/naclport/src/chrome_usb/chrome-login-state-hook-unittest.js
+++ b/third_party/libusb/naclport/src/chrome_usb/chrome-login-state-hook-unittest.js
@@ -1,0 +1,171 @@
+/** @license
+ * Copyright 2018 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+goog.require('GoogleSmartCard.Libusb.ChromeLoginStateHook');
+goog.require('goog.Promise');
+goog.require('goog.testing');
+goog.require('goog.testing.MockControl');
+goog.require('goog.testing.jsunit');
+goog.require('goog.testing.mockmatchers');
+
+goog.setTestOnly();
+
+goog.scope(function() {
+
+/** @const */
+var GSC = GoogleSmartCard;
+
+/** @const */
+var ChromeLoginStateHook = GSC.Libusb.ChromeLoginStateHook;
+/** @const */
+var ignoreArgument = goog.testing.mockmatchers.ignoreArgument;
+
+/** @const */
+var FAKE_DEVICES = [{"device": 0x00, "manufacturerName": "ACS", 
+    "productId": 0x000090CC, "productName": "CCID USB Reader", 
+    "serialNumber": "", "vendorId": 0x0000072F, "version": 0x00000100}];
+/** @const */
+var FAKE_CONFIGS = [{"active": true, "configurationValue": 0x01, 
+    "extra_data": ArrayBuffer, "interfaces": [
+      {"alternateSetting": 0x00, "endpoints": [
+        {"address": 0x81, "direction": "in", "extra_data": ArrayBuffer,
+        "maximumPacketSize": 0x08, "pollingInterval": 0x10, 
+        "type": "interrupt", "usage": "periodic"}, 
+        {"address": 0x02, "direction": "out", "extra_data": ArrayBuffer, 
+        "maximumPacketSize": 0x40, "pollingInterval": 0x00, "type": "bulk"}, 
+        {"address": 0x82, "direction": "in", "extra_data": ArrayBuffer, 
+        "maximumPacketSize": 0x40, "pollingInterval": 0x00, "type": "bulk"}], 
+      "extra_data": ArrayBuffer[0x36, 0x21, 0x00, 0x01, 0x00, 0x07, 0x03, 
+        0x00, 0x00, 0x00, 0xA0, 0x0F, 0x00, 0x00, 0xA0, 0x0F, 0x00, 0x00, 
+        0x00, 0x00, 0x2A, 0x00, 0x00, 0x24, 0x40, 0x05, 0x00, 0x00, 0xF7, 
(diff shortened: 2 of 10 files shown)


third_party/closure-compiler/src_additional/chrome_extensions.js
```diff
@@ -126,3 +126,51 @@ chrome.certificateProvider.SignDigestRequestEvent.prototype.removeListener = fun
...
+/**
+ * @const
+ */
+chrome.loginState = {};
+
+/**
+ * @enum {string}
+ */
+chrome.loginState.ProfileType = {
+  SIGNIN_PROFILE: '',
```
emaxx-google: Thinking more on this, I think the string values should be non-empty, e.g.:

`  SIGNIN_PROFILE: 'SIGNIN_PROFILE',`

otherwise the `chrome.loginState.ProfileType.SIGNIN_PROFILE` expression would emit an empty string?

third_party/libusb/naclport/src/chrome_usb/chrome-usb-backend.js
```diff
@@ -228,10 +231,43 @@ ChromeUsbBackend.prototype.reportRequestError_ = function(
...
 };
 
+/**
+ * @param {!chrome.loginState.ProfileType} profileType
+ * @private
+ */
+ChromeUsbBackend.prototype.receiveProfileType_ = function(profileType) {
+  if (chrome.loginState &&
+      profileType === chrome.loginState.ProfileType.USER_PROFILE) {
+    chrome.loginState.onSessionStateChanged.addListener(
```
emaxx-google: It might be that the session state was already in the undesired state by this point. I think there's no other way than also using getSessionState() in order to handle this correctly.

As a side note, it would be great if we could have unit test coverage for these scenarios.

third_party/libusb/naclport/src/chrome_usb/chrome-usb-backend.js
```diff
@@ -228,10 +231,43 @@ ChromeUsbBackend.prototype.reportRequestError_ = function(
...
+  }
+};
+
+/**
+ * @param {!chrome.loginState.SessionState} sessionState
+ * @private
+ */
+ChromeUsbBackend.prototype.sessionStateChangedListener_ = function(
+    sessionState) {
+  if (chrome.loginState &&
```
emaxx-google: Ditto for asserting the impossible conditions instead of handling them.

third_party/libusb/naclport/src/chrome_usb/chrome-usb-backend.js
```diff
@@ -228,10 +231,43 @@ ChromeUsbBackend.prototype.reportRequestError_ = function(
...
+      sessionState === chrome.loginState.SessionState.IN_LOCK_SCREEN) {
+    this.logger.info("No longer showing USB devices.");
+    this.fakeNoDevices = true;
+  } else {
+    this.logger.info("Showing USB devices.");
+    this.fakeNoDevices = false;
+  }
+};
+
```
emaxx-google: As a general feedback for this CL, I'm thinking whether we can structure this new logic a bit better by extracting into a separate class. What we're adding here is a separate piece of functionality that can be isolated from the existing logic in this class that is, basically, responsible for converting the data between the NaCl library format and the chrome.usb format.

Additionally, extracting new code into a separate class will allow testing it easily. As a bonus, this will help us when doing the migration from the chrome.usb API to the other API (WebUSB?), since the logic around the Lock Screen should stay intact regardless of how exactly the USB APIs are implemented.

There are many ways for implementing this. For example, we can say that the ChromeUsbBackend class provides "addHook"/"removeHook" methods, and the hooks are called on every request and response.

But I haven't looked at your other CL yet - you might want to come up with some approach that can be reused in both CLs.

third_party/libusb/naclport/src/chrome_usb/chrome-usb-backend.js
```diff
@@ -222,12 +243,15 @@ ChromeUsbBackend.prototype.reportRequestError_ = function(
...
  * @param {string} debugRepresentation
+ * @param {string} functionName API function that was called
  * @param {!goog.promise.Resolver} promiseResolver
  * @param {!Array} resultArgs
  * @private
  */
 ChromeUsbBackend.prototype.reportRequestSuccess_ = function(
-    debugRepresentation, promiseResolver, resultArgs) {
+    debugRepresentation, functionName, promiseResolver, resultArgs) {
+  this.requestSuccessHooks_.forEach(function f(hook, key, map) {
```
emaxx-google: nit: The name "f" is not informative. Either make it unnamed, or use a lambda, or maybe just have a loop.

third_party/libusb/naclport/src/chrome_usb/chrome-usb-backend.js
```diff
@@ -69,6 +70,21 @@ var ChromeUsbBackend = GSC.Libusb.ChromeUsbBackend;
 ChromeUsbBackend.prototype.logger = GSC.Logging.getScopedLogger(
     'Libusb.ChromeUsbBackend');
 
+/**
```
emaxx-google: nit: Please write a short comment about when this hook gets called.

third_party/libusb/naclport/src/chrome_usb/chrome-login-state-hook-unittest.js
```diff
@@ -0,0 +1,187 @@
...
+      propertyReplacer.reset();
+      sessionStateListeners = [];
+    }
+    
+    setUpChromeLoginStateMock(
+        propertyReplacer, fakeProfileType, fakeSessionState);
+
+    /** @preserveTry */
+    try {
+      testCallback(new GSC.Libusb.ChromeLoginStateHook());
```
emaxx-google: nit: Any reason for not creating the hook directly inside each test?

third_party/libusb/naclport/src/chrome_usb/chrome-login-state-hook-unittest.js
```diff
@@ -0,0 +1,187 @@
...
+
+/**
+ * Notifies listeners on chrome.loginState.sessionStateChanged with the new
+ * session state. Does not change the value returned by 
+ * chrome.loginState.getSessionState() !
+ * @param {number} fakeNewSessionState
+ */
+function notifySessionStateListeners(fakeNewSessionState) {
+  goog.array.forEach(sessionStateListeners, function(listener) {
+      listener(fakeNewSessionState);
```
emaxx-google: nit: The indentation should be 4 spaces here.

third_party/libusb/naclport/src/chrome_usb/chrome-login-state-hook-unittest.js
```diff
@@ -0,0 +1,187 @@
...
+      }
+    ));
+
+goog.exportSymbol(
+    'test_ChromeLoginStateHook_FiltersInLockScreen', makeTest(
+      userProfileType, inLockScreenState,
+      function(loginStateHook) {
+        var hook = loginStateHook.getRequestSuccessHook();
+
+        var apiCallResult = ['fakeResult'];
```
emaxx-google: nit: Use "let" for mutable values like this (and "const" otherwise).

third_party/libusb/naclport/src/chrome_usb/chrome-usb-backend.js
```diff
@@ -69,6 +70,21 @@ var ChromeUsbBackend = GSC.Libusb.ChromeUsbBackend;
...
+ * @param {string} key
```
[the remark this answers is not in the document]
Fabian-Sommer: goog.events.ListenerMap should only be used in the goog.events package. Even within the package, a string or goog.events.EventId seems to be required when adding a listener.

third_party/libusb/naclport/src/chrome_usb/chrome-login-state-hook.js
```diff
@@ -0,0 +1,115 @@
...
+ * @param {string} functionName chrome.usb function that was called
+ * @param {!Array} callResults
+ * @private
+ */
+ChromeLoginStateHook.prototype.requestSuccessHook_ = function(functionName, 
+                                                              callResults) {
+  if (this.simulateDevicesAbsent_ &&
+      (functionName === 'getDevices' ||
+       functionName === 'getConfigurations')) {
+    callResults.length = 0;
```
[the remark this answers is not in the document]
Fabian-Sommer: This would lead to copying the call results (which are arrays that can hold complex objects) for each hook on every call. Would this be ok performance-wise?

third_party/libusb/naclport/src/chrome_usb/chrome-usb-backend.js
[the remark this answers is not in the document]
emaxx-google: Good point - I didn't see that comment.
Then we can use a simple array of functions? JavaScript supports === for functions, so the removal of the observer can be implemented.

It's just that having to invent a unique string for every added function is an error-prone approach and is not consistent with standard addListener()/removeListener().

third_party/libusb/naclport/src/chrome_usb/chrome-login-state-hook.js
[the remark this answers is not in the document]
emaxx-google: If we return the same array, then it's returned "by reference" without any deep copying.

third_party/libusb/naclport/src/chrome_usb/chrome-login-state-hook-unittest.js
```diff
@@ -0,0 +1,194 @@
...
+}
+
+/**
+ * Simulates Chrome changing the session state.
+ * @param {string} fakeNewSessionState
+ * @param {!goog.testing.PropertyReplacer} propertyReplacer
+ */
+function changeSessionState(fakeNewSessionState, propertyReplacer) {
+  propertyReplacer.replace(
+    chrome.loginState,
```
emaxx-google: nit: +2 spaces of indentation here.

third_party/libusb/naclport/src/chrome_usb/chrome-login-state-hook.js
```diff
@@ -0,0 +1,171 @@
...
+  this.boundOnGotSessionState_ = this.onGotSessionState_.bind(this);
+  /** 
+   * @type {!goog.promise.Resolver}
+   * @private 
+   */
+  this.hookIsReadyResolver_ = goog.Promise.withResolver();
+  if (chrome.loginState) {
+    chrome.loginState.getProfileType(this.onGotProfileType_.bind(this));
+  } else {
+    GSC.Logging.getScopedLogger('Libusb.LoginStateHook').warning(
```
emaxx-google: nit: |this.logger.warning(|?

third_party/libusb/naclport/src/chrome_usb/chrome-login-state-hook.js
```diff
@@ -0,0 +1,171 @@
...
+  if (sessionState === chrome.loginState.SessionState.IN_LOCK_SCREEN
+      && !this.simulateDevicesAbsent_) {
+    this.logger.info("No longer showing USB devices.");
+    this.simulateDevicesAbsent_ = true;
+  } else if (sessionState !== chrome.loginState.SessionState.IN_LOCK_SCREEN
+             && this.simulateDevicesAbsent_) {
+    this.logger.info("Showing USB devices.");
+    this.simulateDevicesAbsent_ = false;
+  }
+  this.hookIsReadyResolver_.resolve();
```
emaxx-google: nit: Probably worth adding a comment that this line can be hit multiple times, but only the first execution will actually resolve the promise and the further ones will be ignored.

third_party/libusb/naclport/src/chrome_usb/chrome-login-state-hook.js
```diff
@@ -0,0 +1,171 @@
...
+ * @param {!Array} callResults
+ * @return {!Array} new callResults
+ * @private
+ */
+ChromeLoginStateHook.prototype.requestSuccessHook_ = function(functionName, 
+                                                              callResults) {
+  if (this.simulateDevicesAbsent_ &&
+      (functionName === 'getDevices' ||
+       functionName === 'getConfigurations')) {
+    callResults[0] = [];
```
emaxx-google: This still modifies the value of the argument, which in general is quite error-prone and I think should only be used for a clear reason and with a specific documentation.

I don't remember - do we have more than one item in |callResults| for these calls? If not, we could just do "return []", otherwise goog.array.clone() or something similar could help.

third_party/libusb/naclport/src/chrome_usb/chrome-usb-backend.js
```diff
@@ -55,12 +57,34 @@ GSC.Libusb.ChromeUsbBackend = function(naclModuleMessageChannel) {
...
+  this.setupDonePromiseResolver_ = goog.Promise.withResolver();
+
+  /** 
+   * @type {!GoogleSmartCard.DeferredProcessor}
+   * Queue of API requests that are not yet processed because setup is still in
+   * progress.
+   * @private  
+   */
+  this.deferredProcessor_ = new GSC.DeferredProcessor(
+    this.setupDonePromiseResolver_.promise);
```
emaxx-google: nit: +2 spaces of indentation here.

third_party/libusb/naclport/src/chrome_usb/chrome-usb-backend.js
```diff
@@ -116,6 +168,17 @@ ChromeUsbBackend.prototype.logDevices_ = function(devices) {
...
+      this.processRequest_.bind(this, payload));
+};
+
+/**
+ * @param {!Object} payload
+ * @return {!goog.Promise}
+ * @private
+ */
+ChromeUsbBackend.prototype.processRequest_ = function(payload) {
+  //extract into processRequest and call DeferredProcessor with it
```
emaxx-google: nit: Was this a temporary comment? It's not clear what it means now.

third_party/libusb/naclport/src/chrome_usb/chrome-usb-backend.js
```diff
@@ -222,12 +290,15 @@ ChromeUsbBackend.prototype.reportRequestError_ = function(
...
+ * @param {string} functionName API function that was called
  * @param {!goog.promise.Resolver} promiseResolver
  * @param {!Array} resultArgs
  * @private
  */
 ChromeUsbBackend.prototype.reportRequestSuccess_ = function(
-    debugRepresentation, promiseResolver, resultArgs) {
+    debugRepresentation, functionName, promiseResolver, resultArgs) {
+  this.requestSuccessHooks_.forEach(function(hook) {
+      resultArgs = hook(functionName, resultArgs);});
```
emaxx-google: nit: -2 spaces of indentation here.
Also normally the "}" closing bracket should go into a new line.

Issue #90, opened 2019-11-29
Add retry logic into reader initialization code


emaxx-google:
Currently, the Smart Card Connector app only makes a single attempt to connect to each USB device.

We should change this to be more bulletproof by retrying the connection attempts (possibly, with an exponential backoff).

PR #102, opened 2019-12-11
Retry connection when initial attempt fails

Fabian-Sommer:
Close issue #90

When connecting to a reader fails, we retry connecting later.
This is done by showing another Bus number for the device.
We need to fake Bus numbers anyway as chrome.usb does not let us retrieve them.
PCSC then views it as a different device from the previous one and attempts
to connect to it once a smartcard reader needs to be accessed.

### DIFF
--- a/third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
+++ b/third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
@@ -23,7 +23,6 @@
 #include <cstring>
 #include <functional>
 #include <limits>
-#include <unordered_map>
 #include <utility>
 #include <vector>
 
@@ -508,8 +507,8 @@ int LibusbOverChromeUsb::LibusbGetDeviceDescriptor(
 
 uint8_t LibusbOverChromeUsb::LibusbGetBusNumber(libusb_device* dev) {
   auto bus_numbers_iterator = 
-      busNumbers_.find(dev->chrome_usb_device().device);
-  if (bus_numbers_iterator != busNumbers_.end()) {
+      bus_numbers_.find(dev->chrome_usb_device().device);
+  if (bus_numbers_iterator != bus_numbers_.end()) {
     return bus_numbers_iterator->second;
   }
   return kDefaultBusNumber;
@@ -538,9 +537,10 @@ int LibusbOverChromeUsb::LibusbOpen(
         "request failed: " << result.error_message();
     // Modify the devices (fake) bus number that we report so that PCSC will 
     // retry to connect to the device once it updates the device list.
-    uint32_t new_bus_number = LibusbGetBusNumber(dev) + 1;
+    uint32_t new_bus_number =
+        static_cast<uint32_t>(LibusbGetBusNumber(dev)) + 1;
     if (new_bus_number <= kMaximumBusNumber) {
-      busNumbers_[dev->chrome_usb_device().device] = new_bus_number;
+      bus_numbers_[dev->chrome_usb_device().device] = new_bus_number;
     }
     return LIBUSB_ERROR_OTHER;
   }
--- a/third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
+++ b/third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
@@ -34,14 +34,15 @@
 #include <google_smart_card_common/pp_var_utils/extraction.h>
 
 //
-// We use stubs for the device bus number (as the chrome.usb API does not provide means of 
-// retrieving it). We modify this for a device when opening the device fails.
-// This makes PCSC recognize it as a new device which causes PCSC to retry opening it.
+// We use stubs for the device bus number (as the chrome.usb API does not 
+// provide means of retrieving it). We modify this for a device when opening
+// the device fails. This makes PCSC recognize it as a new device which causes
+// PCSC to retry opening it. The number of reconnection attempts is limited
+// by kMaximumBusNumber - kDefaultBusNumber.
 //
 
-std::unordered_map<int64_t, uint8_t> kBusNumbers;
 const uint8_t kDefaultBusNumber = 1;
-const uint8_t kMaximumReconnectAttempts = 64;
+const uint8_t kMaximumBusNumber = 64;
 
 //
 // Bit mask values for the bmAttributes field of the libusb_config_descriptor
@@ -506,8 +507,9 @@ int LibusbOverChromeUsb::LibusbGetDeviceDescriptor(
 }
 
 uint8_t LibusbOverChromeUsb::LibusbGetBusNumber(libusb_device* dev) {
-  if (kBusNumbers.find(dev->chrome_usb_device().device) != kBusNumbers.end()) {
-    return kBusNumbers[dev->chrome_usb_device().device];
+  auto busNumbersIterator = busNumbers_.find(dev->chrome_usb_device().device);
+  if (busNumbersIterator != busNumbers_.end()) {
+    return busNumbersIterator->second;
   }
   return kDefaultBusNumber;
 }
@@ -533,16 +535,11 @@ int LibusbOverChromeUsb::LibusbOpen(
   if (!result.is_successful()) {
     GOOGLE_SMART_CARD_LOG_WARNING << "LibusbOverChromeUsb::LibusbOpen " <<
         "request failed: " << result.error_message();
-    // modify the devices (fake) bus number that we report so that PCSC will 
-    // retry to connect to the device once it updates the device list
-    if (kBusNumbers.find(dev->chrome_usb_device().device) != 
-        kBusNumbers.end()) {
-      kBusNumbers[dev->chrome_usb_device().device] = std::min(
-          static_cast<uint8_t>(
-          kBusNumbers[dev->chrome_usb_device().device] + 1),
-          kMaximumReconnectAttempts);
-    } else {
-      kBusNumbers[dev->chrome_usb_device().device] = kDefaultBusNumber + 1;
+    // Modify the devices (fake) bus number that we report so that PCSC will 
+    // retry to connect to the device once it updates the device list.
+    uint32_t newBusNumber = LibusbGetBusNumber(dev) + 1;
+    if (newBusNumber <= kMaximumBusNumber) {
+      busNumbers_[dev->chrome_usb_device().device] = newBusNumber;
     }
     return LIBUSB_ERROR_OTHER;
   }
--- a/third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
+++ b/third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
@@ -23,6 +23,7 @@
 #include <cstring>
 #include <functional>
 #include <limits>
+#include <unordered_map>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
```diff
@@ -32,9 +33,15 @@
...
-// This arbitrary chosen constant is used as a stub for the device bus number
-// (as the chrome.usb API does not provide means of retrieving this).
-const uint8_t kFakeDeviceBusNumber = 42;
+//
+// We use stubs for the device bus number (as the chrome.usb API does not provide means of 
+// retrieving it). We modify this for a device when opening the device fails.
+// This makes PCSC recognize it as a new device which causes PCSC to retry opening it.
+//
+
+std::unordered_map<int64_t, uint8_t> kBusNumbers;
```
emaxx-google: The style guide forbids global objects with nontrivial destructors: https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables
Can we make it a member of the LibusbOverChromeUsb?

Also the name looks like if it were a constant, while it's not.

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
```diff
@@ -498,8 +505,11 @@ int LibusbOverChromeUsb::LibusbGetDeviceDescriptor(
   return LIBUSB_SUCCESS;
 }
 
-uint8_t LibusbOverChromeUsb::LibusbGetBusNumber(libusb_device* /*dev*/) {
-  return kFakeDeviceBusNumber;
+uint8_t LibusbOverChromeUsb::LibusbGetBusNumber(libusb_device* dev) {
+  if (kBusNumbers.find(dev->chrome_usb_device().device) != kBusNumbers.end()) {
```
emaxx-google: nit: Let's use an iterator to avoid searching in the map twice.

(Not only because it's unnecessary, but also to avoid the possibility of a bug that someone changes one place but not the other.)

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
```diff
@@ -523,6 +533,17 @@ int LibusbOverChromeUsb::LibusbOpen(
   if (!result.is_successful()) {
     GOOGLE_SMART_CARD_LOG_WARNING << "LibusbOverChromeUsb::LibusbOpen " <<
         "request failed: " << result.error_message();
+    // modify the devices (fake) bus number that we report so that PCSC will
```
emaxx-google: tiny nit: Please capitalize the first word and end the comment with a period.

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
```diff
@@ -523,6 +533,17 @@ int LibusbOverChromeUsb::LibusbOpen(
...
+    // modify the devices (fake) bus number that we report so that PCSC will 
+    // retry to connect to the device once it updates the device list
+    if (kBusNumbers.find(dev->chrome_usb_device().device) != 
+        kBusNumbers.end()) {
+      kBusNumbers[dev->chrome_usb_device().device] = std::min(
+          static_cast<uint8_t>(
+          kBusNumbers[dev->chrome_usb_device().device] + 1),
+          kMaximumReconnectAttempts);
+    } else {
+      kBusNumbers[dev->chrome_usb_device().device] = kDefaultBusNumber + 1;
```
emaxx-google: nit: Similar to above, let's have a single line that performs the update in the map, and calculate the bus number in advance.

Or, actually, can we use LibusbGetDeviceAddress() here? Then we won't have two different places that hardcode kDefaultBusNumber.

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
```diff
@@ -523,6 +533,17 @@ int LibusbOverChromeUsb::LibusbOpen(
...
     GOOGLE_SMART_CARD_LOG_WARNING << "LibusbOverChromeUsb::LibusbOpen " <<
         "request failed: " << result.error_message();
+    // modify the devices (fake) bus number that we report so that PCSC will 
+    // retry to connect to the device once it updates the device list
+    if (kBusNumbers.find(dev->chrome_usb_device().device) != 
+        kBusNumbers.end()) {
+      kBusNumbers[dev->chrome_usb_device().device] = std::min(
+          static_cast<uint8_t>(
+          kBusNumbers[dev->chrome_usb_device().device] + 1),
+          kMaximumReconnectAttempts);
```
emaxx-google: This is a bit incorrect: note that the first argument of std::min() is the sum of the default bus number and the reconnect attempt count.

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
```diff
@@ -523,6 +533,17 @@ int LibusbOverChromeUsb::LibusbOpen(
   if (!result.is_successful()) {
     GOOGLE_SMART_CARD_LOG_WARNING << "LibusbOverChromeUsb::LibusbOpen " <<
         "request failed: " << result.error_message();
+    // modify the devices (fake) bus number that we report so that PCSC will 
+    // retry to connect to the device once it updates the device list
+    if (kBusNumbers.find(dev->chrome_usb_device().device) != 
+        kBusNumbers.end()) {
+      kBusNumbers[dev->chrome_usb_device().device] = std::min(
+          static_cast<uint8_t>(
+          kBusNumbers[dev->chrome_usb_device().device] + 1),
```
emaxx-google: This looks a bit dangerous regarding the potential integer overflows. I understand that the current code seems to be safe, but rewriting this with an explicit "if" would make it more clear.

Also, BTW, the indentation was missing.

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
```diff
@@ -523,6 +533,17 @@ int LibusbOverChromeUsb::LibusbOpen(
...
+          kMaximumReconnectAttempts);
+    } else {
+      kBusNumbers[dev->chrome_usb_device().device] = kDefaultBusNumber + 1;
+    }
```
emaxx-google: Just thinking out loud: Probably we could extract this custom logic out of the LibusbOverChromeUsb class, so that the latter would behave in a standard way by default, and only a separate "hook" class would be performing our PC/SC-Lite specific logic.

This would be nice to do, but I guess we can skip this if this makes the things simpler.

The main benefit would be that removing these hacks would allow to directly reuse our libusb port in other projects. But I don't know if anyone uses it today, so we can postpone this consideration till later.

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
Fabian-Sommer: Done

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
Fabian-Sommer: Done

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
Fabian-Sommer: Done

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
Fabian-Sommer: Done

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
Fabian-Sommer: Done

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
Fabian-Sommer: Done

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
```diff
@@ -23,6 +23,7 @@
 #include <cstring>
 #include <functional>
 #include <limits>
+#include <unordered_map>
```
emaxx-google: nit: Move this into the .h file.

third_party/libusb/naclport/src/libusb_over_chrome_usb.h
```diff
@@ -139,6 +139,7 @@ class LibusbOverChromeUsb final : public LibusbInterface {
   int LibusbHandleEventsWithTimeout(
       libusb_context* context, int timeout_seconds);
 
+  std::unordered_map<int64_t, uint8_t> busNumbers_;
```
emaxx-google: nit: "bus_numbers_" (see https://google.github.io/styleguide/cppguide.html#Variable_Names)

Please also add a comment explaining what's the key and the value in this map.

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
```diff
@@ -523,6 +535,12 @@ int LibusbOverChromeUsb::LibusbOpen(
   if (!result.is_successful()) {
     GOOGLE_SMART_CARD_LOG_WARNING << "LibusbOverChromeUsb::LibusbOpen " <<
         "request failed: " << result.error_message();
+    // Modify the devices (fake) bus number that we report so that PCSC will 
+    // retry to connect to the device once it updates the device list.
+    uint32_t newBusNumber = LibusbGetBusNumber(dev) + 1;
```
emaxx-google: nit: Let's do the explicit cast:

`   static_cast<uint32_t>(LibusbGetBusNumber(dev))`

, since relying on the implicit integral promotion can be tricky to notice.

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
```diff
@@ -498,8 +506,12 @@ int LibusbOverChromeUsb::LibusbGetDeviceDescriptor(
   return LIBUSB_SUCCESS;
 }
 
-uint8_t LibusbOverChromeUsb::LibusbGetBusNumber(libusb_device* /*dev*/) {
-  return kFakeDeviceBusNumber;
+uint8_t LibusbOverChromeUsb::LibusbGetBusNumber(libusb_device* dev) {
+  auto busNumbersIterator = busNumbers_.find(dev->chrome_usb_device().device);
```
emaxx-google: nit: "bus_numbers_iterator"

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
emaxx-google: nit: "new_bus_number"

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
Fabian-Sommer: Done

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
Fabian-Sommer: Done

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
Fabian-Sommer: Done

third_party/libusb/naclport/src/libusb_over_chrome_usb.h
Fabian-Sommer: Done

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
Fabian-Sommer: Done

PR #107, opened 2020-04-18
Add tests for ChromeLoginStateHook ready promise

emaxx-google:


third_party/libusb/naclport/src/chrome_usb/chrome-login-state-hook-unittest.js
```diff
@@ -122,15 +126,26 @@ function changeSessionState(fakeNewSessionState, propertyReplacer) {
   });
 }
 
+goog.exportSymbol(
+    'test_ChromeLoginStateHook_NoApi', function() {
+      const loginStateHook = new GSC.Libusb.ChromeLoginStateHook();
+      // Expect that getHookReadyPromise() gets rejected.
```
Fabian-Sommer, line 132: nit: Maybe add that rejection is expected since the API doesn't get mocked/set up here.

PR #111, opened 2020-05-18
Implement message channel for example app UI

emaxx-google:
Extend the C++ example code with the boilerplate that allows
to send messages between the UI JavaScript code and the
NaCl module.

Also, to demonstrate the usage of the message channel,
implement the "Run test" button to trigger some trivial
sequence of PC/SC method calls.

### DIFF
--- a/example_cpp_smart_card_client_app/src/ui_bridge.h
+++ b/example_cpp_smart_card_client_app/src/ui_bridge.h
@@ -64,6 +64,10 @@ class UiBridge final : public google_smart_card::TypedMessageListener {
   // Note that if the UI is currently closed, the message is silently discarded.
   void SendMessageToUi(const pp::Var& message);
 
+  // google_smart_card::TypedMessageListener:
+  std::string GetListenedMessageType() const override;
+  bool OnTypedMessageReceived(const pp::Var& data) override;
+
  private:
   struct AttachedState {
     AttachedState(pp::Instance* pp_instance,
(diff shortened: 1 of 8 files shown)


example_cpp_smart_card_client_app/src/window.js
```diff
@@ -47,11 +56,45 @@ function onCloseWindowClicked(e) {
...
+function onRunTestClicked(e) {
+  e.preventDefault();
+  naclModuleMessageChannel.send('ui_backend', {command: 'run_test'});
+}
+
+// Called when a "output_message" message is received from the NaCl module. Adds
+// a line into the "output" <pre> element in the UI.
+function displayOutputMessage(text) {
+  const outputElem = goog.dom.getElement('output');
+  outputElem.innerHTML = text + '\n' + outputElem.innerHTML;
```
Fabian-Sommer, line 70: Without seeing how it looks in practice, based on various chat examples that come to my mind, I would expect that messages should be appended and not prepended. Of course, this only works if there is scrolling etc. when there are too many messages to fit the window.

example_cpp_smart_card_client_app/src/ui_bridge.h
```diff
@@ -0,0 +1,91 @@
...
+      : pp_instance(pp_instance),
+        typed_message_router(typed_message_router) {}
+
+    pp::Instance* const pp_instance;
+    google_smart_card::TypedMessageRouter* const typed_message_router;
+  };
+
+  // google_smart_card::TypedMessageListener:
+  std::string GetListenedMessageType() const override;
+  bool OnTypedMessageReceived(const pp::Var& data) override;
```
Fabian-Sommer, line 80: These methods are defined as public in the base class, so they probably should be here, too.

example_cpp_smart_card_client_app/src/window.js
emaxx-google: Done (there's no scrolling, but we don't care about it too much, as it's just an example without real styling).

example_cpp_smart_card_client_app/src/ui_bridge.h
emaxx-google: Done.

emaxx-google:
Thanks, Fabian, PTAL.

PR #112, opened 2020-05-18
Sequential UI commands in Example App

emaxx-google:
Make the handler of the UI messages use the same mutex as the existing
chrome.certificateProvider message handler, so that they never run
concurrently. This corresponds to the typical single-threaded model of
real-world smart card middleware applications.

Also flip the default in the Example App to make the
chrome.certificateProvider message handler sequential (it was previously
controlled by a boolean flag set to false by default).

Implementation-wise, a cleaner approach than to use mutexes would be to
have a dedicated worker thread that would be used for running all
handler jobs, but it'd be a much bigger change to introduce such a
pattern into the code.


emaxx-google:
PTAL. That's the last planned change for this UI thing.

PR #116, opened 2020-06-04
Import webports repository as third-party dep

emaxx-google:
Hardcode the third-party dependency - the "webports" repository - as a
directory in our source tree.

This will allow making local modifications in follow-up commits.


emaxx-google:
Please review only the "third_party/webports/README.google" file - the rest is just committed from "webports" as-is.

third_party/webports/README.google
```diff
@@ -0,0 +1,7 @@
+URL: https://chromium.googlesource.com/webports.git/+/d866724703e02ecf3e63040d652c2a87b98896f9
+Version: d866724703e02ecf3e63040d652c2a87b98896f9
+License: BSD License 2.0
+License File: LICENSE
+Description:
+Collection of open source libraries and applications that have been ported to
+Native Client, along with set to tools for building and maintaining them..
```
Fabian-Sommer, line 7: nit: there are two . at the end

emaxx-google: Done

PR #119, opened 2020-06-05
Workaround for NaCl init issue on desktops

emaxx-google:
This makes using the "make run"/"make test" commands slightly more
convenient for development/testing on a desktop OS, by reusing the same
Chrome profile directory (--user-data-dir) for all executions.

This partially solves the trouble caused by the asynchronous
installation of the NaCl component in desktop Chrome browsers, which was
manifesting in form of the following error on every run:

  NaCl module load failed: The Portable Native Client (pnacl) component
  is not installed.

While this commit doesn't fix the underlying issue per se, it makes this
error appear only on the first run, therefore streamlining the
developer experience.


emaxx-google:
Thanks for the quick review!

PR #127, opened 2020-06-18
Add help button

ayaNader:
Added a help button that redirects to the Help Article in a new tab.

### DIFF
--- a/smart_card_connector_app/src/manifest.json.template
+++ b/smart_card_connector_app/src/manifest.json.template
@@ -24,6 +24,7 @@
     "alwaysOnTopWindows",
     "loginState",
     "usb",
+    "browser",
     {
       "usbDevices": [
 ${usb_devices}
--- a/smart_card_connector_app/src/window-help-showing.js
+++ b/smart_card_connector_app/src/window-help-showing.js
@@ -1,5 +1,5 @@
 /** @license
- * Copyright 2016 Google Inc. All Rights Reserved.
+ * Copyright 2020 Google Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


smart_card_connector_app/src/window-help-showing.js
```diff
@@ -0,0 +1,65 @@
...
+var HELP_WINDOW_URL = 'https://support.google.com/chrome/a/answer/7014689';
+
+/** @const */
+var GSC = GoogleSmartCard;
+
+/**
+ * @type {!Element}
+ * @const
+ */
+var openHelpElement = /** @type {!Element} */ (goog.dom.getElement(
```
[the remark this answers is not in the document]
ayaNader: actually the inline is the necessary one, it gives an error when removed

smart_card_connector_app/src/manifest.json.template
```diff
@@ -24,6 +24,7 @@
     "alwaysOnTopWindows",
     "loginState",
     "usb",
+    "browser",
```
emaxx-google, line 27: nit: Let's place it between "alwaysOnTopWindows" and "loginState", in order to keep the list sorted.

smart_card_connector_app/src/window-main.js
```diff
@@ -24,6 +24,7 @@ goog.require('GoogleSmartCard.ConnectorApp.Window.AboutShowing');
 goog.require('GoogleSmartCard.ConnectorApp.Window.AppsDisplaying');
 goog.require('GoogleSmartCard.ConnectorApp.Window.DevicesDisplaying');
 goog.require('GoogleSmartCard.ConnectorApp.Window.LogsExporting');
+goog.require('GoogleSmartCard.ConnectorApp.Window.HelpShowing');
```
emaxx-google, line 27: nit: Move this to be between "DevicesDisplaying" and "LogsExporting", to keep the list sorted.

(BTW, hmm, I thought that we have a diagnostics enabled in the Closure Compiler that throws an error when the order is violated. Maybe it stopped working though.)

Issue #104, opened 2020-01-20
UI "add reader" button looks confusing


emaxx-google:
Two things that, according to feedback from users, are confusing in our UI:

1. The "add reader" button may be misinterpreted as an invitation for users to click on it. However, users actually don't need to click it if their reader is already displayed (which should happen for 99% of users).

2. After clicking "add reader", the UI says "No compatible devices found" even when there's a connected smart card reader that was already added before. This confuses as the users may think that their reader isn't recognized.

emaxx-google:
As a quick idea, for item 1 we could rename the button to "add another reader". Addressing item 2 is harder, since the whole dialog with the text "No compatible devices found" is rendered by Chrome itself and doesn't allow to customize these errors.

PR #135, opened 2020-06-28
Change text on Add Reader button depending on list being empty or not

ayaNader:
Gives the add-device button text value of "Add Reader" when no readers are in the list, and gives it the value "Add Another Reader" when 1 or more readers are in the list.
This is to prevent confusion as mentioned in issue #104.

### DIFF
--- a/common/js/src/i18n.js
+++ b/common/js/src/i18n.js
@@ -75,52 +75,55 @@ function transformAllElements(attribute, transformFunction) {
   }
 }
 
-function transformFunctionI18nDataAttribute(element, translatedText) {
+/**
+ * @param {!Element} element 
+ * @param {string} translatedText 
+ */
+function applyTranslation(element, translatedText) {
   logger.fine('Translating element.textContent [' + element.outerHTML +
               '] to "' + translatedText + '"');
   element.textContent = translatedText;
 }
 
-function transformFunctionI18nDataAriaLabelAttribute(element, translatedText) {
+/**
+ * @param {!Element} element 
+ * @param {string} translatedText 
+ */
+function applyAriaLabelTranslation(element, translatedText) {
   logger.fine('Translating element.aria-label [' + element.outerHTML +
               '] to "' + translatedText + '"');
   element.setAttribute('aria-label', translatedText);
 }
 
 /** 
- * Takes the single element passed, replacing element.textContent with translation
- * if it conatins I18N_DATA_ATTRIBUTE, and setting aria-label to translation
- * if it contains I18N_DATA_ARIA_LABEL_ATTRIBUTE.
-*/
+ * Takes the element passed, replacing element.textContent with 
+ * translation if it contains I18N_DATA_ATTRIBUTE, and setting aria-label to 
+ * translation if it contains I18N_DATA_ARIA_LABEL_ATTRIBUTE.
+ * @param {!Element} element
+ */
 GSC.I18n.adjustElementTranslation = function(element) {
-  if (goog.dom.dataset.has(element, 'i18n')) {
+  if (element.hasAttribute(I18N_DATA_ATTRIBUTE)) {
     transformElement(
-      element, I18N_DATA_ATTRIBUTE, 
-      transformFunctionI18nDataAttribute);
+        element, I18N_DATA_ATTRIBUTE, applyTranslation);
   }
 
-  if(goog.dom.dataset.has(element, 'i18nAriaLabel')) {
+  if (element.hasAttribute(I18N_DATA_ARIA_LABEL_ATTRIBUTE)) {
     transformElement(
-      element, I18N_DATA_ARIA_LABEL_ATTRIBUTE, 
-      transformFunctionI18nDataAriaLabelAttribute);
+        element, I18N_DATA_ARIA_LABEL_ATTRIBUTE, applyAriaLabelTranslation);
   }
 };
 
 /**
- * Goes through all the HTML elements in the current window, replacing
- * element.textContent with translation for elements containing attribute
- * I18N_DATA_ATTRIBUTE, and setting attribute aria-label to translation for
- * elements containing attribute I18N_DATA_ARIA_LABEL_ATTRIBUTE. Used for
- * translation that plays nice with Chromevox (accessibility tool).
+ * Applies adjustElementTranslation() to all HTML elements in the current 
+ * document. Used for translation that plays nice with Chromevox (accessibility
+ * tool).
  */
 GSC.I18n.adjustAllElementsTranslation = function() {
   transformAllElements(
-      I18N_DATA_ATTRIBUTE,
-      transformFunctionI18nDataAttribute);
+      I18N_DATA_ATTRIBUTE, applyTranslation);
 
   transformAllElements(
-      I18N_DATA_ARIA_LABEL_ATTRIBUTE,
-      transformFunctionI18nDataAriaLabelAttribute);
+      I18N_DATA_ARIA_LABEL_ATTRIBUTE, applyAriaLabelTranslation);
 };
 
 });  // goog.scope
--- a/common/js/src/i18n.js
+++ b/common/js/src/i18n.js
@@ -22,7 +22,6 @@
 goog.provide('GoogleSmartCard.I18n');
 
 goog.require('GoogleSmartCard.Logging');
-goog.require('goog.dom');
 goog.require('goog.log.Logger');
 
 goog.scope(function() {
--- a/common/js/src/i18n.js
+++ b/common/js/src/i18n.js
@@ -22,6 +22,7 @@
 goog.provide('GoogleSmartCard.I18n');
 
 goog.require('GoogleSmartCard.Logging');
+goog.require('goog.dom');
 goog.require('goog.log.Logger');
 
 goog.scope(function() {
@@ -74,6 +75,37 @@ function transformAllElements(attribute, transformFunction) {
   }
 }
 
+function transformFunctionI18nDataAttribute(element, translatedText) {
+  logger.fine('Translating element.textContent [' + element.outerHTML +
+              '] to "' + translatedText + '"');
+  element.textContent = translatedText;
+}
+
+function transformFunctionI18nDataAriaLabelAttribute(element, translatedText) {
+  logger.fine('Translating element.aria-label [' + element.outerHTML +
+              '] to "' + translatedText + '"');
+  element.setAttribute('aria-label', translatedText);
+}
+
+/** 
+ * Takes the single element passed, replacing element.textContent with translation
+ * if it conatins I18N_DATA_ATTRIBUTE, and setting aria-label to translation
+ * if it contains I18N_DATA_ARIA_LABEL_ATTRIBUTE.
+*/
+GSC.I18n.adjustElementTranslation = function(element) {
+  if (goog.dom.dataset.has(element, 'i18n')) {
+    transformElement(
+      element, I18N_DATA_ATTRIBUTE, 
+      transformFunctionI18nDataAttribute);
+  }
+
+  if(goog.dom.dataset.has(element, 'i18nAriaLabel')) {
+    transformElement(
+      element, I18N_DATA_ARIA_LABEL_ATTRIBUTE, 
+      transformFunctionI18nDataAriaLabelAttribute);
+  }
+};
+
 /**
  * Goes through all the HTML elements in the current window, replacing
  * element.textContent with translation for elements containing attribute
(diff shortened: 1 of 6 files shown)


smart_card_connector_app/src/_locales/en/messages.json
```diff
@@ -35,6 +35,10 @@
     "message": "ADD READER",
     "description": "Title of the button that can be used to open a dialog for adding a new Smart Card reader"
```
Fabian-Sommer, line 36: nit: I would make it explicit here that this only appears when there are no readers yet. Generally, these descriptions exist to give context to the translators, so giving more details is helpful, and this info might make a difference for the translation in some languages.

smart_card_connector_app/src/_locales/en/messages.json
```diff
@@ -35,6 +35,10 @@
...
   },
+  "addAnotherDevice": {
+    "message": "ADD ANOTHER READER",
+    "description": "Title of the button that can be used to open a dialog for adding another Smart Card reader"
```
Fabian-Sommer, line 40: nit: As above, I would make it explicit that this applies only when there is at least one reader already.

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -83,24 +85,25 @@ function displayReaderList(readers) {
...
-                           reader['status'];
+    reader['status'];
     if (reader['status'] == GSC.PcscLiteServer.ReaderStatus.SUCCESS &&
-        reader['isCardPresent']) {
+    reader['isCardPresent']) {
       indicatorClasses = 'reader-card-present-indicator';
     }
     var indicator = goog.dom.createDom('span', indicatorClasses);
-
+
```
Fabian-Sommer, line 96: nit: I think these changes should be reversed - lines should not end with whitespace, and if a statement is not completed in one line the following line should start with at least 4 more space characters than the line before it - often as much that it starts at the same height as the statement that it continues (so the way it was before looks good)

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -116,6 +119,15 @@ function makeReaderNameForDisplaying(readerName) {
   return readerName;
 }
 
+function updateAddDeviceButtonText(readersCount){
+  if(readersCount == 0){
+    addDeviceElement.dataset.i18n = 'addDevice';
+  }
+  else{
```
Fabian-Sommer, line 126: nit: put these in one line and with a space between a bracket and "else":
"} else {"

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -116,6 +119,15 @@ function makeReaderNameForDisplaying(readerName) {
   return readerName;
 }
 
+function updateAddDeviceButtonText(readersCount){
+  if(readersCount == 0){
```
Fabian-Sommer, line 123: nit: Add a space between "if" and "(" and between ")" and "{"

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -83,24 +85,25 @@ function displayReaderList(readers) {
...
     var indicatorContainer = goog.dom.createDom(
-        'span', 'reader-indicator-container', indicator);
-
-    var text = makeReaderNameForDisplaying(reader['name']) +
-        (reader['error'] ? ' (Error ' + reader['error'] + ')' : '');
-
-    var element = goog.dom.createDom('li', undefined, indicatorContainer, text);
-    goog.dom.append(readersListElement, element);
-  }
+      'span', 'reader-indicator-container', indicator);
```
Fabian-Sommer, line 98: nit: If I counted correctly, there should be at least two more spaces in front here.

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -83,24 +85,25 @@ function displayReaderList(readers) {
...
+      
+      var text = makeReaderNameForDisplaying(reader['name']) +
+      (reader['error'] ? ' (Error ' + reader['error'] + ')' : '');
```
Fabian-Sommer, line 101: nit: More space in front here

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -116,6 +119,15 @@ function makeReaderNameForDisplaying(readerName) {
...
 }
 
+function updateAddDeviceButtonText(readersCount){
+  if(readersCount == 0){
+    addDeviceElement.dataset.i18n = 'addDevice';
+  }
+  else{
+    addDeviceElement.dataset.i18n = 'addAnotherDevice';
+  }
+  GSC.I18n.adjustElementsTranslation();
```
emaxx-google, line 129: As discussed offline, it'd be great if we only applied this to the |addDeviceElement|, rather than to the whole page. (The real-world performance gain would be negligible, but it'd still be nice to avoid unnecessary operations.)

For this, we could have a new GSC.I18n.adjustElementTranslation(element) function that calls transformElement().

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -83,24 +85,25 @@ function displayReaderList(readers) {
...
-  }
+      'span', 'reader-indicator-container', indicator);
+      
+      var text = makeReaderNameForDisplaying(reader['name']) +
+      (reader['error'] ? ' (Error ' + reader['error'] + ')' : '');
+      
+      var element = goog.dom.createDom('li', undefined, indicatorContainer, text);
+      goog.dom.append(readersListElement, element);
+    }
+    updateAddDeviceButtonText(readers.length);
```
emaxx-google, line 106: As discussed offline, let's, indeed, move this new call out from displayReaderList(), since it's a bit confusing that displayReaderList() not only updates the list, but also modifies the button title.

For instance, you could add a new function "onReadersChanged()" that would be passed to readerTrackerSubscriber() instead of displayReaderList(). onReadersChanged() would call displayReaderList() and updateAddDeviceButtonText().

smart_card_connector_app/src/window-devices-displaying.js
ayaNader: I was thinking we name it GSC.I18n.adjustSingleElementTranslation(element) to avoid confusion with GSC.I18n.adjustElementsTranslation(), since the difference would be only one letter?

smart_card_connector_app/src/window-devices-displaying.js
ayaNader: not exactly sure i understand you correctly, i didn't change any code in this part.
are you referring to the extra spaces on line 88 and 96?

smart_card_connector_app/src/window-devices-displaying.js
ayaNader: There is an issue with the spacing yeah, but I think it's the opposite, that lines 101-103 should have 2 spaces less.

smart_card_connector_app/src/window-devices-displaying.js
ayaNader: Will do.

smart_card_connector_app/src/window-devices-displaying.js
emaxx-google: From that perspective, I'd prefer we rather rename the old function to be "adjustAllElementsTranslation", and have the new one named just "adjustElementTranslation".

ayaNader: Ok sure, is this a gut feeling, or am I missing something regarding naming conventions?

emaxx-google: There's no hard rule. I was just thinking along these lines: "do something for one thing" is a more basic primitive that probably deserves a shorter and a simpler name, and "do something for a bunch of things" is its derivative and can have a longer name to be clearly distinguishable. Not insisting, and happy to know your opinion too.

P.S. https://martinfowler.com/bliki/TwoHardThings.html :-)

ayaNader: Ok yeah that makes sense

common/js/src/i18n.js
```diff
@@ -74,27 +75,52 @@ function transformAllElements(attribute, transformFunction) {
...
+  element.setAttribute('aria-label', translatedText);
+}
+
+/** 
+ * Takes the single element passed, replacing element.textContent with translation
+ * if it conatins I18N_DATA_ATTRIBUTE, and setting aria-label to translation
+ * if it contains I18N_DATA_ARIA_LABEL_ATTRIBUTE.
+*/
+GSC.I18n.adjustElementTranslation = function(element) {
+  if (goog.dom.dataset.has(element, 'i18n')) {
```
emaxx-google, line 96: nit: I'd suggest using `.hasAttribute(I18N_DATA_ATTRIBUTE)` instead, since not using the constant here kinda defeats its purpose.

Ditto in line 102.

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -72,6 +74,11 @@ var readersListElement = /** @type {!Element} */ (goog.dom.getElement(
 var addDeviceElement = /** @type {!Element} */ (goog.dom.getElement(
     'add-device'));
 
+function onReadersChanged(readers) {
```
emaxx-google, line 77: nit: Ditto `@param`.

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -85,21 +92,21 @@ function displayReaderList(readers) {
     goog.asserts.assert(readersListElement);
 
     var indicatorClasses = 'reader-state-indicator reader-state-indicator-' +
-                           reader['status'];
+                            reader['status'];
```
emaxx-google, line 95: nit: There's still some unwanted indentation change in this line. I don't think the code in master has an error here - the `reader` word should be indented with the `'` character, since it's the beginning of the expression.

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -85,21 +92,21 @@ function displayReaderList(readers) {
...
     if (reader['status'] == GSC.PcscLiteServer.ReaderStatus.SUCCESS &&
         reader['isCardPresent']) {
       indicatorClasses = 'reader-card-present-indicator';
     }
     var indicator = goog.dom.createDom('span', indicatorClasses);
 
     var indicatorContainer = goog.dom.createDom(
-        'span', 'reader-indicator-container', indicator);
+      'span', 'reader-indicator-container', indicator);
```
emaxx-google, line 103: nit: Ditto for unwanted indentation change.

Same in line 106.

See https://google.github.io/styleguide/jsguide.html#formatting-indent

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -85,21 +92,21 @@ function displayReaderList(readers) {
...
 
     var text = makeReaderNameForDisplaying(reader['name']) +
-        (reader['error'] ? ' (Error ' + reader['error'] + ')' : '');
+      (reader['error'] ? ' (Error ' + reader['error'] + ')' : '');
 
     var element = goog.dom.createDom('li', undefined, indicatorContainer, text);
-    goog.dom.append(readersListElement, element);
+      goog.dom.append(readersListElement, element);
```
emaxx-google, line 109: nit: Ditto for unwanted indentation change - these two extra spaces aren't needed.

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -116,6 +123,15 @@ function makeReaderNameForDisplaying(readerName) {
   return readerName;
 }
 
+function updateAddDeviceButtonText(readersCount){
```
emaxx-google, line 126: nit: Ditto `@param`.

emaxx-google, line 126: nit: Add a space before `{`.

See the example formatting in the guideline: https://google.github.io/styleguide/jsguide.html#formatting

common/js/src/i18n.js
```diff
@@ -74,27 +75,52 @@ function transformAllElements(attribute, transformFunction) {
   }
 }
 
+function transformFunctionI18nDataAttribute(element, translatedText) {
```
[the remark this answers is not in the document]
ayaNader: good point, so would "applyDataAriaLabelAttrTranslation" also be too long?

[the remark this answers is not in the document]
emaxx-google: It's at least shorter than the old name. Maybe we can name them simply "applyTranslation", "applyAriaLabelTranslation".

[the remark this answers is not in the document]
ayaNader: yeah i think that's better

smart_card_connector_app/src/window-devices-displaying.js
ayaNader: oh ok i thought it should align with the beginning of the word

PR #137, opened 2020-07-06
Add tests for i18n JS helpers

emaxx-google:


common/js/src/i18n-unittest.js
```diff
@@ -0,0 +1,103 @@
...
+
+goog.scope(function() {
+
+const GSC = GoogleSmartCard;
+
+const propertyReplacer = new goog.testing.PropertyReplacer;
+/** @type {Element?} */
+let pNode;
+/** @type {Element?} */
+let p2Node;
```
Fabian-Sommer, line 32: nit: I would rename this to pNode2 as it is another node of type p - p2Node could be misunderstood that it is a node <p2>. Optionally, pNode could also be renamed to pNode1 for consistency.

emaxx-google:
Thanks, PTAL.

PR #138, opened 2020-07-08
Add tooltips for the reader status indicators

### LINKED ISSUE #48: Add tooltips for the reader status indicators
The meaning of the indicator color (which can be red, yellow and green) is not absolutely clear from the user perspective.

ayaNader:
Tooltips for the reader status indicators.
Closes #48.


smart_card_connector_app/src/window.css
```diff
@@ -175,3 +175,34 @@ body {
...
+  padding: 2px 5px 0px 5px;
+  position: absolute;
+  top: 150%;
+  text-align: center;
+  visibility: hidden;
+  width: max-content;
+}
+
+.tool-tip-init::after {
+  content: "Initializing card...";
```
Fabian-Sommer, line 195: I am not a fan of having the content set by CSS - all content should come from HTML/JS side. CSS should only be concerned with styling the content but not with any concrete strings, just as how HTML should not use any style tags. 
Otherwise, the line between HTML and CSS becomes blurry, and with this approach, i18n also becomes impossible.
Just put the strings that should be displayed as part of the element that you create with JS.

smart_card_connector_app/src/window.css
```diff
@@ -175,3 +175,34 @@ body {
   top: -2px;
   width: 20px;
 }
+
+.tool-tip {
```
emaxx-google, line 179: I see, so this is a hand-written tooltip. Can we use the standard one? I think it's the "title" attribute: https://www.w3schools.com/tags/att_title.asp

The benefit of the standard tooltip is that we can just rely on the browser's implementation of this feature without having to maintain our own implementation. Also the browser's standard implementation is more coherent with the typical UX on the target platform: i.e., how the tooltip is styled, when it appears (usually tooltips are only displayed after some timeout), etc.

smart_card_connector_app/src/window.css
ayaNader: Makes sense. Thanks for that!

smart_card_connector_app/src/window.css
ayaNader: Alright that's better, thanks for the info

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -105,6 +105,17 @@ function displayReaderList(readers) {
     var indicatorContainer = goog.dom.createDom(
         'span', 'reader-indicator-container', indicator);
 
+    var toolTipText = '';
+    if (reader['status'] == GSC.PcscLiteServer.ReaderStatus.INIT) {
```
emaxx-google, line 109: nit: I think it's a good case where "switch" can be used. Also perhaps let's extract this into a separate function, so that the displayReaderList() function doesn't become overly long.

A "switch" is more readable and less error-prone when we need to exhaustively check all possible values of an enum.

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -105,6 +105,17 @@ function displayReaderList(readers) {
...
+      toolTipText = 'Initializing card...';
+    } else if (reader['status'] == GSC.PcscLiteServer.ReaderStatus.SUCCESS) {
+      toolTipText = 'Successful card insertion!';
+    } else if (reader['status'] == GSC.PcscLiteServer.ReaderStatus.FAILURE) {
+      toolTipText = 'Failed to connect card!';
+    }
+
+    indicatorContainer.setAttribute('title', toolTipText);
```
emaxx-google, line 117: Could you please implement it in a localizable way, using the "i18n" functionality that we use for other UI elements?

smart_card_connector_app/src/window.css
```diff
@@ -174,4 +174,4 @@ body {
   position: absolute;
   top: -2px;
   width: 20px;
-}
```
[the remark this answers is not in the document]
ayaNader: yeah, my bad

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -105,6 +105,17 @@ function displayReaderList(readers) {
     var indicatorContainer = goog.dom.createDom(
         'span', 'reader-indicator-container', indicator);
 
+    var toolTipText = '';
+    if (reader['status'] == GSC.PcscLiteServer.ReaderStatus.INIT) {
+      toolTipText = 'Initializing card...';
+    } else if (reader['status'] == GSC.PcscLiteServer.ReaderStatus.SUCCESS) {
+      toolTipText = 'Successful card insertion!';
```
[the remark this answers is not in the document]
ayaNader: hmmm, when i connect my reader without a card it always gives a failure status, so isn't success referring to a successful card connection?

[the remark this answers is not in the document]
emaxx-google: Shouldn't be the case. Double-check that the permissions for the reader are granted as per https://github.com/GoogleChromeLabs/chromeos_smart_card_connector#troubleshooting-apps-under-desktop-oses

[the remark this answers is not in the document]
ayaNader: ok it works now, thanks

smart_card_connector_app/src/window-devices-displaying.js
ayaNader: will do

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -126,6 +129,31 @@ function makeReaderNameForDisplaying(readerName) {
...
 
+/**
+ * @param {!GSC.PcscLiteServer.ReaderInfo} reader 
+ * @return {string}
+ */
+function getTooltipText(reader) {
+  let toolTipText;
+  switch (reader['status']) {
+    case GSC.PcscLiteServer.ReaderStatus.INIT:
+      toolTipText = 'initReaderTooltip';
```
emaxx-google, line 140: nit: We can just do `return 'initReaderTooltip'` here and in other cases - that's shorter and avoids the need in "toolTipText" and "break".

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -126,6 +129,31 @@ function makeReaderNameForDisplaying(readerName) {
...
+      toolTipText = 'successReaderTooltip';
+      break;
+    case GSC.PcscLiteServer.ReaderStatus.FAILURE:
+      toolTipText = 'failureReaderTooltip';
+      break;
+    default:
+      toolTipText = '';
+  }
+  if (reader['isCardPresent']) {
+    toolTipText = 'successCardTooltip';
```
emaxx-google, line 152: nit: This can be moved into the "SUCCESS" case above, making it easier to follow the logic.

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -126,6 +129,31 @@ function makeReaderNameForDisplaying(readerName) {
...
+      toolTipText = 'initReaderTooltip';
+      break;
+    case GSC.PcscLiteServer.ReaderStatus.SUCCESS:
+      toolTipText = 'successReaderTooltip';
+      break;
+    case GSC.PcscLiteServer.ReaderStatus.FAILURE:
+      toolTipText = 'failureReaderTooltip';
+      break;
+    default:
+      toolTipText = '';
```
emaxx-google, line 149: nit: Let's do an assertion here, since this branch should never be reached normally. E.g.,

`GSC.Logging.failWithLogger(logger);`
`return '';`

common/js/src/i18n.js
```diff
@@ -41,6 +41,9 @@ var I18N_DATA_ATTRIBUTE = 'data-i18n';
 /** @const */
 var I18N_DATA_ARIA_LABEL_ATTRIBUTE = 'data-i18n-aria-label';
 
+/** @const */
+var I18N_TITLE_ATTRIBUTE = 'title';
```
emaxx-google, line 45: I think we want this to be "data-title", so that there's no confusion between translated and untranslated text.

If we use the same attribute "title" as it is now, we'll have a problem that calling adjustElementTranslation() twice would call something like `chrome.i18n.getMessage('Initializing...')` and fail.

smart_card_connector_app/src/window-devices-displaying.js
ayaNader: ah that's true

common/js/src/i18n.js
```diff
@@ -94,22 +97,38 @@ function applyAriaLabelTranslation(element, translatedText) {
...
-        element, I18N_DATA_ATTRIBUTE, applyTranslation);
+        element, I18N_DATA_ATTRIBUTE, applyContentTranslation);
   }
 
   if (element.hasAttribute(I18N_DATA_ARIA_LABEL_ATTRIBUTE)) {
     transformElement(
         element, I18N_DATA_ARIA_LABEL_ATTRIBUTE, applyAriaLabelTranslation);
   }
+
+  if (element.hasAttribute(I18N_TITLE_ATTRIBUTE)) {
```
emaxx-google, line 128: Could you please add this new attribute into the test? (would be great to do this in the same CL)

smart_card_connector_app/src/_locales/en/messages.json
```diff
@@ -31,6 +31,22 @@
...
+  "initReaderTooltip": {
+    "message": "Initializing...",
+    "description": "Tooltip indicating that the reader is being initialized"
+  },
+  "successReaderTooltip": {
+    "message": "The reader is initialized.",
+    "description": "Tooltip indicating that the reader is initialized successfully"
+  },
+  "failureReaderTooltip": {
+    "message": "Failed to connect reader.",
```
emaxx-google, line 43: nit: "the reader"

ayaNader: It's grammatically correct and it's shorter so I thought it would be better.

emaxx-google: Is it correct?
I'm not a native speaker, but I would think that as we're showing this text for a *specific* reader, then we have to use "the" with it. Also I checked a couple of webpages regarding English Grammar and couldn't find why this sentence would be a valid case when no article can be used:
* https://7esl.com/use-no-article-rules/
* https://www.englishgrammar.org/omission-articles/

ayaNader: On grammarly.com when writing "Failed to connect device" it gives a grammatical error, but when writing "Error: Failed to connect device" it doesn't, so I'm assuming there are exceptions, but other than that it comes from intuition. Nonetheless, I'll change it.

PR #143, opened 2020-07-28
Add a script to create a list of supported readers

### LINKED ISSUE #141: Document the list of supported readers
Currently, we don't display the list of supported readers in the UI. This means almost zero troubleshooting possibility for the user (or an admin, a customer engineer, etc.) in case they experience some troubles. Also it means that it's hard to know whether a reader would work before buying it.

In fact, the CCID Free Software Driver that we use has a web page with a table of all supported readers: https://ccid.apdu.fr/ccid/section.html . However, the problem is that (1) this page is hard to discover, and (2) it corresponds to the latest version of the driver, meanwhile our Smart Card Connector app frequently has a few months old version of the driver (as we don't have a goal to update it immediately).

The request is to display this information in some form in the Smart Card Connector app's own UI, so that it's easily discoverable and free from versioning issues.

ayaNader:
Add a script that creates a human-readable file with the list of supported readers.
Closes #141


smart_card_connector_app/build/human_readable_supported_readers.txt
```diff
@@ -0,0 +1,845 @@
...
+
+Broadcom Corp 5880
+
+Broadcom Corp 5880
+
+Broadcom Corp 5880
+
+Broadcom Corp 5880
+
+Broadcom Corp 5880
```
Fabian-Sommer, line 171: I think we should filter any duplicates from this list.

smart_card_connector_app/src/create_human_readable_supported_readers.py
```diff
@@ -0,0 +1,132 @@
...
+      ignored_usb_device_count += 1
+
+  if not usb_devices:
+    raise RuntimeError('No supported USB devices were extracted from the CCID '
+                       'supported readers config')
+  print >>sys.stderr, ('Extracted {0} supported USB devices from the CCID '
+                       'supported readers config, and ignored {1} '
+                       'items.'.format(
+                           len(usb_devices), ignored_usb_device_count))
+
```
Fabian-Sommer, line 101: nit: No empty line necessary here.

smart_card_connector_app/src/create_human_readable_supported_readers.py
```diff
@@ -0,0 +1,132 @@
...
+  return usb_devices
+
+def create_readers_list(ccid_supported_readers_file):
+  usb_devices = load_usb_devices(ccid_supported_readers_file)
+  formatted_usb_devices = '\n'.join(sorted(usb_devices))
+
+  return formatted_usb_devices
```
Fabian-Sommer, line 108: nit: This can just return '\n'.join(sorted(usb_devices))

smart_card_connector_app/src/create_human_readable_supported_readers.py
```diff
@@ -0,0 +1,132 @@
...
+      continue
+    if line.startswith(COMMENT_PREFIX):
+      # Ignore comment line
+      continue
+    # Parse reader description line
+    parts = line.split(':', 2)
+    if len(parts) != 3:
+      raise RuntimeError('Failed to parse the reader description from the CCID '
+                         'supported readers config: "{0}"'.format(line))
+    name = parts[2]
```
Fabian-Sommer, line 85: nit: I think we don't need an empty line between each two entries in the resulting file. So I think here name should not end with a \n (probably best to just strip it from the line before splitting the line).

smart_card_connector_app/src/create_human_readable_supported_readers.py
ayaNader: it just makes the code cleaner IMO

smart_card_connector_app/src/create_human_readable_supported_readers.py
```diff
@@ -0,0 +1,133 @@
...
+                         'supported readers config before any section title')
+    if current_section in CCID_SUPPORTED_READERS_CONFIG_SECTIONS_TO_PICK:
+      usb_devices.append(name)
+    else:
+      ignored_usb_devices.append(name)
+
+  if not usb_devices:
+    raise RuntimeError('No supported USB devices were extracted from the CCID '
+                       'supported readers config')
+  print >>sys.stderr, ('Extracted {0} supported USB devices from the CCID '
```
emaxx-google, line 100: Given that this is a new file, let's probably be compatible with the Python 3 syntax (while keeping the Python 2 support as well). Specifically here, we have a choice between `print("foo", file=sys.stderr)` or `sys.stderr.write("foo\n")`. To make the former work on Python 2, you'd need to add a special import: `from __future__ import print_function` (it should be the first import).

You can test this from the terminal, by explicitly running the script with the Python: `python3 ../src/create_human_readable_supported_readers.py ...` and `python2 ../src/create_human_readable_supported_readers.py ...`.

smart_card_connector_app/build/human_readable_supported_readers.txt
```diff
@@ -0,0 +1,393 @@
+ACS ACR 38U-CCID
```
emaxx-google, line 1: Let's also change the Python script to print a clarification like:

`# NOTE: Auto-generated by create_human_readable_supported_readers.py.`
`# DO NOT MODIFY MANUALLY!`

smart_card_connector_app/src/create_human_readable_supported_readers.py
```diff
@@ -0,0 +1,133 @@
...
+  return usb_devices
+
+def create_readers_list(ccid_supported_readers_file):
+  usb_devices = load_usb_devices(ccid_supported_readers_file)
+  return '\n'.join(sorted(usb_devices))
+
+def main():
+  args_parser = argparse.ArgumentParser(
+      description="Creates a human readable file with the list of readers "
+      "supported by the Connector App.")
```
emaxx-google, line 114: nit: This should be indented with the opening `"` character in the previous line.

smart_card_connector_app/src/create_human_readable_supported_readers.py
```diff
@@ -97,21 +97,21 @@ def load_usb_devices(ccid_supported_readers_file):
...
+                   'supported readers config, and ignored {1} '
+                   'items.\n'.format(
+                       len(usb_devices), len(ignored_usb_devices)))
 
   return usb_devices
 
 def create_readers_list(ccid_supported_readers_file):
   usb_devices = load_usb_devices(ccid_supported_readers_file)
-  return '\n'.join(sorted(usb_devices))
+  return '\n'.join(sorted(usb_devices, key = str.lower))
```
emaxx-google, line 109: nit: No spaces around "=".

See https://google.github.io/styleguide/pyguide.html#36-whitespace

smart_card_connector_app/src/create_human_readable_supported_readers.py
```diff
@@ -127,6 +127,9 @@ def main():
   args = args_parser.parse_args()
 
   readers_list = create_readers_list(args.ccid_supported_readers_file)
+  args.target_file.write("# NOTE: Auto-generated by "
+  "create_human_readable_supported_readers.py.\n# DO NOT MODIFY MANUALLY!\n\n")
```
emaxx-google, line 131: nit: Add spaces to align the opening `"` with the `"` in the previous line.

Or indent it like this:

```python
args.target_file.write(
    "..."
    "...")

PR #144, opened 2020-07-29
Add a doc on updating the CCID driver

emaxx-google:


docs/updating-ccid.md
```diff
@@ -0,0 +1,77 @@
+# Updating the CCID free software driver
+
+Note: This page is mainly intended for the Chrome OS Smart Card Connector
+project maintainers.
+
+## Background
+
+The Chrome OS Smart Card Connector app is shipped with a specific version of the
+CCID free software driver. This is a driver for smart card reader (not to be
```
Fabian-Sommer, line 9: nit: readers

docs/updating-ccid.md
```diff
@@ -0,0 +1,77 @@
...
+confused with drivers for smart cards, also sometimes called middleware).
+
+Reasons for pinning to a specific version of the driver:
```
Fabian-Sommer, line 12: nit: Here you imply that we are using a pinned version of CCID, but it may be better to state it outright in this line. (I think when you state above that SCC is shipped with a specific version, it is not yet clear that the version in the repo is also pinned)

docs/updating-ccid.md
```diff
@@ -0,0 +1,77 @@
...
+
+* potentially, adjustments in our code or in the building scripts might be
```
Fabian-Sommer, line 14: nit: Potentially (since this is the beginning of a full sentence)

docs/updating-ccid.md
```diff
@@ -0,0 +1,77 @@
...
+  required after updating to new versions (note that we don't run the driver
+  verbatim, but instead port it under the Native Client technology for running
```
Fabian-Sommer, line 16: nit: 'to run' instead of 'for running'

docs/updating-ccid.md
```diff
@@ -0,0 +1,77 @@
...
+  inside the JavaScript-based App);
+* we can make sure that the driver that we roll out as part of the App underwent
```
Fabian-Sommer, line 18: nit: We (ditto)

Fabian-Sommer: nit: Also, I think this sentence is more complicated than it needs to be. How about 'We can test the driver version (that we roll out) and assure its quality'? (The part in () is optional, I would not literally include ( or ) )

docs/updating-ccid.md
```diff
@@ -0,0 +1,77 @@
...
+  some QA testing.
+
+However, this approach also implies some amount of maintenance work for updating
```
Fabian-Sommer, line 21: nit: 'to update' instead of 'work for updating'

docs/updating-ccid.md
```diff
@@ -0,0 +1,77 @@
...
+the driver (say, every few months). It also means that the users typically get
+a somewhat outdated version of the driver.
+
+## Update procedure
+
+0. The currently used version of the driver is located under the
+   `//third_party/ccid/` directory. Specifically, the
+   `//third_party/ccid/README.google` file should contain the version number,
```
Fabian-Sommer, line 29: nit: 'contains' instead of 'should contain'. Or, in the line below, replace 'contains' with 'should contain' (I would favor 'contains' in both places, but more important is consistency).

docs/updating-ccid.md
```diff
@@ -0,0 +1,77 @@
...
+   project page: [https://ccid.apdu.fr/](https://ccid.apdu.fr/).
+
+2. Delete all files of the previous version from `//third_party/ccid/src/`, and
+   unpack the new files from the downloaded archive into this directory. (Hint:
+   Pay attention to the directory structure; e.g., you should have the `AUTHORS`
+   file located at `//third_party/ccid/src/AUTHORS` in the end.)
+
+3. Apply the patches from the `//third_party/ccid/patches/` directory to the
+   files unpacked at the previous step. (Hint: We only have a single tiny patch
+   currently, so probably it's the simplest to just manually make the
```
Fabian-Sommer, line 43: nit: 'apply' instead of 'make'

docs/updating-ccid.md
```diff
@@ -0,0 +1,77 @@
...
+   modification shown in that patch file.)
+
+4. Edit the URL and the version in the `//third_party/ccid/README.google` file.
+
+5. Edit the version in the `//third_party/ccid/naclport/include.mk` file.
+
+6. Bump the file timestamps for the newly created files, in order to enforce the
+   `make` to recompile the CCID (otherwise, `make` will skip the files that have
+   old timestamps taken from the downloaded archive). To do this, for example,
```
Fabian-Sommer, line 52: nit: delete 'for example'

docs/updating-ccid.md
```diff
@@ -0,0 +1,77 @@
...
+8. In case of any errors discovered at the previous step, investigate and fix
+   them. (Hint: In case a new file was created in the driver, you have to add it
+   into `//third_party/ccid/naclport/build/Makefile` into one of the sections.)
+
+9. Start the just-built Smart Card Connector app (e.g., by running `make run` in
+   the `//smart_card_connector/build/` directory) and perform some manual QA
+   testing using physical reader(s).
+
+10. Check the logs from the App and double-check that they mention the correct
+    version of the driver (it'll be a string like
```
Fabian-Sommer, line 67: nit: 'it should' instead of 'it'll'

docs/updating-ccid.md
emaxx-google: Done. (I hope we won't forget to update the readme file :) )

PR #145, opened 2020-07-29
Split the README into smaller docs per-audience

emaxx-google:
The current single README.rst file became too big, and it was also
confusingly mixing information that is only interesting for specific
audience (e.g., only for developers of the Smart Card Connector app but
not for developers of consumer extensions/apps).


docs/building.md
```diff
@@ -0,0 +1,158 @@
+# Building the code
+
+This document describes steps needed for compiling the code of the Smart Card
+Connector App and the examples of programmatic communication with this App.
```
Fabian-Sommer, line 4: nit: I think 'programmatic communication' usually means something different than 'how two programs communicate with each other'. I would replace 'and the examples ...' with 'and examples of how programs can communicate with the app'.

docs/building.md
```diff
@@ -0,0 +1,158 @@
...
+
+* **Xvfb**.
+
+* (for 64-bit Linux) **32-bit version of libstdc++**.
+
+  For example, on Ubuntu it's provided by the libstdc++6:i386 package.
+
+* **Java Runtime Environment 7**.
+
+In order to **run** the built Apps, you will need *either* of these:
```
Fabian-Sommer, line 33: nit: App should generally not be capitalized according to https://english.stackexchange.com/questions/153696/app-or-app-when-i-refer-to-it-in-a-formal-paper#:~:text=Lowercase%2C%20except%20at%20the%20beginning,should%20consider%20writing%20out%20application. (here and in other places)

docs/connector-app-api.md
```diff
@@ -0,0 +1,165 @@
...
+   Note: this system has nothing to do with the Chrome App permissions model
+   (see
+   [https://developer.chrome.com/apps/declare_permissions](https://developer.chrome.com/apps/declare_permissions)),
+   as Chrome allows to use the messaging API without any additional permissions.
+   This intent is to close the hole left open by cross-app messaging between
+   Apps with different permissions.
+
+2. The Connector App is **bundled with a whitelist of known Extension/App
+   identifiers** (and a mapping to their display names). For all Extensions/Apps
+   not from this list, the user prompt will contain a big scary warning message.
```
Fabian-Sommer, line 32: Question: Is the big scary warning message the same thing as the user prompt dialog mentioned above? If so, points 1 and 2 can probably be merged.

docs/connector-app-api.md
```diff
@@ -0,0 +1,165 @@
...
+    "request_id": <requestId>,
+    "payload": {
+      "function_name": <functionName>,
+      "arguments": <functionArguments>
+    }
+  }
+}
+```
+
+where `<requestId>` should be a number (unique in the whole session of the
```
Fabian-Sommer, line 84: nit: integer instead of number (or are they usually called numbers in JS contexts?)

docs/building.md
emaxx-google: Done. (I was following the capitalization used throughout Chrome docs, like https://developer.chrome.com/apps/about_apps , but maybe it's more appropriate there than here.)

Issue #146, opened 2020-08-04
Reduce log clutter


emaxx-google:
Currently, around a half of a typical log line (from both the Smart Card Connector app and a typical smart card middleware app) is effectively non-informative clutter.

For example, from this log message:

```
[200804 10:20:00.00] [  1.234s] [<background page>.GoogleSmartCard.NaclModule<"nacl_module.nmf">.NaCl] [INFO] [syslog] 00000000 ../../src/src/ifdhandler.c:1961:init_driver() Driver version: 1.4.31
```

perhaps only the following is useful:

```
[200804 10:20:00.00] [  1.234s] [NaCl] [INFO] ifdhandler.c:1961:init_driver() Driver version: 1.4.31
```

The log clutter complicates troubleshooting and increases RAM usage (since some portion of logs is always aggregated).

PR #147, opened 2020-08-04
Omit "<background page>." from logs

emaxx-google:
Don't mention the background page in the JavaScript/NaCl log messages -
this information can be inferred from the absence of the HTML page URL.
Given that the majority of log messages come from the background page,
it makes sense to reduce the log clutter (and the RAM usage) by omitting
this part.

### DIFF
--- a/common/js/src/logging/log-formatting.js
+++ b/common/js/src/logging/log-formatting.js
@@ -43,16 +43,16 @@ textFormatter.showSeverityLevel = true;
  */
 function getFormattedDocumentLocation(documentLocation, isCompact) {
   if (!isCompact)
-    return documentLocation;
+    return `<${documentLocation}>`;
   let uri;
   try {
     uri = goog.Uri.parse(documentLocation);
   } catch (URIError) {
-    return documentLocation;
+    return `<${documentLocation}>`;
   }
   if (uri.getPath() == '/_generated_background_page.html')
-    return 'background page';
-  return uri.getPath();
+    return '';
+  return `<${uri.getPath()}>`;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


common/js/src/logging/log-formatting.js
```diff
@@ -62,10 +62,11 @@ function getFormattedDocumentLocation(documentLocation, isCompact) {
  * @return {!goog.log.LogRecord}
  */
 function prefixLogRecord(documentLocation, logRecord, isCompact) {
+  var loggerNameParts = [];
```
Fabian-Sommer, line 65: nit: let

common/js/src/logging/log-formatting.js
```diff
@@ -95,7 +96,7 @@ GSC.LogFormatting.formatLogRecord = function(documentLocation, logRecord) {
  * given document.
  *
  * The compact representation omits detailed context information, e.g. the
- * extension ID.
+ * extension ID or the background page label.
```
Fabian-Sommer, line 99: optional nit: and instead of or (I think both are correct, but and is the stronger statement)

common/js/src/logging/log-formatting.js
```diff
@@ -62,10 +62,11 @@ function getFormattedDocumentLocation(documentLocation, isCompact) {
...
+  var loggerNameParts = [];
   const formattedDocumentLocation = getFormattedDocumentLocation(
       documentLocation, isCompact);
-  var loggerNameParts = [];
-  loggerNameParts.push('<' + formattedDocumentLocation + '>');
+  if (formattedDocumentLocation)
+    loggerNameParts.push(formattedDocumentLocation);
   if (logRecord.getLoggerName())
     loggerNameParts.push(logRecord.getLoggerName());
   var prefixedLoggerName = loggerNameParts.join('.');
```
Fabian-Sommer, line 72: optional nit: let

PR #152, opened 2020-08-10
Omit "GoogleSmartCard." from logs

emaxx-google:
Stop using the "GoogleSmartCard." prefix in all logger names by default,
since it causes log clutter. The problem that this prefix was solving -
avoiding conflicts with any third-party code that creates its own
loggers - seems to be pretty theoretical, especially now with the
knowledge that the existing developers of middleware extensions don't
really use the Closure Library a lot.

So this commit introduces a parameter that allows to disable this
prefixing, and default-initializes this parameter to "disabled".
Instruction in the example app's makefile describes how to enable the
prefixing back.

The only place where this prefixing is kept enabled is the
"example_js_standalone_smart_card_client_library", which is used across
a bigger variety of apps/extensions that we don't have detailed
knowledge about, so that it's better to be conservative there.


common/js/src/logging/logging.js
```diff
@@ -49,6 +49,26 @@ goog.require('goog.log.Logger');
 goog.scope(function() {
 
 /** @const */
+var GSC = GoogleSmartCard;
```
Fabian-Sommer, line 52: nit: use const directly (here and in other places).
Maybe it's worth a separate CL to change all instances in this file?

emaxx-google: Done - created #155 for this.

PR #155, opened 2020-08-12
Modernize JS in common/js/src/logging/logging.js

emaxx-google:
Use "const"/"let" instead of "@const"/"var".

### DIFF
--- a/common/js/src/logging/logging.js
+++ b/common/js/src/logging/logging.js
@@ -48,17 +48,14 @@ goog.require('goog.log.Logger');
 
 goog.scope(function() {
 
-/** @const */
-var LOGGER_SCOPE = 'GoogleSmartCard';
+const LOGGER_SCOPE = 'GoogleSmartCard';
 
 /**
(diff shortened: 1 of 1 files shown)


common/js/src/logging/logging.js
```diff
@@ -48,46 +48,39 @@ goog.require('goog.log.Logger');
...
 
-/** @const */
-var LOGGER_SCOPE = 'GoogleSmartCard';
+const LOGGER_SCOPE = 'GoogleSmartCard';
 
 /**
  * The logging level that will be applied to the root logger (and therefore
  * would be effective for all loggers unless the ones that have an explicitly
  * set level).
- * @type {!goog.log.Level}
```
Fabian-Sommer, line 58: What is the reason to remove the type here?

emaxx-google: I assume that the compiler will be able to deduce the type automatically, based on the fact that it's a constant and the expression isn't too complex (e.g., without calls to functions that can return values of different types).

Issue #156, opened 2020-08-12
Detect and exit crash-and-restart loops


emaxx-google:
Currently, the Smart Card Connector app reloads itself in case a fatal error (e.g., inability to load the NaCl module or the absence of expected Extension APIs) occurs. This is done in the hope that the app will "self-repair" itself and continue serving user purposes - for instance, in case the crash was caused by some unexpected combination of internal operations, new operations after the restart will likely run in a different order and combination.

The problem with this approach is that sometimes there are unrepairable errors (for example, due to a bug somewhere or a misconfigured device), and in this case the app will end up reloading itself again and again, until Chrome disables the app with the "This extension reloaded itself too frequently" error. The worst thing about this scenario is that there's no way to obtain the app's logs, meaning that the troubleshooting and investigation of bug reports is extremely difficult.

The proposal is to suppress the self-reload functionality in case the crash-and-restart loop is detected (most likely, heuristically).

PR #157, opened 2020-08-13
Drop var_args from logging.js methods

emaxx-google:
These optional formatting arguments are not used anywhere, and will
complicate the upcoming implementation of crash loop detection (tracked
by #156).

Therefore this CL deletes this unused functionality.

### DIFF
--- a/common/js/src/logging/logging.js
+++ b/common/js/src/logging/logging.js
@@ -202,7 +202,7 @@ GSC.Logging.checkWithLogger = function(
 
 /**
  * Fails with the specified message.
-  *
+ *
  * In the debug mode, this function effectively throws an instance of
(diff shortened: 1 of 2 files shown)


common/js/src/logging/logging.js
```diff
@@ -194,31 +193,28 @@ GSC.Logging.check = function(condition, opt_message, var_args) {
...
+    logger, condition, opt_message) {
   if (!condition)
-    GSC.Logging.failWithLogger(logger, opt_message, var_args);
+    GSC.Logging.failWithLogger(logger, opt_message);
 };
 
 /**
  * Fails with the specified message.
- *
+  *
```
Fabian-Sommer, line 205: nit: unintentional whitespace?

PR #158, opened 2020-08-13
Detect crash loops

### LINKED ISSUE #156: Detect and exit crash-and-restart loops

emaxx-google:
Store the recent crash timestamps of the Smart Card Connector app in the
chrome.storage.local data, and skip reloading itself on fatal error in
case this data shows that the reloads have been happening frequently.

This fixes #156.


emaxx-google:
Still working on unit tests for this.

common/js/src/logging/crash-loop-detection.js
```diff
@@ -0,0 +1,132 @@
...
+        -CRASH_LOOP_THRESHOLD_COUNT + 1);
+  }
+  newCrashTimestamps.push(currentTimestamp);
+
+  if (!chrome || !chrome.storage || !chrome.storage.local) {
+    // Cannot store any data without access to the chrome.storage API.
+    return Promise.resolve(newCrashTimestamps);
+  }
+  return new Promise(resolve => {
+    chrome.storage.local.set({[STORAGE_KEY]: newCrashTimestamps}, function() {
```
Fabian-Sommer, line 116: Question: Why is STORAGE_KEY inside [] here?

common/js/src/logging/logging.js
```diff
@@ -201,23 +202,18 @@ GSC.Logging.checkWithLogger = function(
...
- * Fails with the specified message.
-  *
- * In the debug mode, this function effectively throws an instance of
- * goog.asserts.AssertionError.
+ * Throws an exception and emits severe log with the specified message.
  *
- * In the release mode, this function emits severe log message and initiates the
- * App reload.
+ * In the release mode, this additionally asynchronously initiates the App
+ * reload, unless the crash-and-reload loop is detected.
```
Fabian-Sommer, line 208: nit: replace 'the' with 'a'

common/js/src/logging/crash-loop-detection.js
```diff
@@ -0,0 +1,132 @@
...
+const STORAGE_KEY = 'crash_timestamps';
+
+// The time window, during which the crashes are counted and compared against
+// |CRASH_LOOP_THRESHOLD_COUNT|.
+const CRASH_LOOP_WINDOW_MILLISECONDS = 60 * 1000;
+// If the number of crashes reaches this constant within the
+// |CRASH_LOOP_WINDOW_MILLISECONDS| time window, then it's considered to be a
+// crash loop.
+//
+// Note that these two constants have to be chosen to be stricter than the
```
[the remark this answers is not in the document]
emaxx-google: Maybe I'm confusing something, but I think that a longer time window is stricter than a shorter one? I.e., the sequence of crashes "t=0, t=12, t=24, t=36, t=48" is a violation in the "window=60" configuration, but not in the "window=10" one.

P.S. Replaced "two constants" with their names, to avoid confusion.

common/js/src/logging/crash-loop-detection.js
emaxx-google: It's the JS syntax (introduced in ECMAScript2015) that means "read the STORAGE_KEY's value", rather than "literally use the `"STORAGE_KEY"` string".

But I've replaced it with a separate object construction, so that there's no confusion.

emaxx-google:
PTAL - unit tests added.

emaxx-google:
> LGTM % nit
> re strict vs relaxed condition - I suppose it depends on the point of view - this implementation detects a superset of possible runs to be crash loops compared to Chrome. This should make the condition for being in a crash loop more relaxed (which is what I was thinking of), but the condition for not being in a crash loop more strict (I guess this is what you were referring to). Ultimately, it doesn't matter a lot.

I see your point, got it. I've reworded the comment a bit to avoid this ambiguity.

PR #162, opened 2020-08-14
Add flag for disabling the reload-on-failure

emaxx-google:
Add a build-time parameter that allows to enable/disable the
self-reloading of the extension/app in case a fatal error occurs. This
allows to easily disable this functionality in client applications that
don't want this behavior.

Note that this parameter also controls whether the "crash loop
detection" introduced in #156 is enabled.

Also leave this parameter in the "disabled" state for the
"example_js_standalone_smart_card_client_library", since this library is
used in a wide variety of different extensions/apps, and, most likely,
they don't want to be self-reloaded in case the smart card code fails.
It was an oversight that this functionality was enabled there, and now
with addition in #156 of chrome.storage manipulations it became more
apparent that we want to disable it there by default.

### DIFF
--- a/smart_card_connector_app/build/Makefile
+++ b/smart_card_connector_app/build/Makefile
@@ -58,6 +58,10 @@ include $(THIRD_PARTY_DIR_PATH)/pcsc-lite/naclport/server_clients_management/inc
 SOURCES_PATH := ../src
 
 
+JS_COMPILER_FLAGS := \
+	--define='GoogleSmartCard.Logging.SELF_RELOAD_ON_FATAL_ERROR=true' \
+
+
 #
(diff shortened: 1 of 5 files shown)


smart_card_connector_app/build/Makefile
```diff
@@ -58,6 +58,10 @@ include $(THIRD_PARTY_DIR_PATH)/pcsc-lite/naclport/server_clients_management/inc
 SOURCES_PATH := ../src
 
 
+JS_COMPILER_FLAGS := \
```
Fabian-Sommer, line 61: nit: I think one empty line between SOURCES_PATH and JS_COMPILER_FLAGS is sufficient

PR #165, opened 2020-08-19
Delete Boost library from copyright notices page

emaxx-google:
Remove mentions of the Boost library and its license from the "About"
window of the Smart Card Connector app, since this library isn't used
in the project anymore (since #91).


Fabian-Sommer (approved):
LGTM

PR #168, opened 2020-09-24
Cherry-pick Sunrex HP USB Business Slim Smartcard CCID Keyboard suppo…

emaxx-google:
…rt into CCID

Apply a patch from the upstream Free CCID library that enables the
support of the Sunrex HP USB Business Slim Smartcard CCID Keyboard:

LudovicRousseau/CCID@48680b3

This local patch can be removed in the future once the upstream CCID
library releases a new version which includes this change.

### DIFF
--- a/third_party/ccid/patches/0003-Add-Sunrex-HP-USB-Business-Slim-Smartcard-CCID-Keybo.patch
+++ b/third_party/ccid/patches/0003-Add-Sunrex-HP-USB-Business-Slim-Smartcard-CCID-Keybo.patch
@@ -1,0 +1,100 @@
+From 48680b3af325d63f58516b5737e8df8985b53bc7 Mon Sep 17 00:00:00 2001
+From: Ludovic Rousseau <ludovic.rousseau@free.fr>
+Date: Sat, 29 Aug 2020 11:45:10 +0200
+Subject: [PATCH] Add Sunrex HP USB Business Slim Smartcard CCID Keyboard
+
+---
+ ..._Business_Slim_Smartcard_CCID_Keyboard.txt | 58 +++++++++++++++++++
+ readers/supported_readers.txt                 |  5 +-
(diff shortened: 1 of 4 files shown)


emaxx-google:
The support of this reader was requested by one of our customers.

third_party/ccid/patches/0003-Add-Sunrex-HP-USB-Business-Slim-Smartcard-CCID-Keybo.patch
```diff
@@ -0,0 +1,100 @@
...
++   PIN Modification supported
++  bMaxCCIDBusySlots: 1
+diff --git a/readers/supported_readers.txt b/readers/supported_readers.txt
+index 1fcb100..444ac14 100644
+--- a/readers/supported_readers.txt
++++ b/readers/supported_readers.txt
+@@ -1,6 +1,6 @@
+ #
+ # List of readers supported by the CCID driver
+-# Generated: 2020-08-26
```
Fabian-Sommer, line 83: nit: This is probably irrelevant, but this according to the github review interface this used to be 2020-06-25

Fabian-Sommer (approved):
LGTM % nit/question

third_party/ccid/patches/0003-Add-Sunrex-HP-USB-Business-Slim-Smartcard-CCID-Keybo.patch
emaxx-google: This is taken from the upstream patch LudovicRousseau/CCID@48680b3 as-is, so let's probably keep it. You're right that in our repo the timestamp is different, but that's practically unavoidable, as long as we start cherry-picking patches from the middle of history.

emaxx-google:
Thanks for the review.

PR #170, opened 2020-09-29
Cherry-pick ACS ACR1252IMP support into CCID

ankhler:
Apply a patch from the upstream Free CCID library that enables the support of the ACS ACR1252IMP smart card reader:

LudovicRousseau/CCID@183c09c

This local patch can be removed in the future once the upstream CCID library releases a new version which includes this change.


Fabian-Sommer (approved):
LGTM (but be aware that your private email is visible in the .patch file - not sure you want this)

emaxx-google (approved):

PR #171, opened 2020-09-29
Fix incorrect setCertificate API calls

Fabian-Sommer:
Add example usage of setCertificates to the cpp example app.
Fix incorrect API calls that would previously happen when trying to use
ApiBridge::SetCertificates() by transforming the used parameters on the
JS side directly before the API call.

### DIFF
--- a/example_cpp_smart_card_client_app/src/chrome_certificate_provider/bridge-backend.js
+++ b/example_cpp_smart_card_client_app/src/chrome_certificate_provider/bridge-backend.js
@@ -246,19 +246,11 @@ Backend.prototype.handleRequest_ = function(payload) {
   const apiFunction = chrome.certificateProvider[
       remoteCallMessage.functionName];
   if (apiFunction) {
-    if (remoteCallMessage.functionName == 'setCertificates') {
-      // The arguments need to be transformed in order to be recognized by the
-      // API.
-      const certificates = goog.array.map(
-          remoteCallMessage.functionArguments[0]["clientCertificates"],
-          createClientCertificateInfo);
-      remoteCallMessage.functionArguments[0]["clientCertificates"] =
-          certificates;
-    }
-
+    const transformedFunctionArguments = transformFunctionArguments(
+        remoteCallMessage.functionName, remoteCallMessage.functionArguments);
     /** @preserveTry */
     try {
-      remoteCallMessage.functionArguments.push(function() {
+      transformedFunctionArguments.push(function() {
         if (chrome.runtime.lastError) {
           promiseResolver.reject(new Error(goog.object.get(
               chrome.runtime.lastError, 'message', 'Unknown error')));
@@ -266,7 +258,7 @@ Backend.prototype.handleRequest_ = function(payload) {
         }
         promiseResolver.resolve(Array.prototype.slice.call(arguments));
       });
-      apiFunction.apply(null, remoteCallMessage.functionArguments);
+      apiFunction.apply(null, transformedFunctionArguments);
     } catch (exc) {
       promiseResolver.reject(exc);
     }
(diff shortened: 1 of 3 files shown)


example_cpp_smart_card_client_app/src/pp_module.cc
```diff
@@ -372,16 +404,19 @@ class PpInstance final : public pp::Instance {
...
   // This method is executed on a background thread after all of the
   // initialization steps finish.
-  static void Work() {
+  static void Work(
+      std::weak_ptr<ccp::ApiBridge> chrome_certificate_provider_api_bridge) {
     //
     // CHANGE HERE:
     // Place your custom initialization code here:
     //
+    ReportAvailableCertificates(chrome_certificate_provider_api_bridge);
```
emaxx-google, line 419: nit: Maybe let's add a hint that this call is still expected to be done in the real-world application. E.g.:

```
    //
    // CHANGE HERE:
    // Place your custom initialization code here:
    //

    // Report the currently available list of certificates immediately after the
    // initialization, according to requirements imposed by Chrome - see
    // <https://developer.chrome.com/extensions/certificateProvider#method-setCertificates>.
    ReportAvailableCertificates(chrome_certificate_provider_api_bridge);
```

example_cpp_smart_card_client_app/src/pp_module.cc
```diff
@@ -54,6 +54,56 @@ namespace smart_card_client {
...
+      locked_chrome_certificate_provider_api_bridge =
+          chrome_certificate_provider_api_bridge.lock();
+  if (!locked_chrome_certificate_provider_api_bridge) {
+    GOOGLE_SMART_CARD_LOG_INFO << "Cannot provide certificates: The " <<
+        "shutdown process has started";
+    return false;
+  }
+
+  std::vector<ccp::ClientCertificateInfo> certificates;
+  if (GetCertificates(&certificates)) {
```
emaxx-google, line 96: So when this function fails, we won't notify Chrome with an empty list? I guess this might be useful in case of some transient error. But that shouldn't be used in case the failure is simply "there's no smart card available", since instead of immediately notifying Chrome about the empty list we'll wait forever.

I'm concerned that the extension author might confuse these two situations and cause weird side effects. So I'd lead towards dropping the bool result, and only return the vector from the function. WDYT?

emaxx-google (changes requested):
Thanks. Let's also rename the pull request - I think the main point is that we're fixing the setCertificates bug.

example_cpp_smart_card_client_app/src/pp_module.cc
Fabian-Sommer: Done

example_cpp_smart_card_client_app/src/chrome_certificate_provider/bridge-backend.js
```diff
@@ -520,6 +512,24 @@ Backend.prototype.rejectedCertificatesCallback_ = function(
...
+ */
+function transformFunctionArguments(functionName, functionArguments) {
+  const transformedArguments = functionArguments.slice();
+  if (functionName === 'setCertificates') {
+    // The certificates need to be transformed in order to be recognized by the
+    // API.
+    const certificates = goog.array.map(
+        transformedArguments[0]["clientCertificates"],
+        createClientCertificateInfo);
+    transformedArguments[0]["clientCertificates"] = certificates;
```
emaxx-google, line 528: I think this modifies the key in the original object that was passed into |functionArguments|?

emaxx-google (changes requested):

example_cpp_smart_card_client_app/src/chrome_certificate_provider/bridge-backend.js
```diff
@@ -523,9 +523,9 @@ function transformFunctionArguments(functionName, functionArguments) {
     // The certificates need to be transformed in order to be recognized by the
     // API.
     const certificates = goog.array.map(
-        transformedArguments[0]["clientCertificates"],
+        functionArguments[0]["clientCertificates"],
         createClientCertificateInfo);
-    transformedArguments[0]["clientCertificates"] = certificates;
+    transformedArguments[0] = {"clientCertificates": certificates};
```
emaxx-google, line 528: There can also be optional keys `certificatesRequestId` and `error`.

emaxx-google (approved):
Thanks!

PR #172, opened 2020-09-30
Omit '<"nacl_module.nmf">.NaCl' from logs

emaxx-google:
Don't add the '<"nacl_module.nmf">.NaCl' suffix to the logger name
used for NaCl-related log messages. This information about the URL of
the NaCl module is redundant - we don't use multiple NaCl modules
currently.

Also drop the ".NaCl" suffix for messages forwarded from the NaCl
module, and instead add ".Wrapper" for all other messages logged
by the JS code of this class. This way, we move the log clutter from
typically long NaCl logs into usually less interesting JS logs.

Example before:
 [200930 04:21:34.51] [  7.244s] [NaclModule<"nacl_module.nmf">.NaCl] [syslog] 00000000 ../../src/src/ifdhandler.c:2008:init_driver() Driver version: 1.4.33 
Example after:
 [200930 04:21:34.51] [  7.244s] [NaclModule] [syslog] 00000000 ../../src/src/ifdhandler.c:2008:init_driver() Driver version: 1.4.33  

For the potential case when the detailed logging might be needed, a
boolean flag is kept in the code, allowing to revert to the old
behavior. Currently the flag is always set to false.

### DIFF
--- a/common/js/src/nacl-module/nacl-module-log-messages-receiver.js
+++ b/common/js/src/nacl-module/nacl-module-log-messages-receiver.js
@@ -39,9 +39,6 @@ var GSC = GoogleSmartCard;
 var SERVICE_NAME = 'log_message';
 
 /** @const */
-var LOGGER_TITLE = 'NaCl';
-
-/** @const */
 var LOG_LEVEL_MESSAGE_KEY = 'log_level';
 /** @const */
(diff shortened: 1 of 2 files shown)


Fabian-Sommer (approved):
LGTM

PR #173, opened 2020-10-06
Framework for JS-and-C++ integration tests

emaxx-google:
Implement a mini-framework that allows to implement hybrid tests that
exercise both JavaScript and C++ code. In such tests, the main test
script will be written in JavaScript, meanwhile the C++ counterpart is
structured as a set of "helpers" that can be called from the JS code.
The implemented framework abstracts away some boilerplate related to
JS-and-C++ communication and code compilation.

This commit paves the way for fixing the missing test coverage for the
parts that cross the language boundary. Up to this point, we only had
unit test coverage for pieces within single language (only within
JavaScript or only within C/C++), which has been insufficient.

### DIFF
--- a/common/integration_testing/build/.gitignore
+++ b/common/integration_testing/build/.gitignore
@@ -1,0 +1,2 @@
+/headers_installed.stamp
+/pnacl/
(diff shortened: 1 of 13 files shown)


emaxx-google:
PTAL. Sorry for the big CL - it's mainly because the problem of cross-language integration testing is hard, and so far we haven't built anything to simplify it.

A follow-up PR will add first tests using this new framework.

Fabian-Sommer (approved):
LGTM

PR #174, opened 2020-10-06
Smoke tests for setCertificates() C++ adapter

emaxx-google:
Add first integration tests that verify the triggered JavaScript
Extensions API calls after running the helper C++ function
chrome_certificate_provider::ApiBridge::SetCertificates().

These first tests cover only the simplest scenario: when the function is
called with an empty list of certificates. The tests, however, cover
both cases of the up-to-date and the legacy chrome.certificateProvider
APIs.

### DIFF
--- a/example_cpp_smart_card_client_app/src/chrome_certificate_provider/bridge-integrationtest.js
+++ b/example_cpp_smart_card_client_app/src/chrome_certificate_provider/bridge-integrationtest.js
@@ -40,6 +40,7 @@ let setCertificatesApiExpectation;
  * Sets up stub implementation of the chrome.certificateProvider API.
  */
 function setUpApiStubs() {
+  setCertificatesApiExpectation = goog.Promise.withResolver();
   testController.propertyReplacer.set(chrome, 'certificateProvider', {
     'setCertificates': stubSetCertificatesApi,
   });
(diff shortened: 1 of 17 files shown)


example_cpp_smart_card_client_app/src/chrome_certificate_provider/bridge-integrationtest.js
```diff
@@ -0,0 +1,104 @@
...
+function stubSetCertificatesApi(details, callback) {
+  setCertificatesApiExpectation.resolve(details);
+  if (callback)
+    callback();
+}
+
+goog.exportSymbol('testChromeCertificateProviderApiBridge', {
+  setUp: function() {
+    testController = new GSC.IntegrationTestController();
+    setCertificatesApiExpectation = goog.Promise.withResolver();
```
Fabian-Sommer, line 70: nit: I think this should be initialized either in setUpApiStubs or in testSetCertificates_empty

Fabian-Sommer (approved):
LGTM % nit
I guess unfortunately this would not catch the bug that we had recently, with the API call being mocked out, but actually doing so might be impossible short of bringing in a complete instance of Chrome.

emaxx-google:
@googlebot I consent

emaxx-google:
Thanks for the review.

> LGTM % nit
> I guess unfortunately this would not catch the bug that we had recently, with the API call being mocked out, but actually doing so might be impossible short of bringing in a complete instance of Chrome.

I'll add a more nontrivial test in a follow-up, and it'll verify that the call argument is equal to the expected object. As long as the test expectations are conforming to the Chrome API definition, this will give us the confidence in correct behavior.

PR #175, opened 2020-10-06
Fix JS Requester post-after-dispose bug

emaxx-google:
Fix a bug that could lead to JavaScript exception in a corner case when
the Requester's postRequest() method is called after the object is
already disposed of. The error was:

  TypeError: Cannot read property 'get' of null
  at GoogleSmartCard.Requester.popRequestPromiseResolver_

Also add a regression unit test against this bug.


Fabian-Sommer (approved):
LGTM

PR #176, opened 2020-10-06
Nonempty test case for setCertificates() C++ adapter

emaxx-google:
Add an integration test that verifies correct conversion of certificate
data from the C++ struct into the JavaScript object when the
chrome.certificateProvider.setCertificates() method is called.

This is a regression test for the bug fixed in #171.


emaxx-google:
Note: Depending on resolution of https://github.com/google/closure-library/issues/1090, I might need to add a custom assertion comparator into the test, or cherry-pick the Closure Library fix in case that one will be written.

Fabian-Sommer (approved):
LGTM

PR #178, opened 2020-10-12
Skip existing files&dirs in env/initialize.sh

emaxx-google:
Change the env/initialize.sh script to be idempotent by default, so that
already existing tools and libraries are left intact by it. Add the "-f"
option that allows to revert to the previous "destructive" behavior.

This is a preparation change for introducing the WebAssembly build tools
initialization in #177, since it'll make it possible for developers to
rerun "initialize.sh" in order to get WebAssembly tools without
destroying their existing Native Client related state.

### DIFF
--- a/env/initialize.sh
+++ b/env/initialize.sh
@@ -24,7 +24,7 @@
 
 set -eu
 
-is_forced=0
+force_reinitialization=0
 
 
 log_message() {
@@ -40,7 +40,7 @@ log_error_message() {
 }
 
 initialize_depot_tools() {
-  if [ -d ./depot_tools -a "${is_forced}" -eq "0" ]; then
+  if [ -d ./depot_tools -a "${force_reinitialization}" -eq "0" ]; then
     log_message "depot_tools already present, skipping."
     return
   fi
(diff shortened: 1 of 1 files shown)


env/initialize.sh
```diff
@@ -14,9 +14,18 @@
...
+# all necessary tools and libraries.
+#
+# By default, existing directories and files are left intact (except the
+# "./activate" file), so it's safe to run this script multiple times. Add the
+# "-f" flag in order to enforce overwriting the existing state.
+
 
 set -eu
 
+is_forced=0
```
Fabian-Sommer, line 27: nit: This name doesn't really convey what is enforced - maybe rename it to force_reinitialization or so

env/initialize.sh
```diff
@@ -31,6 +40,10 @@ log_error_message() {
 }
 
 initialize_depot_tools() {
+  if [ -d ./depot_tools -a "${is_forced}" -eq "0" ]; then
```
Fabian-Sommer, line 43: Not familiar with the syntax here - I assume this checks 1) existence of the folder `./depot_tools` and 2) is_forced being equal to 0 - right?

Fabian-Sommer (approved):
LGTM % nit

env/initialize.sh
emaxx-google: Done

env/initialize.sh
emaxx-google: Correct - see, e.g., https://tldp.org/LDP/abs/html/comparison-ops.html .

PR #180, opened 2020-10-12
Download Emscripten in env/initialize.sh

emaxx-google:
Add download and installation of Emscripten into the env/initialize.sh
script. This will be used in the future for building the WebAssembly
ports of the C/C++ code in the Smart Card Connector app and related
projects (instead of the currently used Native Client) - see #177.

The version of Emscripten is pinned, so that all developers get
consistent results and avoid any potential breakages related to uprevs.

### DIFF
--- a/env/constants.sh
+++ b/env/constants.sh
@@ -18,9 +18,8 @@
 DEPOT_TOOLS_REPOSITORY_URL="https://chromium.googlesource.com/chromium/tools/depot_tools.git"
 
 EMSCRIPTEN_SDK_REPOSITORY_URL="https://github.com/emscripten-core/emsdk.git"
+EMSCRIPTEN_VERSION="2.0.6"
 
 NACL_SDK_VERSION="47"
 
 WEBPORTS_TARGETS="glibc-compat openssl"
-
-EMSCRIPTEN_VERSION="2.0.6"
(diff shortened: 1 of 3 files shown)


env/constants.sh
```diff
@@ -17,6 +17,10 @@
 
 DEPOT_TOOLS_REPOSITORY_URL="https://chromium.googlesource.com/chromium/tools/depot_tools.git"
 
+EMSCRIPTEN_SDK_REPOSITORY_URL="https://github.com/emscripten-core/emsdk.git"
+
 NACL_SDK_VERSION="47"
 
 WEBPORTS_TARGETS="glibc-compat openssl"
+
+EMSCRIPTEN_VERSION="2.0.6"
```
MahmoudAGawad, line 26: nit: move all EMSCRIPTEN constants next to each others.

MahmoudAGawad (approved):

env/initialize.sh
```diff
@@ -114,6 +130,7 @@ while getopts ":f" opt; do
   esac
 done
 
+initialize_emscripten
```
Fabian-Sommer, line 133: nit: Unless there is some reason not to, we can maintain alphabetical order here

Fabian-Sommer (approved):
LGTM % nit

env/initialize.sh
emaxx-google: Done, and also added comments for items that actually depend on others' result. (Accidentally, the alphabetical order is correct for them too.)

emaxx-google:
Thanks!

PR #182, opened 2020-10-13
Introduce variant Value type

emaxx-google:
Initial implementation of the "Value" class that, approximately, can
hold any JSONifiable object.

This class will be used in follow-ups to replace the widespread usage of
the Native Client / Pepper specific "pp::Var" class, so that all
depending code can be reused in WebAssembly builds (that effort is
tracked by #185).

The class is additionally inspired by the Chromium's "base::Value", but
we're having a much simpler custom implementation here, because
Chromium's implementation brings a bunch of dependencies and is not
trivial to compile.

### DIFF
--- a/common/cpp/src/google_smart_card_common/value.h
+++ b/common/cpp/src/google_smart_card_common/value.h
@@ -25,7 +25,7 @@
 
 namespace google_smart_card {
 
-// A variadic data type that approximately corresponds to JSONifiable types.
+// A variant data type that approximately corresponds to JSONifiable types.
 //
 // Is intended to be used in generic interfaces related to message exchanging
 // with remote callers/receivers, for instance, for sending/receiving messages
@@ -82,9 +82,10 @@ class Value final {
   double GetFloat() const;
   const std::string& GetString() const;
   const BinaryStorage& GetBinary() const;
-  const DictionaryStorage& GetDictionaryItems() const;
-  const ArrayStorage& GetArrayItems() const;
+  const DictionaryStorage& GetDictionary() const;
+  const ArrayStorage& GetArray() const;
 
+  // Returns null when the key isn't present.
   const Value* GetDictionaryItem(const std::string& key) const;
 
   void SetDictionaryItem(std::string key, Value value);
--- a/common/cpp/src/google_smart_card_common/value.h
+++ b/common/cpp/src/google_smart_card_common/value.h
@@ -1,0 +1,137 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_SMART_CARD_COMMON_VALUE_H_
+#define GOOGLE_SMART_CARD_COMMON_VALUE_H_
+
+#include <stdint.h>
+
+#include <map>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace google_smart_card {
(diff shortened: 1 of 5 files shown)


common/cpp/src/google_smart_card_common/value.h
```diff
@@ -0,0 +1,137 @@
...
+  bool is_binary() const { return type_ == Type::kBinary; }
+  bool is_dictionary() const { return type_ == Type::kDictionary; }
+  bool is_array() const { return type_ == Type::kArray; }
+
+  bool GetBoolean() const;
+  int64_t GetInteger() const;
+  double GetFloat() const;
+  const std::string& GetString() const;
+  const BinaryStorage& GetBinary() const;
+  const DictionaryStorage& GetDictionaryItems() const;
```
Fabian-Sommer, line 85: nit: I think GetDictionary() would be a better name, since this actually returns the entire dictionary and not just the items contained in it

common/cpp/src/google_smart_card_common/value.h
```diff
@@ -0,0 +1,137 @@
...
+  const ArrayStorage& GetArrayItems() const;
```
Fabian-Sommer, line 86: Ditto

common/cpp/src/google_smart_card_common/value.h
```diff
@@ -0,0 +1,137 @@
...
+
+  const Value* GetDictionaryItem(const std::string& key) const;
```
Fabian-Sommer, line 88: nit: I think this could use a comment about it returning nullptr when the key is not present

common/cpp/src/google_smart_card_common/value_unittest.cc
```diff
@@ -0,0 +1,339 @@
...
+  EXPECT_TRUE(value.is_null());
+  EXPECT_FALSE(value.is_integer());
+  EXPECT_FALSE(value.is_float());
+  EXPECT_FALSE(value.is_string());
+  EXPECT_FALSE(value.is_binary());
+  EXPECT_FALSE(value.is_dictionary());
+  EXPECT_FALSE(value.is_array());
+}
+
+TEST(ValueTest, Integer) {
```
Fabian-Sommer, line 53: nit: Add another test that uses the int64_t ctor.

common/cpp/src/google_smart_card_common/value_unittest.cc
```diff
@@ -0,0 +1,339 @@
...
+  const Value value(bytes);
+  EXPECT_EQ(value.type(), Value::Type::kBinary);
+  EXPECT_FALSE(value.is_null());
+  EXPECT_FALSE(value.is_integer());
+  EXPECT_FALSE(value.is_float());
+  EXPECT_FALSE(value.is_string());
+  EXPECT_TRUE(value.is_binary());
+  EXPECT_FALSE(value.is_dictionary());
+  EXPECT_FALSE(value.is_array());
+  EXPECT_EQ(value.GetBinary(), bytes);
```
Fabian-Sommer, line 144: Does this work as expected (i.e., does it actually compare the contents of the arrays)?

Fabian-Sommer (approved):
LGTM % nits

common/cpp/src/google_smart_card_common/value.h
```diff
@@ -0,0 +1,137 @@
...
+
+#include <map>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace google_smart_card {
+
+// A variadic data type that approximately corresponds to JSONifiable types.
```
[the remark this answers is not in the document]
emaxx-google: Replaced with the "variant" word, as this is somewhat similar to C++17's std::variant. Let me know if you'd like a different wording.

common/cpp/src/google_smart_card_common/value_unittest.cc
emaxx-google: Yes. Verified this manually by comparing with wrong bytes - it failed with:
```
../../src/google_smart_card_common/value_unittest.cc:177: Value of: bytes2
  Actual: { '\x1' (1), '\x2' (2), '\x3' (3), '\x4' (4) }
Expected: value.GetBinary()
Which is: { '\x1' (1), '\x2' (2), '\x3' (3) }
```

MahmoudAGawad (approved):
LGTM.

emaxx-google:
Thanks for the reviews!

PR #183, opened 2020-10-13
Drop NACL_LIBRARY_BUILD_RULE Makefile rule

emaxx-google:
Use the LIB_RULE macro rule directly instead of the
NACL_LIBRARY_BUILD_RULE alias - the latter wasn't making a big
difference anyway, and also reducing the number of entities in our build
scripts will make WebAssembly migration (as tracked by #177) simpler.


Fabian-Sommer (approved):
LGTM

Issue #185, opened 2020-10-14
Add WebAssembly support into //common/cpp


emaxx-google:
The C++ common utilities library (//common/cpp) needs to be refactored and extended in order to support working under WebAssembly. Currently it only compiles under Native Client.

PR #186, opened 2020-10-14
Add conversion from NaCl's pp::Var into Value

emaxx-google:
Implement helper function that transforms a Native Client SDK's pp::Var
object into a Value object.

This helper will be used in follow-ups in order to make most of the
//common/cpp library be toolchain-independent and work under both Native
Client and WebAssembly (as tracker by #185), since the Value class is
toolchain-independent.

### DIFF
--- a/common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.cc
+++ b/common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.cc
@@ -1,0 +1,105 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <google_smart_card_common/value_nacl_pp_var_conversion.h>
+
+#include <stdint.h>
+
+#include <cstring>
+#include <limits>
+#include <string>
+#include <utility>
+
+#include <ppapi/cpp/var.h>
+#include <ppapi/cpp/var_array.h>
+#include <ppapi/cpp/var_array_buffer.h>
+#include <ppapi/cpp/var_dictionary.h>
+
+#include <google_smart_card_common/logging/logging.h>
+#include <google_smart_card_common/unique_ptr_utils.h>
+#include <google_smart_card_common/value.h>
+
+namespace google_smart_card {
+
+namespace {
+
+pp::Var CreateIntegerVar(int64_t integer_value) {
+  // `pp::Var` can only hold 32-bit integers. Larger numbers need to be
+  // represented via double.
+  if (integer_value >= std::numeric_limits<int32_t>::min() &&
+      integer_value <= std::numeric_limits<int32_t>::max())
+    return pp::Var(static_cast<int32_t>(integer_value));
+  return pp::Var(static_cast<double>(integer_value));
+}
+
+pp::VarArrayBuffer CreateVarArrayBuffer(
+    const Value::BinaryStorage& binary_storage) {
+  pp::VarArrayBuffer var_array_buffer(binary_storage.size());
+  if (!binary_storage.empty()) {
+    std::memcpy(var_array_buffer.Map(), &binary_storage.front(),
+                binary_storage.size());
+  }
+  var_array_buffer.Unmap();
+  return var_array_buffer;
+}
+
+pp::VarDictionary CreateVarDictionary(const Value& value) {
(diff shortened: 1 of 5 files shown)


common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.cc
```diff
@@ -0,0 +1,105 @@
...
+    case Value::Type::kFloat:
+      return pp::Var(value.GetFloat());
+    case Value::Type::kString:
+      return pp::Var(value.GetString());
+    case Value::Type::kBinary:
+      return CreateVarArrayBuffer(value.GetBinary());
+    case Value::Type::kDictionary:
+      return CreateVarDictionary(value);
+    case Value::Type::kArray:
+      return CreateVarArray(value);
```
MahmoudAGawad, line 100: same:  why not passing value.GetArray()?

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.cc
```diff
@@ -0,0 +1,105 @@
...
+    var_dictionary.Set(item_key, ConvertValueToPpVar(*item_value));
+  }
+  return var_dictionary;
+}
+
+pp::VarArray CreateVarArray(const Value& value) {
+  GOOGLE_SMART_CARD_CHECK(value.is_array());
+  const Value::ArrayStorage& array_storage = value.GetArray();
+  pp::VarArray var_array;
+  var_array.SetLength(array_storage.size());
```
MahmoudAGawad, line 72: nit: GOOGLE_SMART_CARD_CHECK(var_array.SetLength(array_storage.size()));

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.cc
```diff
@@ -0,0 +1,105 @@
...
+}
+
+pp::VarArrayBuffer CreateVarArrayBuffer(
+    const Value::BinaryStorage& binary_storage) {
+  pp::VarArrayBuffer var_array_buffer(binary_storage.size());
+  if (!binary_storage.empty()) {
+    std::memcpy(var_array_buffer.Map(), &binary_storage.front(),
+                binary_storage.size());
+  }
+  var_array_buffer.Unmap();
```
MahmoudAGawad, line 53: should we move this inside the enclosing bracket as Map()? It is not clear what will happen if .Unmap() is called without .Map() being called first.

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion_unittest.cc
```diff
@@ -0,0 +1,142 @@
...
+  EXPECT_EQ(var.AsString(), kString);
+}
+
+TEST(ValueNaclPpVarConversion, BinaryValue) {
+  const std::vector<uint8_t> kBinary = {1, 2, 3};
+  const pp::Var var = ConvertValueToPpVar(Value(kBinary));
+  ASSERT_TRUE(var.is_array_buffer());
+  pp::VarArrayBuffer var_array_buffer(var);
+  EXPECT_EQ(var_array_buffer.ByteLength(), kBinary.size());
+  const void* const buffer_contents = var_array_buffer.Map();
```
MahmoudAGawad, line 90: should we always call Unmap() after Map()?

MahmoudAGawad (approved):
LGTM % nits.

Can you also double check the commit message `Implement helper function that transforms a Native Client SDK's pp::Var object into a Value object.`? I see the helper function transforms a Value object into pp::Var.

Fabian-Sommer (approved):
LGTM

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.cc
```diff
@@ -0,0 +1,105 @@
...
+    case Value::Type::kInteger:
+      return CreateIntegerVar(value.GetInteger());
+    case Value::Type::kFloat:
+      return pp::Var(value.GetFloat());
+    case Value::Type::kString:
+      return pp::Var(value.GetString());
+    case Value::Type::kBinary:
+      return CreateVarArrayBuffer(value.GetBinary());
+    case Value::Type::kDictionary:
+      return CreateVarDictionary(value);
```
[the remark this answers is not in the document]
emaxx-google: Good idea.

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.cc
emaxx-google: Good catch!

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion_unittest.cc
emaxx-google: Not strictly required (as this will be done when the variable goes out of scope), but let's do that for consistency.

emaxx-google:
Thanks! I've also added a couple of more tests.

emaxx-google:
> Can you also double check the commit message `Implement helper function that transforms a Native Client SDK's pp::Var object into a Value object.`? I see the helper function transforms a Value object into pp::Var.

You're right, I messed up the commit message. Thanks for noticing this!

PR #187, opened 2020-10-14
Apply clang-format and fix minor style issues

emaxx-google:
Reformat all C++ code written in this project (i.e., not verbatim
imported from third-party projects) according to clang-format with
"Google" Style Guide.

Additionally manually do a few other style fixes that aren't performed
automatically by clang-format, namely:

* Use "constexpr" instead of "const" whenever possible;
* Put constants into unnamed namespaces;
* Explicitly denote C++ classes as copyable/movable/non-movable;
* Initialize class members in the class definition when possible.

### DIFF
--- a/common/cpp/src/google_smart_card_common/external_logs_printer.cc
+++ b/common/cpp/src/google_smart_card_common/external_logs_printer.cc
@@ -55,7 +55,7 @@ void StructConverter<ExternalLogMessageData>::VisitFields(
 
 ExternalLogsPrinter::ExternalLogsPrinter(
     const std::string& listened_message_type)
-  : listened_message_type_(listened_message_type) {}
+    : listened_message_type_(listened_message_type) {}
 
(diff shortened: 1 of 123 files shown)


emaxx-google:
PTAL. This is a pretty much mechanical change.

Fabian-Sommer (approved):
LGTM.
Does it make sense to add rules to Makefiles to format all changes?

emaxx-google:
> Does it make sense to add rules to Makefiles to format all changes?

This might be cool, but won't be trivial. Unfortunately, clang-format messes up the include order in many cases (e.g., it doesn't understand that `#include <google_smart_card_common/foo.h>` should come first in "foo.cc", and also it moves into the same group system-level and nacl-specific .h files, etc.). So, unfortunately, this has to be a manual process - until we find a way to tweak clang-format to do everything properly.

PR #188, opened 2020-10-15
Add conversion from NaCl's pp::Var into Value

emaxx-google:
Implement helper function that transforms a Native Client SDK's
pp::Var object into a Value object.

This helper will be used in follow-ups in order to make most of the
//common/cpp library be toolchain-independent and work under both Native
Client and WebAssembly (as tracker by #185), since the Value class is
toolchain-independent.

### DIFF
--- a/common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.cc
+++ b/common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.cc
@@ -18,6 +18,7 @@
 
 #include <cstring>
 #include <limits>
+#include <memory>
 #include <string>
 #include <utility>
 
@@ -28,15 +29,24 @@
 
 #include <google_smart_card_common/formatting.h>
 #include <google_smart_card_common/logging/logging.h>
+#include <google_smart_card_common/optional.h>
 #include <google_smart_card_common/unique_ptr_utils.h>
 #include <google_smart_card_common/value.h>
 
 namespace google_smart_card {
 
-namespace {
+namespace internal {
 
-constexpr char kUnsupportedPpVarTypeError[] =
+const char kUnsupportedPpVarTypeConversionError[] =
     "Error converting: unsupported type \"%s\"";
+const char kPpVarDictionaryItemConversionError[] =
+    "Error converting dictionary item \"%s\": %s";
+const char kPpVarArrayItemConversionError[] =
+    "Error converting array item #%d: %s";
+
+}  // namespace internal
+
+namespace {
 
 pp::Var CreateIntegerVar(int64_t integer_value) {
   // `pp::Var` can only hold 32-bit integers. Larger numbers need to be
@@ -89,7 +99,7 @@ optional<Value> CreateValueFromPpVarArray(const pp::VarArray& var,
     if (!converted_item) {
       if (error_message) {
         *error_message = FormatPrintfTemplate(
-            "Error converting array item #%d: %s", static_cast<int>(index),
+            internal::kPpVarArrayItemConversionError, static_cast<int>(index),
             error_message->c_str());
       }
       return {};
@@ -112,7 +122,7 @@ optional<Value> CreateValueFromPpVarDictionary(const pp::VarDictionary& var,
     if (!converted_item_value) {
       if (error_message) {
         *error_message = FormatPrintfTemplate(
-            "Error converting dictionary item \"%s\": %s",
+            internal::kPpVarDictionaryItemConversionError,
             item_key.AsString().c_str(), error_message->c_str());
       }
       return {};
@@ -156,14 +166,14 @@ pp::Var ConvertValueToPpVar(const Value& value) {
 
 optional<Value> ConvertPpVarToValue(const pp::Var& var,
                                     std::string* error_message) {
-  if (var.is_undefined()) return Value();
-  if (var.is_null()) return Value();
+  if (var.is_undefined() || var.is_null()) return Value();
   if (var.is_bool()) return Value(var.AsBool());
   if (var.is_string()) return Value(var.AsString());
-  if (var.is_object()) {
+  if (var.is_object() || var.is_resource()) {
     if (error_message) {
       *error_message =
-          FormatPrintfTemplate(kUnsupportedPpVarTypeError, "object");
+          FormatPrintfTemplate(internal::kUnsupportedPpVarTypeConversionError,
+                               var.is_object() ? "object" : "resource");
     }
     return {};
   }
@@ -172,13 +182,6 @@ optional<Value> ConvertPpVarToValue(const pp::Var& var,
   if (var.is_dictionary()) {
     return CreateValueFromPpVarDictionary(pp::VarDictionary(var),
                                           error_message);
-  }
-  if (var.is_resource()) {
-    if (error_message) {
-      *error_message =
-          FormatPrintfTemplate(kUnsupportedPpVarTypeError, "resource");
-    }
-    return {};
   }
   if (var.is_int()) return Value(var.AsInt());
   if (var.is_double()) return Value(var.AsDouble());
--- a/common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.cc
+++ b/common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.cc
@@ -35,17 +35,6 @@
 
 namespace google_smart_card {
 
-namespace internal {
-
-const char kUnsupportedPpVarTypeConversionError[] =
-    "Error converting: unsupported type \"%s\"";
-const char kPpVarDictionaryItemConversionError[] =
-    "Error converting dictionary item \"%s\": %s";
-const char kPpVarArrayItemConversionError[] =
-    "Error converting array item #%d: %s";
-
-}  // namespace internal
-
 namespace {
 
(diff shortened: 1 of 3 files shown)


common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.cc
```diff
@@ -100,4 +154,37 @@ pp::Var ConvertValueToPpVar(const Value& value) {
...
+  }
+  if (var.is_resource()) {
+    if (error_message) {
+      *error_message =
+          FormatPrintfTemplate(kUnsupportedPpVarTypeError, "resource");
+    }
+    return {};
+  }
+  if (var.is_int()) return Value(var.AsInt());
+  if (var.is_double()) return Value(var.AsDouble());
```
emaxx-google, line 184: Open question: Should we actually return an integer Value in case the passe double value is actually an integer and fits within the range of precisely representable integer numbers (which is typically 53 bits)?

I do this in the existing code that, for example, converts between `pp::Var` and `int64_t`, but in that code we always know the target type. Here in this PR we don't know how the resulting `Value` will be used (as a double or as an integer), so it's up to us whether we want to try attempting the casting early...

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.h
```diff
@@ -18,15 +18,26 @@
...
 
 namespace google_smart_card {
 
 // Converts the given `Value` into a `pp::Var`.
 pp::Var ConvertValueToPpVar(const Value& value);
 
+// Converts the given `pp::Var` into a `Value`.
+//
+// When the conversion isn't possible (e.g., when the passed variable contains a
+// Pepper Resource object), returns a null optional and, if provided, sets
```
MahmoudAGawad, line 36: nit: `pp:Resource`

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.cc
```diff
@@ -100,4 +154,37 @@ pp::Var ConvertValueToPpVar(const Value& value) {
...
+    return CreateValueFromPpVarDictionary(pp::VarDictionary(var),
+                                          error_message);
+  }
+  if (var.is_resource()) {
+    if (error_message) {
+      *error_message =
+          FormatPrintfTemplate(kUnsupportedPpVarTypeError, "resource");
+    }
+    return {};
+  }
```
MahmoudAGawad, line 182: can be merged with `if (var.is_object()) {` in line 163

MahmoudAGawad, line 182: can be merged with line 163.

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion_unittest.cc
```diff
@@ -186,4 +188,269 @@ TEST(ValueNaclPpVarConversion, ArrayValue) {
...
+  {
+    const optional<Value> value = ConvertPpVarToValue(pp::Var(pp::Resource()));
+    EXPECT_FALSE(value);
+  }
+
+  {
+    std::string error_message;
+    const optional<Value> value =
+        ConvertPpVarToValue(pp::Var(pp::Resource()), &error_message);
+    EXPECT_THAT(error_message, ::testing::HasSubstr("resource"));
```
MahmoudAGawad, line 275: can we use `FormatPrintfTemplate(kUnsupportedPpVarTypeError, "resource");` here?

MahmoudAGawad (approved):
LGTM % nits

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.cc
MahmoudAGawad: Good question.

I think it depends on the user expectation of such functionality as Value::Type would be different in both cases.

IMHO, I am leaning towards keeping it as it is (return Value::Type::kFloat value) and let the user decide what they want to do with the value.

The only use-case I find it beneficial to return integer is: since pp:Var only supports int32_t, the user will have to use double instead, in this case we give the user a ready-to-use Value::Type. (Instead they would need to cast).
But at the same time, we would hurt other use-cases where the user really expects Value::Type::kFloat (even if the value happens to be an integer).

MahmoudAGawad (approved):

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion_unittest.cc
emaxx-google: Done. I originally wanted to avoid writing complex code for verifying those messages, but perhaps it made tests weirder indeed...

emaxx-google (commented):
Thanks!

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.cc
Fabian-Sommer: I also think we should return a double here - it would be confusing if one converts a double and in some cases it comes out as int.

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion_unittest.cc
```diff
@@ -186,4 +195,295 @@ TEST(ValueNaclPpVarConversion, ArrayValue) {
...
+    const optional<Value> value = ConvertPpVarToValue(pp::Var(pp::Resource()));
+    EXPECT_FALSE(value);
+  }
+
+  {
+    std::string error_message;
+    const optional<Value> value =
+        ConvertPpVarToValue(pp::Var(pp::Resource()), &error_message);
+    EXPECT_EQ(error_message,
+              FormatPrintfTemplate(
```
Fabian-Sommer, line 283: nit: Wouldn't it make more sense to hardcode the string here? This would additionally catch a weirdly formatted error message.

MahmoudAGawad:
Thanks you! Still LGTM.

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion_unittest.cc
```diff
@@ -186,4 +195,295 @@ TEST(ValueNaclPpVarConversion, ArrayValue) {
...
+    EXPECT_FALSE(value);
+  }
+
+  {
+    std::string error_message;
+    const optional<Value> value =
+        ConvertPpVarToValue(inner_var_dict, &error_message);
+    EXPECT_EQ(
+        error_message,
+        FormatPrintfTemplate(
```
Fabian-Sommer, line 380: ditto

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion_unittest.cc
```diff
@@ -186,4 +195,295 @@ TEST(ValueNaclPpVarConversion, ArrayValue) {
...
+                .c_str()));
+    EXPECT_FALSE(value);
+  }
+
+  {
+    std::string error_message;
+    const optional<Value> value = ConvertPpVarToValue(var_dict, &error_message);
+    EXPECT_EQ(
+        error_message,
+        FormatPrintfTemplate(
```
Fabian-Sommer, line 393: ditto

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion_unittest.cc
```diff
@@ -186,4 +195,295 @@ TEST(ValueNaclPpVarConversion, ArrayValue) {
...
+    EXPECT_FALSE(value);
+  }
+
+  {
+    std::string error_message;
+    const optional<Value> value =
+        ConvertPpVarToValue(inner_var_array, &error_message);
+    EXPECT_EQ(
+        error_message,
+        FormatPrintfTemplate(
```
Fabian-Sommer, line 463: ditto

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion_unittest.cc
```diff
@@ -186,4 +195,295 @@ TEST(ValueNaclPpVarConversion, ArrayValue) {
...
+    EXPECT_FALSE(value);
+  }
+
+  {
+    std::string error_message;
+    const optional<Value> value =
+        ConvertPpVarToValue(var_array, &error_message);
+    EXPECT_EQ(
+        error_message,
+        FormatPrintfTemplate(
```
Fabian-Sommer, line 477: ditto

Fabian-Sommer (approved):
LGTM % nits

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion_unittest.cc
emaxx-google: Done.

common/cpp/src/google_smart_card_common/value_nacl_pp_var_conversion.cc
emaxx-google: Thanks, Mahmoud and Fabian. That makes sense, I'll keep that behavior in the PR.

PR #189, opened 2020-10-15
Add Value debug dumping helpers

emaxx-google:
Add functions that generate debug dumps of Value instances, for logging
purposes.

This is equivalent to the currently existing helpers in
//common/cpp/src/pp_var_utils/debug_dump.h. As in those existing
helpers, two versions are provided: the full dump and the privacy-safe
compact dump.

These helpers for Value instances is a preparation step for migrating the
//common/cpp library under WebAssembly (as tracked by #185).

### DIFF
--- a/common/cpp/src/google_smart_card_common/value_debug_dumping.cc
+++ b/common/cpp/src/google_smart_card_common/value_debug_dumping.cc
@@ -1,0 +1,124 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <google_smart_card_common/value_debug_dumping.h>
+
+#include <string>
+
+#include <google_smart_card_common/logging/hex_dumping.h>
+#include <google_smart_card_common/logging/logging.h>
+#include <google_smart_card_common/value.h>
+
+namespace google_smart_card {
+
+namespace {
+
+std::string GetValueTypeTitle(const Value& value) {
+  switch (value.type()) {
+    case Value::Type::kNull:
+      return Value::kNullTypeTitle;
+    case Value::Type::kBoolean:
+      return Value::kBooleanTypeTitle;
+    case Value::Type::kInteger:
+      return Value::kIntegerTypeTitle;
+    case Value::Type::kFloat:
+      return Value::kFloatTypeTitle;
+    case Value::Type::kString:
+      return Value::kStringTypeTitle;
+    case Value::Type::kBinary:
+      return Value::kBinaryTypeTitle;
+    case Value::Type::kDictionary:
+      return Value::kDictionaryTypeTitle;
+    case Value::Type::kArray:
+      return Value::kArrayTypeTitle;
+  }
+  GOOGLE_SMART_CARD_NOTREACHED;
+}
+
+std::string DebugDumpBoolean(bool value) { return value ? "true" : "false"; }
+
+std::string DebugDumpString(const std::string& value) {
+  return '"' + value + '"';
+}
+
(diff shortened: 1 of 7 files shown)


common/cpp/src/google_smart_card_common/value_debug_dumping.h
```diff
@@ -0,0 +1,38 @@
...
+// builds, this only dumps the title of the value's type; in Debug builds, this
+// is equivalent to `DebugDumpValueFull()`.
+std::string DebugDumpValueSanitized(const Value& value);
+
+// Generates a full debug representation of the given value.
+//
+// NOTE: It's dangerous to use this function with variables that might
+// potentially contain privacy/security sensitive data. Use
+// `DebugDumpValueSanitized()` instead.
+std::string DebugDumpValueFull(const Value& value);
```
Fabian-Sommer, line 34: Would it make sense to only define this if NDEBUG is not set?

common/cpp/src/google_smart_card_common/value_debug_dumping_unittest.cc
```diff
@@ -0,0 +1,193 @@
...
+  EXPECT_EQ(sanitized_false, Value::kBooleanTypeTitle);
+  EXPECT_EQ(sanitized_true, Value::kBooleanTypeTitle);
+#else
+  EXPECT_EQ(sanitized_false, full_false);
+  EXPECT_EQ(sanitized_true, full_true);
+#endif
+}
+
+TEST(ValueDebugDumpingTest, Integer) {
+  constexpr int64_t kInteger = 123;
```
Fabian-Sommer, line 56: nit: Again I think that the expected results should be hardcoded here, making the expectations explicit and allowing to catch errors in the underlying functions.

common/cpp/src/google_smart_card_common/value_debug_dumping.cc
```diff
@@ -0,0 +1,124 @@
...
+    result += DebugDumpString(item_key);
+    result += ": ";
+    result += DebugDumpValueFull(item_value);
+  }
+  result += "}";
+  return result;
+}
+
+std::string DebugDumpBinary(const Value::BinaryStorage& binary_value) {
+  std::string result = std::string(Value::kBinaryTypeTitle) + "[";
```
MahmoudAGawad, line 83: nit: This is completely not needed (and more verbose), feel free to drop it.

I see you are **only** prefixing the Value::BinaryStorage with its type string `std::string(Value::kBinaryTypeTitle)` (I presume this is because you want to differentiate between Array and Binary types, no?). Would it be useful to add the type before each DebugDump*, as well? for example, in `DebugDumpBoolean(bool value) { return std::string(Value::kBooleanTypeTitle) + ": " + (value ? "true" : "false"); }`

common/cpp/src/google_smart_card_common/value_debug_dumping_unittest.cc
Fabian-Sommer: (Here and in other places in this file)

Fabian-Sommer (approved):
LGTM % nits

MahmoudAGawad (approved):
LGTM % nit

common/cpp/src/google_smart_card_common/value_debug_dumping.cc
emaxx-google: Yeah, I was trying to keep these log strings reasonably short and clean. Things like `true` can be undoubtedly interpreted by the human as "boolean with the true value", so we don't really need the type information there. I understand your idea about being more consistent, but would prefer keeping the more compact format with one special case for binary data.
But I've added a comment about this into the code.

common/cpp/src/google_smart_card_common/value_debug_dumping.h
emaxx-google: Ack - The idea is that this function should be used in places where the caller knows for sure that the data is non-sensitive.

For example, the existing code does this when logging parameters for the `chrome.certificateProvider.requestPin()` method: https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/blob/4d02c5320eb824fad87e377d28e2fb87218180aa/example_cpp_smart_card_client_app/src/chrome_certificate_provider/api_bridge.cc#L140
(A side note: In the existing code these `pp::Var`-based helpers are named `DumpVar()` and `DebugDumpVar()`, but I guess these were not really informative names. Hopefully, the naming proposed in this PR will be less ambiguous and harder to confuse.)

emaxx-google (commented):
Thanks!

emaxx-google:
I'm assuming no one has concerns about the latest patchset, so will proceed with landing it. Thanks for the reviews!

PR #190, opened 2020-10-15
Add optional::reset() method

emaxx-google:
Extend our polyfill for the "optional" C++ class to have the reset()
method, and use it in places where the optional was reset by other
means.

Historically, the original C++ Standard proposal for the std::optional
type didn't have the reset() method, which is why we didn't add into our
polyfill as well. However, afterwards this method was added into the
final std::optional interface, which went into C++17.

In the future, once we fully migrate to WebAssembly and drop the Google
Native Client support, we'll be able to delete the whole polyfill class
and use the "optional" class provided by the C++ Standard Library.

### DIFF
--- a/common/cpp/src/google_smart_card_common/optional.h
+++ b/common/cpp/src/google_smart_card_common/optional.h
@@ -80,6 +80,8 @@ class optional final {
 
   T&& value() && { return operator*(); }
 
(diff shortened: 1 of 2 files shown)


Fabian-Sommer (approved):
LGTM

PR #191, opened 2020-10-15
Use -Wall -pedantic in more places

emaxx-google:
Add "-Wall" and "-pedantic" C/C++ compilation flags in all places where
this doesn't break the compilation. Suppress particular warnings in a
couple of places where they are expected or hard to fix.

This should help maintaining the code base by using compiler built-in
diagnostics to a larger extent.


Fabian-Sommer (approved):
LGTM

PR #193, opened 2020-10-19
Add trivial conversions for Value to/from C++ type

emaxx-google:
Implement set of helper functions for trivial conversions between a
Value object and a C++ type (bool, number or string).

These helpers, once fully implemented for all other needed types, will
be used to migrate the rest of the //common/cpp library onto the
toolchain-independent Value class (as tracked by #185). Currently
many parts of this library use the NaCl-specific pp::Var class and the
previously implemented helpers for converting to/from pp::Var objects.

### DIFF
--- a/common/cpp/src/google_smart_card_common/numeric_conversions.cc
+++ b/common/cpp/src/google_smart_card_common/numeric_conversions.cc
@@ -21,8 +21,8 @@ namespace google_smart_card {
 namespace internal {
 
 // Definitions of the constants declared in the header file
-constexpr int64_t kDoubleExactRangeMax =
-    1LL << std::numeric_limits<double>::digits;
+constexpr int64_t kDoubleExactRangeMax = 1LL
+                                         << std::numeric_limits<double>::digits;
 constexpr int64_t kDoubleExactRangeMin =
     -(1LL << std::numeric_limits<double>::digits);
 
@@ -32,10 +32,14 @@ bool CastDoubleToInt64(double value, int64_t* result,
                        std::string* error_message) {
   if (!(internal::kDoubleExactRangeMin <= value &&
         value <= internal::kDoubleExactRangeMax)) {
-    *error_message = FormatPrintfTemplate(
-        "The real value is outside the exact integer representation range: %f "
-        "not in [%" PRId64 "; %" PRId64 "]",
-        value, internal::kDoubleExactRangeMin, internal::kDoubleExactRangeMax);
+    if (error_message) {
+      *error_message = FormatPrintfTemplate(
+          "The real value is outside the exact integer representation range: "
+          "%f "
+          "not in [%" PRId64 "; %" PRId64 "]",
+          value, internal::kDoubleExactRangeMin,
+          internal::kDoubleExactRangeMax);
+    }
     return false;
   }
   *result = static_cast<int64_t>(value);
--- a/common/cpp/src/google_smart_card_common/value.h
+++ b/common/cpp/src/google_smart_card_common/value.h
@@ -74,7 +74,7 @@ class Value final {
   explicit Value(ArrayStorage array_value);
   // Forbid construction from pointers other than `const char*`. Without this
   // deleted overload, the `bool`-argument version would be silently picked up.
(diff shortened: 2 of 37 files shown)


emaxx-google:
PTAL. There are gaps in test coverage, but I'd like to tackle them after doing the first round of reviews, since there'll be a bunch of copy-paste between some of tests.

common/cpp/src/google_smart_card_common/value.h
```diff
@@ -72,10 +72,15 @@ class Value final {
   explicit Value(BinaryStorage binary_value);
   explicit Value(DictionaryStorage dictionary_value);
   explicit Value(ArrayStorage array_value);
+  // Forbid construction from pointers other than `const char*`. Without this
+  // deleted overload, the `bool`-argument version would be silently picked up.
+  explicit Value(void*) = delete;
```
emaxx-google, line 77: Self-review: this actually has to be `const void*` in order to catch more cases. Ditto in value_conversion.h.

common/cpp/src/google_smart_card_common/numeric_conversions.cc
```diff
@@ -32,10 +32,14 @@ bool CastDoubleToInt64(double value, int64_t* result,
...
+      *error_message = FormatPrintfTemplate(
+          "The real value is outside the exact integer representation range: "
+          "%f "
+          "not in [%" PRId64 "; %" PRId64 "]",
+          value, internal::kDoubleExactRangeMin,
+          internal::kDoubleExactRangeMax);
+    }
     return false;
   }
   *result = static_cast<int64_t>(value);
```
emaxx-google, line 45: In case you're wondering: yes, apparently I don't validate the double value to have the zero (or "almost zero") fractional part. This is orthogonal to the current WebAssembly effort, but probably we'll need to tighten this function to refuse fractional numbers.

Fabian-Sommer (approved):
LGTM

MahmoudAGawad (approved):
LGTM.

common/cpp/src/google_smart_card_common/value_conversion.cc
```diff
@@ -0,0 +1,158 @@
...
+  // The Standard doesn't give an upper bound on the `unsigned` size, so be on
+  // the safe side (despite that a typical implementation has it <=4 bytes).
+  GOOGLE_SMART_CARD_CHECK(
+      CompareIntegers(number, std::numeric_limits<int64_t>::max()) <= 0);
+  return Value(static_cast<int64_t>(number));
+}
+
+Value ConvertToValue(unsigned long number) {
+  // In case `unsigned long` is an 8-byte type, it's not possible to convert
+  // some of it value (extremely huge ones).
```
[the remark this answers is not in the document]
emaxx-google: Yup, I believe it won't be used by any of the code in the project. Anyway, if we discover to need it, we can easily add more helpers here at any time.

common/cpp/src/google_smart_card_common/value_conversion.cc
```diff
@@ -0,0 +1,158 @@
...
+  }
+  return CastInt64ToInteger(int64_number, type_name, number, error_message);
+}
+
+}  // namespace
+
+Value ConvertToValue(unsigned number) {
+  // The Standard doesn't give an upper bound on the `unsigned` size, so be on
+  // the safe side (despite that a typical implementation has it <=4 bytes).
+  GOOGLE_SMART_CARD_CHECK(
```
emaxx-google, line 63: Actually, folks, WDYT - given these error scenarios, should I also make the `ConvertToValue()` functions be able to report the error rather than crash?

This would also be consistent with the `ConvertFromValue()` interface.

MahmoudAGawad: I think this is a good idea.

emaxx-google: Done

emaxx-google:
I got rid of the `optional<>` overloads, since they're easy to confuse with non-optional overloads and clutter tests, and also most future callsites won't benefit from `optional<>` at all.

PR #194, opened 2020-10-19
Add conversions for Value to/from C++ enums

emaxx-google:
Implement helpers for semi-automated implementation that maps between
number-based C++ enums and string-based Value enums (in practice, the
latter will typically come from the JavaScript side).

This commit implements analogous functionality similar to the already
existing Native Client pp::Var specific helpers (in
pp_var_utils/enum_converter.h), however this new implementation requires
much less boilerplate code to be written for every enum.


emaxx-google:
Folks, PTAL at the next CL in the chain. Thanks in advance!

Fabian-Sommer (changes requested):

common/cpp/src/google_smart_card_common/value_conversion.h
```diff
@@ -106,8 +233,21 @@ inline bool ConvertToValue(std::string characters, Value* value,
...
 // overload, the `bool`-argument overload would be silently picked up.
-Value ConvertToValue(const void* pointer_value, Value* value,
-                     std::string* error_message = nullptr) = delete;
+bool ConvertToValue(const void* pointer_value, Value* value,
+                    std::string* error_message = nullptr) = delete;
+
+// Converts from an enum into a string `Value`. The enum type has to be
+// registered via the `EnumValueDescriptor` class.
+template <typename T>
+typename std::enable_if<std::is_enum<T>::value, bool>::type ConvertToValue(
```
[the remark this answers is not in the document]
emaxx-google: Added a comment.

emaxx-google (commented):
Thanks!

Fabian-Sommer (approved):
LGTM

common/cpp/src/google_smart_card_common/value_conversion.h
```diff
@@ -27,21 +27,153 @@
...
+          from_value_converter_(from_value_converter) {}
+    Description(Description&&) = default;
+    Description& operator=(Description&&) = default;
+    ~Description() = default;
+
+    const char* type_name() const { return type_name_; }
+
+    // Adds the given item into the enum's description: |enum_item_name| is the
+    // `Value` representation of |enum_item|.
+    // Returns a rvalue reference to |this| and uses the "&&" ref-qualifier, so
```
MahmoudAGawad, line 135: typo: Returns ~a~ **an** rvalue

common/cpp/src/google_smart_card_common/value_conversion.h
```diff
@@ -27,21 +27,153 @@
...
+    // that the method calls can be easily chained and the final result can be
+    // returned without an explicit std::move() boilerplate.
+    Description&& WithItem(T enum_item, const char* enum_item_name) && {
+      const int64_t enum_item_number = static_cast<int64_t>(enum_item);
+      if (to_value_converter_)
+        to_value_converter_->HandleItem(enum_item_number, enum_item_name);
+      else
+        from_value_converter_->HandleItem(enum_item_number, enum_item_name);
```
MahmoudAGawad, line 143: what if from_value_converter is nullptr?

common/cpp/src/google_smart_card_common/value_conversion.h
Fabian-Sommer: I think a is correct here

MahmoudAGawad: as a quick reference, please take a look at https://screenshot.googleplex.com/9QFaerZLyEM72G2 [Sorry, Google internal only]

Also take a look here:
https://ontariotraining.net/grammar-tip-a-or-an-with-acronyms/#:~:text=When%20it%20comes%20to%20abbreviations,first%20sound%20is%20a%20consonant.&text=Therefore%2C%20both%20%E2%80%9Ca%20Law%20Society,an%20LSUC%20form%E2%80%9D%20are%20correct

Fabian-Sommer: :thumbsup:

common/cpp/src/google_smart_card_common/value_conversion.h
```diff
@@ -27,21 +27,153 @@
...
+  static_assert(std::is_enum<T>::value,
+                "The EnumValueDescriptor parameter must be an enum");
+
+  // Class that should be used for describing items of the enum type. Should be
+  // instantiated via the `Describe()` method.
+  class Description final {
+   public:
+    Description(const char* type_name,
+                internal::EnumToValueConverter* to_value_converter,
+                internal::EnumFromValueConverter* from_value_converter)
```
MahmoudAGawad, line 123: I may be missing something but does Description class needs to know about whether it is EnumToValue or EnumFromValue converter? If it doesn't need to know, then we might have a super class `EnumValueConverter` that both `EnumToValueConverter` and EnumFromValueConverter` inherit from. and we use this class to pass on one instance and verify it is not null.

MahmoudAGawad: Please, discard.

ooh I missed that this class can be used to convert to and from at the same time.

MahmoudAGawad (approved):
LGTM

common/cpp/src/google_smart_card_common/value_conversion.h
emaxx-google: Done - Even though exactly one of pointers should be non-null, it costs almost nothing to check it here explicitly.

emaxx-google:
Thanks, folks!

PR #195, opened 2020-10-19
Fix compilation warnings in Value implementation

emaxx-google:
Fix "unused variable" and "unused function" warnings that were somehow
missed while landing previous patchsets (namely, #189 and #193).


Fabian-Sommer (approved):
LGTM

PR #196, opened 2020-10-20
Migrate MessageListener from pp::Var to Value

emaxx-google:
Change the MessageListener C++ interface and all its implementations to
stop using the Native Client's pp::Var class, and use the
toolchain-independent Value class.

This is part of the effort on adding WebAssembly support into
//common/cpp (as tracked by #185). This MessageListener class is,
effectively, an entry point into the C/C++ code of incoming JavaScript
message handling. It's called by toolchain-specific code (currently,
only Native Client specific wrappers, but in the future an alternative
WebAssembly specific wrapper will be added). This commit only affects
one of the first steps of the pipeline - subsequent ones will be done in
follow-up commits.


emaxx-google:
PTAL - this is the first commit that touches the real production code. (Everything so far were preparatory libraries or refactorings.)

This change introduces the first usage of `Value` in the data pipeline; currently it's just one chain in the pipeline, so the `Value` is immediately converted back into `pp::Var`, but follow-up commits will extend the `Value` usage.

common/cpp/src/google_smart_card_common/messaging/message_listener.h
```diff
@@ -26,8 +28,10 @@ class MessageListener {
 
   // Called when a message is received.
   //
-  // Returns whether the message was handled.
-  virtual bool OnMessageReceived(const pp::Var& message) = 0;
+  // Returns whether the message was handled; in case it was not, the error is
```
Fabian-Sommer, line 31: nit: `an error` instead of `the error`

Fabian-Sommer (approved):
LGTM

common/cpp/src/google_smart_card_common/messaging/typed_message_router.cc
```diff
@@ -43,15 +48,36 @@ void TypedMessageRouter::RemoveRoute(TypedMessageListener* listener) {
...
+
   TypedMessage typed_message;
-  if (!VarAs(message, &typed_message, &error_message)) return false;
+  // TODO: Delete `local_error_message` in favor of (optional) `error_message`.
+  std::string local_error_message;
+  if (!VarAs(pp_message, &typed_message, &local_error_message)) {
+    if (error_message) {
+      *error_message = FormatPrintfTemplate("Cannot parse typed message: %s",
+                                            local_error_message.c_str());
+    }
```
MahmoudAGawad, line 63: nit: this is not sepcific to this CL. These 3 lines are repeated in quite some places, and is kinda interfering with the main logic.

As a suggestion (feel free to discard): Can we encapsulate the if inside the `FormatPrintfTemplate` itself and pass the error_message parameter along: it would be something like: `FormatPrintfTemplate(char* error_message, const char* format, ...)`

MahmoudAGawad (approved):
LGTM.

common/cpp/src/google_smart_card_common/messaging/typed_message_router.cc
emaxx-google: That's a good point.

The downside will be that the formatting arguments will always be calculated, but, practically, the performance penalty should be negligible and only affecting "unsuccessful" flows. Also there are some tricky cases like `FormatPrintfTemplate(error_message, ..., error_message->c_str())` that would crash if we dereference the null, but those can be solved by putting that nested error into a separate local variable.

So overall I agree that helper is what we should do; will prepare a separate PR soon. Thanks.

emaxx-google (commented):
Thanks!

PR #197, opened 2020-10-21
Helper for creating error messages if requested

emaxx-google:
Add another variant of the FormatPrintfTemplate() C++ function that
writes its result to the given pointer, after checking the latter for
being non-null.

This allows to simplify code to avoid writing boilerplate "if
(error_message)" conditions. This change isn't expected to cause any
real-world performance regressions, despite generating error message
parameters even in cases when the result is not used, since it only
affects "unsuccessful" code paths that should only be triggered rarely.

### DIFF
--- a/common/cpp/src/google_smart_card_common/formatting.cc
+++ b/common/cpp/src/google_smart_card_common/formatting.cc
@@ -15,6 +15,7 @@
 #include <google_smart_card_common/formatting.h>
 
 #include <cstdio>
+#include <utility>
 #include <vector>
 
 namespace google_smart_card {
--- a/common/cpp/src/google_smart_card_common/formatting.cc
+++ b/common/cpp/src/google_smart_card_common/formatting.cc
@@ -51,13 +51,11 @@ std::string FormatPrintfTemplate(const char* format, va_list var_args) {
 
 void FormatPrintfTemplateAndSet(std::string* output_string, const char* format,
                                 ...) {
+  if (!output_string) return;
   va_list var_args;
   va_start(var_args, format);
-  std::string formatted = FormatPrintfTemplate(format, var_args);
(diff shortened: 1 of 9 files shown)


emaxx-google:
PTAL - this is a follow-up to https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/pull/196#discussion_r508821591 .

Fabian-Sommer (approved):
LGTM

common/cpp/src/google_smart_card_common/formatting.cc
```diff
@@ -48,4 +49,15 @@ std::string FormatPrintfTemplate(const char* format, va_list var_args) {
...
 
+void FormatPrintfTemplateAndSet(std::string* output_string, const char* format,
+                                ...) {
+  va_list var_args;
+  va_start(var_args, format);
+  std::string formatted = FormatPrintfTemplate(format, var_args);
+  va_end(var_args);
+  // Note: It's incorrect to move this check to the beginning of the function,
+  // since we need to read the variable arguments off the stack in any case.
+  if (output_string) *output_string = std::move(formatted);
```
MahmoudAGawad, line 60: nit: maybe return early if output_string is null already to avoid doing the above logic regardless.

MahmoudAGawad (approved):
Thank you for the change!

LGTM % nit.

common/cpp/src/google_smart_card_common/formatting.cc
emaxx-google: Done. I actually thought it's illegal to do so (assuming that the stack would be corrupted if we don't ask `printf()` to read all arguments), but after some googling it seems to be safe to do so.
I've also added a unit test for this, and it passes successfully.

PR #199, opened 2020-10-22
Add identity transform for Value into itself

emaxx-google:
Add overloads for the ConvertToValue/ConvertFromValue functions that
perform identity transform, i.e., return the input argument without
changes.

This is needed in order to make generic transformation code work in case
when the same Value class is used in both source and destination
representation - for example, the TypedMessage struct's |data| field.

This is part of the effort on migrating the //common/cpp library to use
the toolchain-independent Value class, making it work under both
Native Client and WebAssembly (#185).


Fabian-Sommer (approved):
LGTM

emaxx-google:
Thanks for the prompt reply.

PR #200, opened 2020-10-22
Add "OrDie" Value conversion variants

emaxx-google:
Add new variants of the ConvertToValue/ConvertFromValue functions that
immediately crash the program in case the conversion fails.

These functions should be used in cases when the error is really
unexpected (e.g., when the input data is guaranteed to be of specific
form) and/or when there's no recovery possible from the error.

This is part of the effort on migrating the //common/cpp library to use
the toolchain-independent Value class, making it work under both
Native Client and WebAssembly (#185).


emaxx-google:
PTAL - when I was prototyping the migration of the rest of the //common/cpp library, I found it useful to use these helpers, at least temporarily.

Fabian-Sommer (approved):
LGTM

emaxx-google:
Thanks!

PR #201, opened 2020-10-22
Migrate typed_message.h to Value

emaxx-google:
Change the typed_message.h and the corresponding .cc file to use the
Value class and drop all usages of Native Client specific pp::Var
classes from it.

This removes a bunch of boilerplate related to the struct-to-pp::Var and
inverse conversions. Consequently, the unit tests are deleted, since
there's nothing to test here anymore (conversions between structs and
Value objects are covered by a generic unit test).

by #185).


Fabian-Sommer (approved):
LGTM % nit

common/cpp/src/google_smart_card_common/messaging/typed_message_router.cc
```diff
@@ -50,16 +51,12 @@ void TypedMessageRouter::RemoveRoute(TypedMessageListener* listener) {
 
 bool TypedMessageRouter::OnMessageReceived(Value message,
                                            std::string* error_message) {
-  // TODO: Remove this conversion and use `Value` directly.
-  const pp::Var pp_message = ConvertValueToPpVar(message);
-
   TypedMessage typed_message;
-  // TODO: Delete `local_error_message` in favor of (optional) `error_message`.
```
emaxx-google, line 57: As per offline discussion, I've changed it back to `local_error_message`, so that the error message creation code is simpler to understand.

emaxx-google:
Thanks for the quick turnaround!

PR #203, opened 2020-10-23
Migrate non-//common to Value-based TypedMessage

emaxx-google:
Adjust the C++ code outside //common/cpp to the new implementation of
TypedMessage that doesn't support pp::Var anymore.

This CL is a follow-up to #201, and is a stop-gap solution to make the
whole project still compile and work while we're working on the migration
of the //common/cpp library (as tracked by #185). After completion of
that part, we'll clean up the non-//common code to delete all those
back-and-forth pp::Var conversions and to use Value wherever possible.

### DIFF
--- a/common/cpp/build/Makefile
+++ b/common/cpp/build/Makefile
@@ -112,6 +112,7 @@ INSTALLING_HEADERS := \
 	$(ROOT_SOURCES_SUBDIR):$(SOURCES_PATH):tuple_unpacking.h \
 	$(ROOT_SOURCES_SUBDIR):$(SOURCES_PATH):unique_ptr_utils.h \
 	$(ROOT_SOURCES_SUBDIR):$(SOURCES_PATH):value.h \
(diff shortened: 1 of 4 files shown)


common/cpp/build/Makefile
```diff
@@ -112,6 +112,7 @@ INSTALLING_HEADERS := \
 	$(ROOT_SOURCES_SUBDIR):$(SOURCES_PATH):tuple_unpacking.h \
 	$(ROOT_SOURCES_SUBDIR):$(SOURCES_PATH):unique_ptr_utils.h \
 	$(ROOT_SOURCES_SUBDIR):$(SOURCES_PATH):value.h \
+	$(ROOT_SOURCES_SUBDIR):$(SOURCES_PATH):value_conversion.h \
```
emaxx-google, line 115: I discovered that I forgot to add installation of this header.

Fabian-Sommer (approved):
LGTM

PR #204, opened 2020-10-23
Split out toolchain-specific .mk build rules

emaxx-google:
Add single nacl_module_building_common.mk file that provides a
toolchain-independent interface for building C/C++ code, and move the
Native Client specific implementation of this interface into
internal/binary_executable_building_nacl.mk.

This structure is one of the steps of adding the support of WebAssembly
builds (as tracked by #177) while keeping the Native Client support for
the transition period.

This is a mechanical change, except minor changes in the build rules
documentation.

### DIFF
--- a/common/make/binary_executable_building.mk
+++ b/common/make/binary_executable_building.mk
@@ -1,0 +1,90 @@
+# Copyright 2020 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This file contains helper definitions for building an executable binary that
+# can be run in web-based applications.
+#
+# The actually used technology depends on the TOOLCHAIN environment variable.
+# Currently, the only supported value is "pnacl", which corresponds to the
+# Google Native Client technology.
+#
+# common.mk must be included before including this file.
+
+# Default to using the "pnacl" toolchain.
+TOOLCHAIN ?= pnacl
+
+# Common documentation for definitions provided by this file (they are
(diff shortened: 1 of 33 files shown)


common/make/binary_executable_building.mk
```diff
@@ -0,0 +1,90 @@
+# Copyright 2020 Google Inc. All Rights Reserved.
```
emaxx-google, line 1: WDYT about the name of this file?

Other ideas I had were "c_building_common.mk", "c_cpp_building.mk", "native_code_building.mk", etc. But wasn't sure about referring to specific "C"/"C++" languages; as for "native_code" - it seems confusing to refer to the sandboxed NaCl/Wasm code as "native".

common/make/binary_executable_building.mk
```diff
@@ -0,0 +1,90 @@
...
+#   Links object files that correspond to the given source files into an
+#   executable program. The resulting files are put into the out directory.
+#   Arguments:
+#   $1 ("SOURCES"): Source file paths,
+#   $2 ("LIBS", optional): Static libraries to link against,
+#   $3 ("DEPS", optional): Libraries whose building will be added as
+#     dependencies of this target.
+#   $4 ("LDFLAGS", optional): Linker flags.
+#
+# * NACL_LIBRARY_HEADERS_INSTALLATION_DIR_PATH variable:
```
Fabian-Sommer, line 53: nit: If the interface defined in this file should be toolchain-agnostic, it should not define macros that, based on their name, are toolchain-specific.

common/make/binary_executable_building.mk
Fabian-Sommer: Agree on not using "native". I think referring to c/cpp specifically is fine, but wasm does support more than that (NaCl doesn't). So I think the current name is the best (I think it could be shortened to `executable_building.mk` - are there executables that are not binary?).

Fabian-Sommer (approved):
LGTM % nit

common/make/binary_executable_building.mk
emaxx-google: Added TODOs (didn't want to make the current PR excessively big due to renaming all of these).

common/make/binary_executable_building.mk
emaxx-google: Done - renamed to executable_building.mk.

MahmoudAGawad (approved):
LGTM % Fabian's nits

emaxx-google (commented):
Thanks!

PR #205, opened 2020-10-23
Migrate typed_message_listener.h to Value

emaxx-google:
Change the typed_message_listener.h and all child classes to use the
Value class in the interface, rather than using the Native Client
specific pp::Var.

As an additional outcome, the implementation of the
typed_message_router.cc file becomes completely free of pp::Var usages
as well.

by #185).


Fabian-Sommer (approved):
LGTM

emaxx-google:
Thanks.

PR #206, opened 2020-10-26
Migrate requester_message.h to Value

emaxx-google:
Change the requester_message.h header to use the toolchain-independent
Value class instead of Native Client specific pp::Var class. Besides the
mechanical replacement, this commit introduces dedicated structs for the
"request"/"response" messages, since it makes the interface more concise
and allows to reuse automatic to/from Value conversion helpers.

by #185).

### DIFF
--- a/common/cpp/src/google_smart_card_common/requesting/requester_message.h
+++ b/common/cpp/src/google_smart_card_common/requesting/requester_message.h
@@ -67,7 +67,7 @@ struct RequestMessageData {
   // messages (see `ResponseMessageData`) with the requests.
   //
   // Note that this field must be unique among all requests with the same
-  // `name`; in general, requests with different `name` can use overlapping IDs.
+  // `name`. Requests with different `name` can use overlapping IDs.
   RequestId request_id;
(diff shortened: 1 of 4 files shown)


common/cpp/src/google_smart_card_common/requesting/requester_message.h
```diff
@@ -42,29 +41,59 @@ std::string GetRequestMessageType(const std::string& name);
...
+//   type = "say_hello::response"
+//   data = serialized ResponseMessageData structure with the following fields:
+//     request_id = 123
+//     payload = "Hello response from JS"
+struct RequestMessageData {
+  // Unique identifier of the request; is used in order to correlate response
+  // messages (see `ResponseMessageData`) with the requests.
+  //
+  // Note that this field must be unique among all requests with the same
+  // `name`; in general, requests with different `name` can use overlapping IDs.
```
Fabian-Sommer, line 65: nit: Remove "in general" and turn the rest into a full sentence.

Fabian-Sommer (approved):
LGTM % nits

PR #207, opened 2020-10-26
Fix test compilation

emaxx-google:
Fix several issues with tests compilation:
* //common/integration_testing/ and //common/tests-runner/ build
  scripts were pointing to a non-existing .mk file (since #204);
* //common/integration_testing/ C++ code wasn't updated to use Value
  instead of pp::Var (since #196);
* //common/cpp/ build scripts weren't installing value_debug_dumping.h
  into a public header location, which made it impossible to use it in
  tests (originally forgotten in #189);
* //test-all.sh wasn't updated to include the new
  //example_cpp_smart_card_client_app/build/integration_tests/ target.


emaxx-google:
PTAL.

P.S. This is what happens when dealing with many parallel changes without a proper commit waterfall :)

Fabian-Sommer (approved):
LGTM

PR #208, opened 2020-10-26
Add Emscripten barebone build implementation

emaxx-google:
This adds the first version of Emscripten-specific Makefile definitions,
allowing to compile the C/C++ code under WebAssembly. This new
functionality can be enabled by setting the TOOLCHAIN environment
variable to "emscripten" (the variable still defaults to "pnacl" to use
Google Native Client, as before).

Note that this is only a barebone implementation, since it doesn't
support several important features, namely:
* Auto-detecting related .h files and triggering recompilation of .cc
  files on their changes;
* Installation of headers into shared location;
* Building unit tests;
* etc.

These will be addressed in follow-ups, as tracked by #177.

Another note - Right now only very few files in the project can be
compiled using this new Emscripten toolchain, since many files are using
Native Client specific definitions (like the pp::Var class). Their
migration is a separate ongoing effort.

### DIFF
--- a/common/make/internal/executable_building_emscripten.mk
+++ b/common/make/internal/executable_building_emscripten.mk
@@ -15,7 +15,7 @@
 
 #
 # This file contains the implementation of the ../executable_building.mk
-# interface that builds using Emscripten (that generates WebAssembly binaries).
+# interface that builds using Emscripten (which generates WebAssembly binaries).
 #
 # common.mk must be included before including this file.
 #
--- a/common/make/internal/executable_building_emscripten.mk
+++ b/common/make/internal/executable_building_emscripten.mk
@@ -19,6 +19,10 @@
 #
 # common.mk must be included before including this file.
 #
+# Troubleshooting notes:
+# * Set the "EMCC_DEBUG" environment variable to 1 in order to enable more
+#   verbose output from Emscripten's tools.
+#
 
 
 # Sanity-check that the environment is set up.
--- a/common/make/internal/executable_building_emscripten.mk
+++ b/common/make/internal/executable_building_emscripten.mk
@@ -1,0 +1,195 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


common/make/internal/executable_building_emscripten.mk
```diff
@@ -0,0 +1,195 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
```
MahmoudAGawad, line 1: nit: 2020

common/make/internal/executable_building_emscripten.mk
```diff
@@ -0,0 +1,195 @@
...
+
+# Flags passed to the Emscripten compiler/linker tools.
+#
+# Explanation:
+# bind: Enables "Embind" (the technology used by our C++ wrappers for talking
+#   with JavaScript).
+# pthread: Enables Pthreads support (for C/C++ multi-threading, etc.).
+# MODULARIZE: Puts Emscripten module JavaScript loading code into a factory
+#   function, in order to control its loading from other JS code and to avoid
+#   avoid name conflicts with unrelated code.
```
MahmoudAGawad, line 63: nit: remove repeated "avoid"

common/make/internal/executable_building_emscripten.mk
```diff
@@ -0,0 +1,195 @@
...
+ifeq ($(CONFIG),Release)
+
+# Add flags specific to release builds.
+#
+# Explanation:
+# O3: Enable advanced optimizations.
+EMSCRIPTEN_FLAGS += \
+	-O3 \
+
+else ifeq ($(CONFIG),Debug)
```
MahmoudAGawad, line 84: not related to this line but I find that using the `EMCC_DEBUG` env variable can be helpful while in Debug mode: https://emscripten.org/docs/porting/Debugging.html#debug-mode-emcc-debug

MahmoudAGawad (approved):
LGTM % nits

Fabian-Sommer (approved):
LGTM % nit

emaxx-google:
Thanks!

PR #209, opened 2020-10-27
Migrate [js_]requester.h to Value

emaxx-google:
Change requester.h and js_requester.h headers and the corresponding
implementation .cc files to use the toolchain-independent Value
class intead of Native Client specific pp::Var class in most places.
The only left place is the delegate for sending messages to JS, which
will be refactored separately.

by #185).


emaxx-google:
PTAL at the next refactoring CL in the chain... We're about halfway from the goal, I'd say. Thanks in advance.

Fabian-Sommer (approved):
LGTM

PR #210, opened 2020-10-27
Add conversion from std::vector to Value

emaxx-google:
Implement the ConvertToValue() helper for the std::vector<> input type,
with recursively applying the same ConvertToValue() transform to each
item.


Fabian-Sommer (approved):
LGTM

emaxx-google:
thank you.

PR #211, opened 2020-10-27
Add conversion from Value into Emscripten's val

emaxx-google:
Implement helper function that transforms a Value object into
the Emscripten Embind's emscripten::val object.

This will be used in follow-ups in order to send messages to
JavaScript in Emscripten/WebAssembly builds, with the messages
prepared by a toolchain-independent Value class, contributing to
the goal of making most of the //common/cpp library be
toolchain-independent (as tracked by #185).


common/cpp/src/google_smart_card_common/value_emscripten_val_conversion_unittest.cc
```diff
@@ -0,0 +1,205 @@
...
+    ASSERT_TRUE(converted.isString());
+    EXPECT_EQ(converted.as<std::string>(), kFoo);
+  }
+}
+
+TEST(ValueEmscriptenValConversion, BinaryValue) {
+  {
+    const emscripten::val converted =
+        ConvertValueToEmscriptenVal(Value(Value::Type::kBinary));
+    ASSERT_TRUE(converted.instanceof (emscripten::val::global("ArrayBuffer")));
```
emaxx-google, line 126: Hmm, this looks like a clang-format bug. This `instanceof()` is a regular C++ class method, there shouldn't be a space after it. I guess we can keep it as it is. The alternative is to remove the space and surround this code with `// clang-format off`.

Fabian-Sommer (approved):
LGTM

common/cpp/src/google_smart_card_common/value_emscripten_val_conversion_unittest.cc
MahmoudAGawad: I am leaning towards disabling clang-format in this line. It feels like a special operator while in fact it is a class method. Or maybe clang has an rational explanation for this? (I couldn't find any)

MahmoudAGawad (approved):
LGTM.

PR #212, opened 2020-10-27
Add conversion from Value to std::vector

emaxx-google:
Implement the ConvertFromValue() helper for the std::vector<> output
type, with recursively applying the same ConvertFromValue() transform
to each array item.

### DIFF
--- a/common/cpp/src/google_smart_card_common/value_conversion.cc
+++ b/common/cpp/src/google_smart_card_common/value_conversion.cc
@@ -305,7 +305,7 @@ bool ConvertFromValue(Value value, std::vector<uint8_t>* bytes,
                       std::string* error_message) {
   if (value.is_binary()) {
     // This is a special case that is the reason why the standard
-    // array-to-vector template is overloaded by our function.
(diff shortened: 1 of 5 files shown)


common/cpp/src/google_smart_card_common/value_conversion.cc
```diff
@@ -293,10 +295,34 @@ bool ConvertFromValue(Value value, std::string* characters,
...
+      error_message, internal::kErrorWrongTypeValueConversion,
+      Value::kStringTypeTitle, DebugDumpValueSanitized(value).c_str());
   return false;
 }
 
+bool ConvertFromValue(Value value, std::vector<uint8_t>* bytes,
+                      std::string* error_message) {
+  if (value.is_binary()) {
+    // This is a special case that is the reason why the standard
+    // array-to-vector template is overloaded by our function.
```
Fabian-Sommer, line 308: nit: replace "our" with "this"

Fabian-Sommer (approved):
LGTM % nit

emaxx-google:
Thanks!

Issue #132, opened 2020-06-22
Stop installing headers from our libraries into the common place


emaxx-google:
Currently, the build process is organized in a way that each sub-library (like //common/cpp) is built separately and then has its .lib+.h files installed into a common place (in the NaCl environment). Copying the headers is managed via a separate fake "headers_installed.stamp" file (which has a dependency on each of the shared .h file and therefore triggers re-installation once any of them changes).

Originally, that was introduced primarily because the NaCl SDK and naclports were using that model. This model allows, for example:
* using a "stable" version of a library (the one that is already installed into the SDK) while developing a new version of the library;
* deleting the library sources and temporary artifacts after building it once;
* protection against accidental inclusion of the library's internal .h file.

None of those benefits really bring much in the context of sub-libraries within our project, and overall this approach causes a bunch of complexity and fragility in the build scripts. Examples of practical problems:
* parallel builds might fail due to trying to install the headers simultaneously;
* recreating the "env" would result in a stale state with compilation errors that the .h files cannot be found, unless all .stamp files are deleted.

Therefore the suggestion is to get rid of that model, and instead follow a simpler approach that all headers are included directly from the source tree.

PR #213, opened 2020-10-28
Stop installing //common/cpp headers

emaxx-google:
This commit disables installation of "public" headers from the
//common/cpp library into a common include directory. Instead, the
//common/cpp/src directory is simply added into the include search path
in needed makefiles.

More background on this change is in #132 - to put it simply, this is a
pure refactoring that will remove a bunch of compelxity from our build
scripts and additionally make migrating them to Emscripten/WebAssembly
simpler.


Fabian-Sommer (approved):
LGTM

PR #214, opened 2020-10-28
Build //common NaCl-specific files only under NaCl

emaxx-google:
This fixes the makefiles in //common/cpp to exclude source files from
compilation in case the toolchain is non Native Client and the files
can only be compiled under the Native Client.

This fixes the Emscripten/WebAssembly builds of //common/cpp (as tracked
by #177). Note that there are two groups of excluded files currently:
those that provide NaCl-specific capabilities, and those that aren't
migrated to support Emscripten yet. The latter ones will be gradually
migrated and reenabled in follow-ups.


Fabian-Sommer (approved):
LGTM

emaxx-google:
Thank you for the quick review.

PR #215, opened 2020-10-28
Migrate async_request.h and related to Value

emaxx-google:
Change async_request.h, async_requests_storage.h, request_result.h
to use the toolchain-independent Value class intead of Native Client
specific pp::Var class in most places. As an additional bonus, the
request_result.cc, requester.cc, requester_message.cc files became
completely free of NaCl-specific definitions as well.

Besides the mechanical replacement, there's a bit of refactoring due to
using std::move() instead of the previously used copying in related
classes. The background here is that the NaCl's pp::Var class was
copyable (with ref-counting), meanwhile the new Value class is only
moveable.

by #185).

### DIFF
--- a/example_cpp_smart_card_client_app/src/chrome_certificate_provider/api_bridge.cc
+++ b/example_cpp_smart_card_client_app/src/chrome_certificate_provider/api_bridge.cc
@@ -17,6 +17,8 @@
 #include <thread>
 #include <utility>
 
+#include <ppapi/cpp/var.h>
+
 #include <google_smart_card_common/logging/function_call_tracer.h>
 #include <google_smart_card_common/logging/logging.h>
 #include <google_smart_card_common/pp_var_utils/construction.h>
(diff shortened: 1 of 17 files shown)


example_cpp_smart_card_client_app/src/chrome_certificate_provider/api_bridge.cc
```diff
@@ -57,8 +60,9 @@ void ProcessCertificatesRequest(
       locked_certificates_request_handler = certificates_request_handler.lock();
   GOOGLE_SMART_CARD_CHECK(locked_certificates_request_handler);
   if (locked_certificates_request_handler->HandleRequest(&certificates)) {
+    // TODO: Build `Value` directly, without converting from `pp::Var`.
```
Fabian-Sommer, line 63: optional nit: add issue number (here and in other places)

Fabian-Sommer (approved):
LGTM

PR #218, opened 2020-10-29
Add conversion from Emscripten's val into Value

emaxx-google:
Implement helper functions that transform an Emscripten Embind's
emscripten::val object into a Value object.

This will be used in follow-ups in order to receive messages from
JavaScript in Emscripten/WebAssembly builds, with the messages
handling code being toolchain-independent, contributing to the goal
of making most of the //common/cpp library be toolchain-independent
(as tracked by #185).


MahmoudAGawad (approved):
LGTM.

Fabian-Sommer (approved):
LGTM % nit

PR #219, opened 2020-10-29
Add script for building GoogleTest in Emscripten

emaxx-google:
Add a Makefile that performs building of GoogleTest/GoogleMock libraries
and installation of them into the shared static libraries directory.

This Makefile is only supported for Emscripten builds (whose support is
being added at #177), since NaCl SDK already comes with its own port
of GoogleTest.

Note: The Makefile doesn't contain any detection of source file changes
and automated recompilation, since it'd be pretty complex to implement
(because, under the hood, GoogleTest uses a different build system -
CMake) and it's not needed that often (only when someone will uprev the
pinned GoogleTest sources in our repo to a new version). Rebuilding can
be enforced manually by running "make clean".

### DIFF
--- a/third_party/googletest/webport/build/Makefile
+++ b/third_party/googletest/webport/build/Makefile
@@ -1,0 +1,116 @@
+# Copyright 2020 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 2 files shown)


emaxx-google:
PTAL. There's a bunch of copy-paste in the Makefile, but I couldn't find an easy way to reduce it while maintaining readability.

third_party/googletest/webport/build/Makefile
```diff
@@ -0,0 +1,116 @@
+# Copyright 2020 Google Inc. All Rights Reserved.
```
emaxx-google, line 1: An open question is how to name the directories where I put these Emscripten/NaCl specific additions to third-party libraries. Previously I've been using "naclport" in this project. Now, with addition of Emscripten, we need a new name. In this CL I'm using the "webport" word, but I'm open to other suggestions.

MahmoudAGawad: "webport" is good. The first name popped up in my mind is "platform_port".

MahmoudAGawad (approved):
LGTM.

third_party/googletest/webport/build/Makefile
Fabian-Sommer: I like webport.

Fabian-Sommer (approved):
LGTM

PR #221, opened 2020-10-30
Fix errors in the example app built-in PIN dialog

emaxx-google:
1. The dialog not being displayed, due to incorrectly sending a null
   value instead of an empty dictionary value when making a C++-to-JS
   request. (The error introduced in #209.)
2. Fix the "Cancel" button title not being correctly loaded, due to a
   typo in the localized message name (introduced back in #113).


Fabian-Sommer (approved):
LGTM

PR #222, opened 2020-10-30
Convert vector<uint8_t> into binary/ArrayBuffer

emaxx-google:
Change the vector-to-Value and the [obsolete, NaCl-specific]
vector-to-pp::Var conversion implementations to have a special
treatment for the "std::vector<uint8_t>" input type: convert it into a
binary/ArrayBuffer object, instead of the normal conversion into an
array with numbers.

This suits several purposes:
1. Allow seamless transformations in case the ArrayBuffer is really
   required on the receiver side - for example, when calling an
   Extensions API that uses an ArrayBuffer parameter.
2. Avoid introducing a special C++ type that would behave like an
   std::vector<uint8_t>, but denote "convert me into ArrayBuffer,
   unlike any real std::vector". (Note that previously, before we
   started working on WebAssembly, the pp::VarArrayBuffer type
   was suiting that purpose.)
3. Increase the performance (as a single binary object uses O(1)
   allocations, meanwhile an array object uses O(n) ones).

### DIFF
--- a/common/cpp/src/google_smart_card_common/pp_var_utils/construction.h
+++ b/common/cpp/src/google_smart_card_common/pp_var_utils/construction.h
@@ -85,6 +85,9 @@ inline pp::Var MakeVar(const std::vector<T>& value) {
   return result;
 }
 
+// Creates an array buffer.
+pp::Var MakeVar(const std::vector<uint8_t>& value);
+
 // Returns a string in which all characters that cannot be represented in a
 // Pepper value are replaced with a placeholder.
 std::string CleanupStringForVar(const std::string& string);
--- a/common/cpp/src/google_smart_card_common/pp_var_utils/construction.cc
+++ b/common/cpp/src/google_smart_card_common/pp_var_utils/construction.cc
@@ -66,6 +66,10 @@ pp::Var MakeVar(const std::string& value) {
   return value;
 }
 
+pp::Var MakeVar(const std::vector<uint8_t>& value) {
+  return MakeVarArrayBuffer(value);
+}
+
 std::string CleanupStringForVar(const std::string& string) {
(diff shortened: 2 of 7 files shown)


common/cpp/src/google_smart_card_common/pp_var_utils/construction.h
```diff
@@ -85,6 +85,9 @@ inline pp::Var MakeVar(const std::vector<T>& value) {
   return result;
 }
 
+// Creates an array buffer.
```
emaxx-google, line 88: TODO: Add code on the JS side that would convert ArrayBuffer into Uint8Array when sending cross-extension messages. (Chrome doesn't support ArrayBuffer in cross-extension messages, unfortunately.)

Before this is done, the CL shouldn't be committed, since it'll be a regression.

emaxx-google (commented):
Fabian, PTAL.

Note that this removes some tedious conversions on the JS side that we had to add previously into the certificateProvider API bridge.

P.S. I also realized that I actually misguided you in the past by saying that it's not possible to send an ArrayBuffer from C++ to JS - this is really possible, meanwhile what's not possible is to send them from one extension to another. Anyway, hopefully, this CL adds more consistency into our data formats.

Fabian-Sommer (approved):
LGTM

common/cpp/src/google_smart_card_common/pp_var_utils/construction.h
emaxx-google: This is done in #226. After that one lands, will land the current PR as well.

emaxx-google:
Thanks. #226 landed, so it should be safe to proceed with the current PR.

PR #223, opened 2020-10-30
Introduce interface for toolchain-specific C++ ops

emaxx-google:
Introduce a GlobalContext interface that abstracts away some of the
differences between Native Client and Emscripten, such as different ways
of sending messages to JavaScript.

This commit provides implementations of this interface for both Native
Client and Emscripten. Currently both the interface and the
implementations are unused; follow-up commits will migrate the existing
codebase in //common/cpp to use this interface instead of various
NaCl-specific delegates (contributing to the goal of making most of the
//common/cpp library be toolchain-independent, as tracked by #185).

### DIFF
--- a/common/cpp/src/google_smart_card_common/global_context.h
+++ b/common/cpp/src/google_smart_card_common/global_context.h
@@ -35,6 +35,10 @@ class GlobalContext {
   // Returns whether the current thread is the main event loop thread. Is
   // intended to be used to avoid blocking/deadlocking the main thread.
   virtual bool IsMainEventLoopThread() const = 0;
+
+  // Disables communication with the JavaScript side. All calls to
+  // `PostMessageToJs()` after this point will return `false`.
+  virtual void DisableJsCommunication() = 0;
 };
 
 }  // namespace google_smart_card
--- a/common/cpp/src/google_smart_card_common/global_context.h
+++ b/common/cpp/src/google_smart_card_common/global_context.h
@@ -1,0 +1,42 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GOOGLE_SMART_CARD_COMMON_GLOBAL_CONTEXT_H_
+#define GOOGLE_SMART_CARD_COMMON_GLOBAL_CONTEXT_H_
+
+#include <google_smart_card_common/value.h>
+
+namespace google_smart_card {
+
+// Global context is an interface that abstracts away webport-specific
+// operations.
+//
+// The implementation of this class is required to be thread-safe.
+class GlobalContext {
+ public:
+  virtual ~GlobalContext() = default;
+
+  // Sends the given message to the JavaScript side. Returns whether the message
+  // was sent successfully (note that the status doesn't tell anything about the
+  // result of the message handling on the JS side).
+  virtual bool PostMessageToJs(const Value& message) = 0;
+
+  // Returns whether the current thread is the main event loop thread. Is
+  // intended to be used to avoid blocking/deadlocking the main thread.
+  virtual bool IsMainEventLoopThread() const = 0;
+};
+
+}  // namespace google_smart_card
+
+#endif  // GOOGLE_SMART_CARD_COMMON_GLOBAL_CONTEXT_H_
--- a/common/cpp/src/google_smart_card_common/global_context_impl_emscripten.cc
+++ b/common/cpp/src/google_smart_card_common/global_context_impl_emscripten.cc
@@ -1,0 +1,50 @@
+// Copyright 2020 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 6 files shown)


common/cpp/src/google_smart_card_common/global_context.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <google_smart_card_common/value.h>
+
+namespace google_smart_card {
+
+// Global context is an interface that abstracts away webport-specific
+// operations.
+//
+// The implementation of this class is required to be thread-safe.
+class GlobalContext {
```
emaxx-google, line 26: WDYT? It's not the most elegant interface, admittedly, but it's somewhat dictated by the current code structure. For example, it's a drop-in replacement for the `JsRequester::Delegate` delegate: https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/blob/6257dbf03a65b721b007d5bd1e044ca23163fd75/common/cpp/src/google_smart_card_common/requesting/js_requester.h#L51

As an alternative, for example, I'd rather see this as an `MessagePipe` interface with methods like `PostMessages()` and `RegisterTypedMessageHandler()`. But doing that now would increase the already big amount of refactoring to be done.

common/cpp/src/google_smart_card_common/global_context_impl_emscripten.cc
```diff
@@ -0,0 +1,50 @@
...
+
+void GlobalContextImplEmscripten::DetachFromPostMessage() {
+  const std::unique_lock<std::mutex> lock(mutex_);
+  post_message_callback_ = emscripten::val::undefined();
+}
+
+bool GlobalContextImplEmscripten::PostMessageToJs(const Value& message) {
+  const std::unique_lock<std::mutex> lock(mutex_);
+  if (post_message_callback_.isUndefined()) return false;
+  post_message_callback_(ConvertValueToEmscriptenVal(message));
```
emaxx-google, line 42: self-nit: The conversion should be performed before entering the mutex, in order to minimize the time spent inside the lock.

common/cpp/src/google_smart_card_common/global_context.h
Fabian-Sommer: I think this is fine for now. We can tackle any other refactoring after successfully migrating to Wasm.

common/cpp/src/google_smart_card_common/global_context_impl_nacl.h
```diff
@@ -0,0 +1,57 @@
...
+  // this point will return `false`.
+  void DetachFromPpInstance();
+
+  // GlobalContext:
+  bool PostMessageToJs(const Value& message) override;
+  bool IsMainEventLoopThread() const override;
+
+ private:
+  pp::Core* const pp_core_;
+  std::mutex mutex_;
```
Fabian-Sommer, line 51: nit: Put mutex_ below pp_instance_ or above pp_core_

common/cpp/src/google_smart_card_common/global_context_impl_nacl.cc
```diff
@@ -0,0 +1,49 @@
...
+
+void GlobalContextImplNacl::DetachFromPpInstance() {
+  const std::unique_lock<std::mutex> lock(mutex_);
+  pp_instance_ = nullptr;
+}
+
+bool GlobalContextImplNacl::PostMessageToJs(const Value& message) {
+  const std::unique_lock<std::mutex> lock(mutex_);
+  if (!pp_instance_) return false;
+  pp_instance_->PostMessage(ConvertValueToPpVar(message));
```
Fabian-Sommer, line 41: nit: Convert message before locking

Fabian-Sommer (approved):
LGTM % nit

common/cpp/src/google_smart_card_common/global_context_impl_nacl.h
emaxx-google: Ack - It's actually intentional, because the mutex protects the members that come after it. There's no hard rule, but I believe a common convention to order the members like this.

emaxx-google (commented):
Thanks!

common/cpp/src/google_smart_card_common/global_context_impl_emscripten.h
```diff
@@ -0,0 +1,63 @@
...
+  GlobalContextImplEmscripten(std::thread::id main_thread_id,
+                              emscripten::val post_message_callback);
+  GlobalContextImplEmscripten(const GlobalContextImplEmscripten&) = delete;
+  GlobalContextImplEmscripten& operator=(const GlobalContextImplEmscripten&) =
+      delete;
+  ~GlobalContextImplEmscripten() override;
+
+  // Disables posting new messages to JS. All calls to `PostMessageToJs()` after
+  // this point will return `false`.
+  void DetachFromPostMessage();
```
emaxx-google, line 47: BTW, maybe I should call it and the similar method in the other class simply `ShutDown()`. Happy to rename if you agree it's better.

MahmoudAGawad: IMHO, maybe not `ShutDown()` as it indicates the pointer (or generally the service) will not be valid for subsequent method calls, like `PostMessageToJs()` in this case.

I have a question, though: why don't you define this common method in the interface `GlobalContext`? I see they share the same intention but differ in implementation. For the name, it can be `DetachFromPostMessage()` or `DisablePostMessageToJs()`.

MahmoudAGawad (approved):
LGTM.

common/cpp/src/google_smart_card_common/global_context_impl_emscripten.h
emaxx-google: Yup, good observation regarding making this method virtual. I was originally thinking that these methods in two classes are conceptually different (one is when the NaCl framework tells us to destroy the `Instance`, and another will be used when the JS code destroys the proxy object), but, looking at them now, they're essentially about the same operation.

What about a name like `DisableJsCommunication()`? This will be a bit more future-proof, in case we add more stuff that needs to be "shut down" together with this JS-to-C++ pipe.

MahmoudAGawad: yes, I like the name.

emaxx-google: Done.

emaxx-google:
I assume "still LGTM", therefore landing this CL.

PR #224, opened 2020-10-30
Use "-pthread" for GoogleTest builds in Emscripten

emaxx-google:
Add the "-pthread" flag when compiling and linking the GoogleTest static
libraries in Emscripten builds.

Meanwhile Pthreads aren't actually used by GoogleTest itself (due to the
"gtest_disable_pthreads" parameter that we specify), it's still necessary
because we use "-pthread" when compiling other libraries and source
files, and Emscripten seems to require this flag to be used consistently
everywhere.

This is a preparatory fix for introducing Emscripten builds of unit
tests, as tracked by #177.

### DIFF
--- a/third_party/googletest/webport/build/Makefile
+++ b/third_party/googletest/webport/build/Makefile
@@ -78,7 +78,13 @@ endif
 #   the compilation according to prepared scripts.
 #
 # Explanation of parameters to cmake:
-# B: build directory.
+# E env: Wraps the further second cmake call into the "env" tool that allows to
+#   specify environment variables.
+# CXXFLAGS: Enables Emscripten Pthreads support (despite that it's not used by
+#   GoogleTest itself due to the other parameter below, Emscripten requires this
+#   flag to be set if a library is linked later with other files that have it).
+# LDFLAGS: Enables Emscripten Pthreads support.
+# B: Build directory.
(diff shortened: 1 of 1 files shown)


third_party/googletest/webport/build/Makefile
```diff
@@ -78,14 +78,24 @@ endif
 #   the compilation according to prepared scripts.
 #
 # Explanation of parameters to cmake:
-# B: build directory.
+# E env: Wraps the further second cmake call into the "env" tool that allows to
```
Fabian-Sommer, line 81: nit: Isn't it the first cmake call, rather than the second? Also, delete "further".

Fabian-Sommer (approved):
LGTM % nit

third_party/googletest/webport/build/Makefile
emaxx-google: The first `cmake E env ...` call is used to set up the environment values for the second `cmake ../../src` one. If the comment doesn't do a good job in explaining that, I'm happy to rephrase it.

Fabian-Sommer: I see. I think I would go with `Wraps the succeeding cmake call`

MahmoudAGawad (approved):
LGTM.

emaxx-google:
Thanks!

PR #225, opened 2020-10-30
Temporarily disable tests that fail in Emscripten

emaxx-google:
Disable the following tests that are currently failing in Emscripten
builds:
* RequestingAsyncRequestsStorageTest.MultiThreading;
* MessagingTypedMessageRouterTest.MultiThreading.

Both tests are related to threading; they're failing due to
out-of-memory (due to hitting the default memory limit when allocating
memory for each thread). Naively increasing the limit doesn't help - the
tests begin to crash with more mysterious errors. Therefore fixing these
tests will require separate later investigation (tracked by #185).


Fabian-Sommer (approved):
LGTM

emaxx-google:
Thanks!

PR #226, opened 2020-10-30
Substitute ArrayBuffer in cross-extension messages

emaxx-google:
Replace ArrayBuffer's with byte Array's when sending cross-extension
messages.

The background for this change is that Chrome Extension APIs don't
support sending ArrayBuffer's inside messages from one extension/app to
another. Currently the code in this project never attempts to do that,
but the refactoring proposed in #222 will start putting ArrayBuffer's
into the sent messages. Therefore the current commit anticipates this
upcoming change and adds an automatic conversion that will prevent data
loss in cross-extension messages.

### DIFF
--- a/common/js/src/messaging/port-message-channel.js
+++ b/common/js/src/messaging/port-message-channel.js
@@ -23,6 +23,7 @@
 
 goog.provide('GoogleSmartCard.PortMessageChannel');
 
+goog.require('GoogleSmartCard.ContainerHelpers');
 goog.require('GoogleSmartCard.DebugDump');
 goog.require('GoogleSmartCard.Logging');
 goog.require('GoogleSmartCard.MessageChannelPinging.PingResponder');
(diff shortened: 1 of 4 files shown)


common/js/src/messaging/port-message-channel.js
```diff
@@ -106,7 +107,10 @@ PortMessageChannel.prototype.send = function(serviceName, payload) {
   GSC.Logging.checkWithLogger(this.logger, goog.isObject(payload));
   goog.asserts.assertObject(payload);
 
-  var typedMessage = new GSC.TypedMessage(serviceName, payload);
+  const normalizedPayload =
+      GSC.ContainerHelpers.substituteArrayBuffersRecursively(payload);
+
+  var typedMessage = new GSC.TypedMessage(serviceName, normalizedPayload);
   var message = typedMessage.makeMessage();
```
Fabian-Sommer, line 114: nit: `let` instead of `var`

Fabian-Sommer (approved):
LGTM % nit

PR #227, opened 2020-10-31
Register enums/structs in to/from Value conversion

emaxx-google:
Add helper definitions that allow conversions to/from Value objects for
every relevant enum/struct in this project (i.e., for every enum/struct
that currently support conversions to/from Native Client pp::Var objects).

This paves the way for continuing the //common/cpp WebAssembly migration
effort (as tracked by #185) without breaking the compilation of the
whole project. Specifically, some upcoming commits will change the way
how outgoing request parameters are serialized before being sent to the
JavaScript side: they'll be converted into Value instead of pp::Var.

Note that the current commit still keeps the old pp::Var conversion
helpers: those will be cleaned up in later follow-ups once they become
unused.

The commit also introduces behavior change: some of the struct fields
are switched to use std::vector<uint8_t>, which, starting from #222,
will make them be serialized into JavaScript ArrayBuffer instead of
bytes Array. It's expected to be a non-breaking change, thanks to #226.

### DIFF
--- a/common/cpp/src/google_smart_card_common/external_logs_printer.cc
+++ b/common/cpp/src/google_smart_card_common/external_logs_printer.cc
@@ -23,6 +23,7 @@
 #include <google_smart_card_common/messaging/typed_message_listener.h>
 #include <google_smart_card_common/pp_var_utils/struct_converter.h>
 #include <google_smart_card_common/value.h>
+#include <google_smart_card_common/value_conversion.h>
 #include <google_smart_card_common/value_nacl_pp_var_conversion.h>
 
 namespace google_smart_card {
@@ -43,6 +44,14 @@ void PrintExternalLogMessage(const ExternalLogMessageData& message_data) {
 }
 
 }  // namespace
+
+template <>
+StructValueDescriptor<ExternalLogMessageData>::Description
+StructValueDescriptor<ExternalLogMessageData>::GetDescription() {
+  return Describe("ExternalLogMessageData")
+      .WithField(&ExternalLogMessageData::formatted_log_message,
+                 "formatted_log_message");
+}
 
 template <>
 constexpr const char*
--- a/common/cpp/src/google_smart_card_common/external_logs_printer.cc
+++ b/common/cpp/src/google_smart_card_common/external_logs_printer.cc
@@ -48,6 +48,8 @@ void PrintExternalLogMessage(const ExternalLogMessageData& message_data) {
 template <>
 StructValueDescriptor<ExternalLogMessageData>::Description
 StructValueDescriptor<ExternalLogMessageData>::GetDescription() {
+  // Note: Strings passed to WithField() below must match the property names in
+  // //common/js/src/logging/log-buffer-forwarder.js.
   return Describe("ExternalLogMessageData")
       .WithField(&ExternalLogMessageData::formatted_log_message,
                  "formatted_log_message");
--- a/third_party/libusb/naclport/src/chrome_usb/types.cc
+++ b/third_party/libusb/naclport/src/chrome_usb/types.cc
@@ -104,6 +104,8 @@ bool ControlTransferInfo::operator==(const ControlTransferInfo& other) const {
 template <>
 EnumValueDescriptor<Direction>::Description
 EnumValueDescriptor<Direction>::GetDescription() {
+  // Note: Strings passed to WithItem() below must match the enum names in the
+  // chrome.usb API.
   return Describe("chrome_usb::Direction")
       .WithItem(Direction::kIn, "in")
       .WithItem(Direction::kOut, "out");
@@ -126,6 +128,8 @@ void DirectionConverter::VisitCorrespondingPairs(Callback callback) {
 template <>
 StructValueDescriptor<Device>::Description
 StructValueDescriptor<Device>::GetDescription() {
+  // Note: Strings passed to WithField() below must match the property names in
+  // the chrome.usb API.
   return Describe("chrome_usb::Device")
       .WithField(&Device::device, "device")
       .WithField(&Device::vendor_id, "vendorId")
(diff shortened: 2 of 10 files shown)


common/cpp/src/google_smart_card_common/external_logs_printer.cc
```diff
@@ -44,6 +45,14 @@ void PrintExternalLogMessage(const ExternalLogMessageData& message_data) {
 
 }  // namespace
 
+template <>
+StructValueDescriptor<ExternalLogMessageData>::Description
+StructValueDescriptor<ExternalLogMessageData>::GetDescription() {
+  return Describe("ExternalLogMessageData")
+      .WithField(&ExternalLogMessageData::formatted_log_message,
```
emaxx-google, line 52: Question: Maybe I should introduce a short alias like `T` that would allow to shorten the descriptions, to something like:

```
  .WithField(&T::formatted_log_message, "formatted_log_message");
```

... or does the clarity of the current approach outweigh its verbosity?

third_party/libusb/naclport/src/libusb_over_chrome_usb.cc
```diff
@@ -92,16 +92,15 @@ constexpr int kLibusbRequestTypeMask =
     LIBUSB_REQUEST_TYPE_STANDARD | LIBUSB_REQUEST_TYPE_CLASS |
     LIBUSB_REQUEST_TYPE_VENDOR | LIBUSB_REQUEST_TYPE_RESERVED;
 
-std::unique_ptr<uint8_t> CopyRawData(const uint8_t* data, size_t byte_count) {
-  std::unique_ptr<uint8_t> result(new uint8_t[byte_count]);
+std::unique_ptr<uint8_t[]> CopyRawData(const uint8_t* data, size_t byte_count) {
```
emaxx-google, line 95: (Fixed it while we're here. Technically, the old code was correct as well, since we never destroy those `std::unique_ptr`, but always do `.release() + delete[]`. But it was probably confusing and not future-proof to omit these brackets here.)

third_party/libusb/naclport/src/chrome_usb/types.cc
```diff
@@ -212,6 +263,20 @@ void EndpointDescriptorUsageConverter::VisitCorrespondingPairs(
...
+StructValueDescriptor<EndpointDescriptor>::GetDescription() {
+  return Describe("chrome_usb::EndpointDescriptor")
+      .WithField(&EndpointDescriptor::address, "address")
+      .WithField(&EndpointDescriptor::type, "type")
+      .WithField(&EndpointDescriptor::direction, "direction")
+      .WithField(&EndpointDescriptor::maximum_packet_size, "maximumPacketSize")
+      .WithField(&EndpointDescriptor::synchronization, "synchronization")
+      .WithField(&EndpointDescriptor::usage, "usage")
+      .WithField(&EndpointDescriptor::polling_interval, "pollingInterval")
+      .WithField(&EndpointDescriptor::extra_data, "extra_data");
```
Fabian-Sommer, line 277: nit: `extraData` (I see the same is done in some VisitFields methods, it may be good to change it there too)

third_party/libusb/naclport/src/chrome_usb/types.cc
```diff
@@ -233,6 +298,20 @@ void EndpointDescriptorConverter::VisitFields(const EndpointDescriptor& value,
...
+StructValueDescriptor<InterfaceDescriptor>::GetDescription() {
+  return Describe("chrome_usb::InterfaceDescriptor")
+      .WithField(&InterfaceDescriptor::interface_number, "interfaceNumber")
+      .WithField(&InterfaceDescriptor::alternate_setting, "alternateSetting")
+      .WithField(&InterfaceDescriptor::interface_class, "interfaceClass")
+      .WithField(&InterfaceDescriptor::interface_subclass, "interfaceSubclass")
+      .WithField(&InterfaceDescriptor::interface_protocol, "interfaceProtocol")
+      .WithField(&InterfaceDescriptor::description, "description")
+      .WithField(&InterfaceDescriptor::endpoints, "endpoints")
+      .WithField(&InterfaceDescriptor::extra_data, "extra_data");
```
Fabian-Sommer, line 312: ditto

third_party/libusb/naclport/src/chrome_usb/types.cc
```diff
@@ -254,6 +333,20 @@ void InterfaceDescriptorConverter::VisitFields(const InterfaceDescriptor& value,
...
+StructValueDescriptor<ConfigDescriptor>::GetDescription() {
+  return Describe("chrome_usb::ConfigDescriptor")
+      .WithField(&ConfigDescriptor::active, "active")
+      .WithField(&ConfigDescriptor::configuration_value, "configurationValue")
+      .WithField(&ConfigDescriptor::description, "description")
+      .WithField(&ConfigDescriptor::self_powered, "selfPowered")
+      .WithField(&ConfigDescriptor::remote_wakeup, "remoteWakeup")
+      .WithField(&ConfigDescriptor::max_power, "maxPower")
+      .WithField(&ConfigDescriptor::interfaces, "interfaces")
+      .WithField(&ConfigDescriptor::extra_data, "extra_data");
```
Fabian-Sommer, line 347: ditto

Fabian-Sommer (approved):
LGTM % nits

third_party/libusb/naclport/src/chrome_usb/types.cc
emaxx-google: Ack - This naming is coming from the chrome.usb API: https://developer.chrome.com/apps/usb#type-EndpointDescriptor
We have to use the same here, otherwise we'll not be able to convert this object coming from the Javascript-based Extensions API into a C++ struct (and vice versa - the API will raise an error if we call a method with a struct with a wrongly named field).

The .h file, which admittedly isn't visible in the PR, mentions that the structs are mimicking the Extensions API.

third_party/libusb/naclport/src/chrome_usb/types.cc
emaxx-google: Same as above - it's dictated by the naming in Chrome Extensions API.

third_party/libusb/naclport/src/chrome_usb/types.cc
emaxx-google: Same as above - it's dictated by the naming in Chrome Extensions API.

emaxx-google (commented):
Thanks, Fabian.

third_party/libusb/naclport/src/chrome_usb/types.cc
emaxx-google: As per offline discussion, I've added a comment into every relevant `GetDescription()` function that explains where these strings are coming from.

PR #228, opened 2020-11-02
Migrate request_handler.h to Value

emaxx-google:
Change the request_handler.h header and all related code to use the
pp::Var class.

by #185).


Fabian-Sommer (approved):
LGTM

PR #230, opened 2020-11-03
Fix Emscripten link rule wrong arguments position

emaxx-google:
Fix the LINK_EXECUTABLE_RULE makefile rule in Emscripten builds to have
the correct list of arguments: the rule was incorrectly written as if
there was an additional first argument that specifies the name of the
executable to be linked. In reality, as documented in the "interface"
file and implemented in the NaCl build scripts, there's no such
argument, and the TARGET variable should be used for that purpose.

This fixes the mistake introduced in the original PR #208. The umbrella
issue that tracks the Emscripten build implementation is #177.


Fabian-Sommer (approved):
LGTM

emaxx-google:
Thanks for the quick review!

PR #231, opened 2020-11-03
Disable one more failing test in Emscripten

emaxx-google:
Disable the following test that is currently failing in Emscripten
builds:
* RequestingAsyncRequestTest.AsyncRequestStateMultiThreading.

Same as with the tests disabled in #225, the issue is related to the
multi-threading functionality and will need to be investigated
separately as part of the effort tracked in #185.


Fabian-Sommer (approved):
LGTM

PR #232, opened 2020-11-03
Implement C++ unit test runner for Emscripten

emaxx-google:
This commit adds the support of building and running C++ unit tests in
Emscripten builds.

In order to achieve this, there's also a refactoring made that
simplifies how the tests' makefiles are organized and allows to abstract
away differences between Native Client and Emscripten.

This is an improvement for the Emscripten/WebAssembly build
infrastructure that is tracked by #177.

### DIFF
--- a/common/cpp/build/tests/Makefile
+++ b/common/cpp/build/tests/Makefile
@@ -12,13 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-#
 # Makefile for the C++ library tests.
 #
-# The tests are executed through the separate tests runner, and therefore its
-# makefile has to be used - with our specific options plugged into it through a
-# bunch of make variables.
-#
+# The makefile is based on the helper definitions provided by
+# "cpp_unit_test_runner".
 
 include ../../../cpp_unit_test_runner/include.mk
 
--- a/common/cpp/build/tests/Makefile
+++ b/common/cpp/build/tests/Makefile
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 #
 # Makefile for the C++ library tests.
 #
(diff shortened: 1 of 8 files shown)


common/cpp_unit_test_runner/src/build_emscripten.mk
```diff
@@ -0,0 +1,74 @@
...
+	$(LIB_DIR)/libgtest.a \
+	$(LIB_DIR)/libgtest_main.a \
+
+$(GOOGLETEST_LIBS) &:
+	$(MAKE) -C $(ROOT_PATH)/third_party/googletest/webport/build
+
+# Documented at ../include.mk.
+#
+# Implementation notes:
+# The execution is performed via Node.js.
```
emaxx-google, line 63: @MahmoudAGawad PTAL at this part as well, in case you have any experience/suggestions here. Thanks.

common/cpp_unit_test_runner/src/build_emscripten.mk
```diff
@@ -0,0 +1,74 @@
...
+# Flags passed to the Emscripten compiler/linker tools in test builds.
+#
+# Explanation:
+# MODULARIZE: Disable putting the Emscripten module JavaScript loading code into
+#   a factory function, so that the test runner module is loaded automatically
+#   on startup.
+TEST_ADDITIONAL_EMSCRIPTEN_FLAGS := \
+	-s MODULARIZE=0 \
+
+# Documented at ../include.mk.
```
Fabian-Sommer, line 27: nit: Documented in ../include.mk (same in other places in this CL)

common/cpp_unit_test_runner/src/build_emscripten.mk
```diff
@@ -0,0 +1,74 @@
...
+# builds the C++ unit test runner using the Emscripten toolchain.
+
+# Flags passed to the Emscripten compiler/linker tools in test builds.
+#
+# Explanation:
+# MODULARIZE: Disable putting the Emscripten module JavaScript loading code into
+#   a factory function, so that the test runner module is loaded automatically
+#   on startup.
+TEST_ADDITIONAL_EMSCRIPTEN_FLAGS := \
+	-s MODULARIZE=0 \
```
Fabian-Sommer, line 25: Do we expect that all emscripten specific flags are applied to both compiling and linking? If not, it may be warranted to split this variable into two, or just use it directly in the two instances where it is used (though this solution does provide a nice place for the comment about MODULARIZE).

common/cpp/build/tests/Makefile
```diff
@@ -21,46 +20,46 @@
...
 
-ADDITIONAL_TEST_CPPFLAGS := \
-	-I$(TEST_SOURCES_PATH) \
+CXXFLAGS := \
+	-I$(ROOT_SOURCES_PATH)/ \
+	-pedantic \
+	-Wall \
+	-Werror \
+	-Wextra \
+	-std=gnu++11 \
```
Fabian-Sommer, line 56: I would prefer `-std=c++11` (if that does not work, perhaps document the reason for it - not in this file, but in a more central place). Either thing should probably be a separate CL.

Fabian-Sommer (approved):
LGTM % nits

common/cpp/build/tests/Makefile
emaxx-google: In NaCl there were some problems with `c++11`, which is why is had to use `gnu++11` (don't remember specific examples right now).
But you're right that it's not documented anywhere in this project, and also it'll probably be better to achieve "`gnu+11` in NaCl, and `c++11` in WebAssembly". I'll look into this later.

common/cpp_unit_test_runner/src/build_emscripten.mk
emaxx-google: I'm not very experienced in Emscripten, but I was following their guideline that recommends using the same arguments: https://github.com/emscripten-core/emscripten/blob/9d37058cedb5ffa482202eb50812e67cd9bec810/src/settings.js#L30
I already experienced it with things like `-pthread` which result in weird compilation issues when it's missing somewhere. I can imagine some other flags can even cause runtime issues if used inconsistently.

Having said that, we definitely can move a flag into a specific location (like CXXFLAGS or LDFLAGS), as long as we 100% sure that it's needed only on a specific stage. But I haven't found any reference source for that yet.

common/cpp_unit_test_runner/src/build_emscripten.mk
MahmoudAGawad: Unfortunately, I don't have experience with this :(

MahmoudAGawad (approved):
I defer to Fabian.

emaxx-google:
Alright, thanks!

PR #234, opened 2020-11-04
Migrate remote_call_message.{h|cc} to Value

emaxx-google:
Change the remote_call_message.h and .cc files to use the
pp::Var class. Brief overview of changes in this commit:

* Introduce a dedicated RemoteCallRequestPayload struct for holding all
  fields of the message's payload - this allows to reuse the automatic
  to/from Value conversion mechanism and reduces the boilerplate when
  passing this data around.
* Implement the "any-to-RemoteCallRequestPayload" conversion. (The old
  NaCl-specific analog was MakeVarArray() from
  pp_var_utils/construction.h.)
  This conversion is covered with its own unit tests.
* Mechanical replacements to support the new interface.

by #185).


Fabian-Sommer (approved):
LGTM (but may need to be re-formatted)

emaxx-google:
Thank you.

PR #235, opened 2020-11-04
Switch C++ sources format to Chromium code style

emaxx-google:
Change the format-code.sh script to tell clang-format to use the
"Chromium" style instead of the "Google" one. Also reformat all of the
existing code according to this new style (except the third-party code).

Given that the main product of this repository is the Smart Card
Connector app which is one of Chrome OS components (even if optional and
separately distributed), and given that all of this repository's
maintainers are also Chromium developers, it makes sense to switch to
the "Chromium" style guide.

Subjectively speaking, a big benefit of the "Chromium" style is that it
doesn't collapse short "if" statements into a single line - a two-line
format is, arguably, easier to read.

This commit is a follow-up to #202 and #187 which introduced the
strict "Google" style following in this repository.

### DIFF
--- a/common/cpp/src/google_smart_card_common/external_logs_printer.cc
+++ b/common/cpp/src/google_smart_card_common/external_logs_printer.cc
@@ -64,7 +64,8 @@ StructConverter<ExternalLogMessageData>::GetStructTypeName() {
 template <>
 template <typename Callback>
 void StructConverter<ExternalLogMessageData>::VisitFields(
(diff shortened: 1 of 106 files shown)


emaxx-google:
PTAL - I actually like the result. Thanks for pointing this out in the offline conversation!
The PR is huge, but only the `format-code.sh` file was edited manually, and the rest is the output of clang-format as-is.

Fabian-Sommer (approved):
LGTM

PR #236, opened 2020-11-04
Compile all tests in make-all.sh

emaxx-google:
Extend the //make-all.sh script to additionally compile all unit and
integration test targets.

This makes it simpler to perform a routine compilation check in the
middle of some refactoring; the existing //test-all.sh script allowed
to achieve the same, but it was also launching all tests which
significantly increased the required time.

Additionally, this change makes sure that the
//common/integration_testing helper library is built before the targets
that use it are built, which avoids some flaky compilation errors (due
to our imperfect handling of concurrent builds of recursive makefiles).


Fabian-Sommer (approved):
LGTM

PR #237, opened 2020-11-04
Migrate js_requester.{h|cc} to GlobalContext

emaxx-google:
This commit changes the JsRequester class to use the
toolchain-independent GlobalContext interface instead of directly
operating Native Client specific objects (pp::Core, pp::Instance).

This makes the whole JsRequester class work under both
Emscripten/WebAssembly and Native Client (contributing to the
effort tracked by #185). Besides the mechanical replacement, the
commit had to introduce construction of the GlobalContext object
in entry point .cc files and passing them through down to the
JsRequester constructors.

* google_smart_card_libusb/global.{h|cc};
* google_smart_card_pcsc_lite_client/global.{h|cc}.


Fabian-Sommer (approved):
LGTM

emaxx-google:
Thank you.

PR #238, opened 2020-11-05
Migrate remote_call_adaptor.{h|cc} to Value

emaxx-google:
This commit changes the RemoteCallAdaptor class to use the
toolchain-independent Value class instead of Native Client specific
classes (like pp::Var). This contributes to the effort tracked
by #185.

Besides mechanical replacements, this change introduces a helper class
for converting the input/output remote call arguments from Values into
any supported C++ type. Unlike old NaCl-based conversion helpers,
this class is more flexible (e.g., in case the argument list doesn't have a
fixed length) and produces better error messages.

* chrome_usb/api_bridge.cc;
* pcsc_lite_over_requester.cc.

### DIFF
--- a/common/cpp/build/Makefile
+++ b/common/cpp/build/Makefile
@@ -37,6 +37,7 @@ SOURCES := \
 	$(SOURCES_PATH)/messaging/typed_message_router.cc \
 	$(SOURCES_PATH)/multi_string.cc \
 	$(SOURCES_PATH)/numeric_conversions.cc \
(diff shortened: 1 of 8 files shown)


emaxx-google:
PTAL. This is probably the last "scary variadic templates" CL :-)

emaxx-google:
@Fabian-Sommer - friendly ping.

Fabian-Sommer (approved):
LGTM

emaxx-google:
Thanks!

PR #239, opened 2020-11-05
Migrate js_request_receiver.{h|cc} to GlobalContext

emaxx-google:
Change the JsRequestReceiver class to use the toolchain-independent
GlobalContext interface instead of directly operating Native Client
specific objects (pp::Core, pp::Instance).

This makes the whole JsRequestReceiver class work under both
Emscripten/WebAssembly and Native Client (contributing to the
effort tracked by #185).

* google_smart_card_pcsc_lite_server_clients_management/backend.{h|cc}.

### DIFF
--- a/third_party/pcsc-lite/naclport/server_clients_management/src/clients_manager.cc
+++ b/third_party/pcsc-lite/naclport/server_clients_management/src/clients_manager.cc
@@ -34,6 +34,7 @@
 #include <ppapi/cpp/var.h>
 
 #include <google_smart_card_common/formatting.h>
+#include <google_smart_card_common/global_context.h>
 #include <google_smart_card_common/logging/logging.h>
 #include <google_smart_card_common/pp_var_utils/struct_converter.h>
 #include <google_smart_card_common/requesting/remote_call_message.h>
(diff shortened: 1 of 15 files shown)


third_party/pcsc-lite/naclport/server_clients_management/src/clients_manager.cc
```diff
@@ -136,7 +137,6 @@ void PcscLiteServerClientsManager::Detach() {
   typed_message_router_->RemoveRoute(&create_handler_message_listener_);
   typed_message_router_->RemoveRoute(&delete_handler_message_listener_);
   DeleteAllHandlers();
-  pp_instance_ = nullptr;
```
Fabian-Sommer, line 139: nit: `global_context_ = nullptr`

Fabian-Sommer (approved):
LGTM % nit

third_party/pcsc-lite/naclport/server_clients_management/src/clients_manager.cc
emaxx-google: Ack - It's intentional that we're keeping `global_context_`, since this object is intended to be available even after "detaching". The idea is that the consumer can still call methods of this class, which simplifies the consumer's implementation, it's just that the `PostMessageToJs()` will return an error if used after detaching.

emaxx-google:
Thanks.

Issue #220, opened 2020-10-29
Add WebAssembly support into //example_cpp_smart_card_client_app


emaxx-google:
The Example C++ Client App (//example_cpp_smart_card_client_app) and its dependencies (namely, //third_party/pcsc-lite/naclport/cpp_client) need to be refactored and extended in order to support working under WebAssembly. Currently it only compiles under Native Client.

Depends on: #185 (migration of the //common/cpp library to WebAssembly).

PR #240, opened 2020-11-05
Migrate ui_bridge.{h|cc} to GlobalContext

emaxx-google:
Change the UiBridge class to use the toolchain-independent GlobalContext
interface instead of directly operating Native Client specific objects
(pp::Core, pp::Instance).

This contributes to the goal of making the UiBridge implementation work
under both Emscripten/WebAssembly and Native Client (as part of the
bigger effort tracked by #220).


Fabian-Sommer (approved):
LGTM

emaxx-google:
Thanks. I've changed the PR description to note that we're only on the way to make UiBridge be toolchain-independent - there still are some "pp::Var" usages in it, which will be tackled in a separate PR.

PR #242, opened 2020-11-06
Migrate ui_bridge.{h|cc} to Value

emaxx-google:
Change the UiBridge class to use the toolchain-independent Value class,
instead of the Native Client specific pp::Var class.

This completes the migration of the UiBridge class in order to support
both Native Client and Emscripten/WebAssembly builds (as part of the
effort tracked by #220).


Fabian-Sommer (approved):
LGTM

PR #243, opened 2020-11-06
Migrate [js_]request_receiver.{h|cc} to Value

emaxx-google:
Change the RequestReceiver and JsRequestReceiver classes to use the
pp::Var class.

This makes these classes compile and work both under Native Client
and Emscripten/WebAssembly, contributing to the effort of migration of
the //common/cpp library (as tracked by #185).


Fabian-Sommer (approved):
LGTM

emaxx-google:
Thank you for the quick review.

PR #244, opened 2020-11-06
Convert //common/integration_testing to Value

emaxx-google:
Change the implementation of test helpers in
//common/integration_testing to use the platform-independent Value
class, instead of Native Client specific pp::Var class. This contributes
to the goal of making the common C++ libraries in this project work
under Emscripten/WebAssembly (as tracked by #185).

Other changes made in this commit:
* Helper methods added for a single-line conversion of incoming remote
  call arguments from Values into C++ objects;
* api_bridge_integration_test_helper.cc became, as a bonus side effect,
  toolchain-independent as well.


Fabian-Sommer (approved):
LGTM

PR #246, opened 2020-11-07
Fix goog.async.Deferred not available in JS code

emaxx-google:
This commit adds the
//third_party/closure-library/src/third_party/closure directory into the
list of source directories for JavaScript builds performed by Closure
Compiler.

This fixes the issue with inability to use the goog.async.Deferred
class, since it's defined in a file that was under this third_party
subdirectory of the Closure Library, which we didn't include before.

This is a preparatory CL for introducing Emscripten/WebAssembly support
into the Example C++ App (tracked by #220), since the proposed way of
loading the Emscripten module will involve goog.async.Deferred.

### DIFF
--- a/third_party/closure-library/include.mk
+++ b/third_party/closure-library/include.mk
@@ -40,6 +40,9 @@ CLOSURE_LIBRARY_SOURCES_AND_DIRS := \
 # Function for obtaining the input list for the Closure Compiler that correspond
 # to the Closure Library.
 #
+# The "third_party/closure" directory is additionally included, since it
+# contains the deferred.js file.
+#
 # Test and perf files are excluded from compilation, because some of them may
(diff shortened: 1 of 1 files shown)


third_party/closure-library/include.mk
```diff
@@ -40,13 +40,17 @@ CLOSURE_LIBRARY_SOURCES_AND_DIRS := \
 # Function for obtaining the input list for the Closure Compiler that correspond
 # to the Closure Library.
 #
+# The "third_party/closure" directory is additionally included, since it
```
Fabian-Sommer, line 43: nit: `The "third_party/closure" directory is included since it ...`

Fabian-Sommer (approved):
LGTM % nit

PR #248, opened 2020-11-09
Migrate built_in_pin_dialog_server.cc to Value

emaxx-google:
Change the built_in_pin_dialog_server.cc file in the C++ Example App
to use the toolchain-independent Value class, instead of Native Client
specific pp::Var class and related.

This contributes to the effort of migrating the
//example_cpp_smart_card_client_app App to support WebAssembly (as
tracked by #220).


Fabian-Sommer (approved):
LGTM

PR #249, opened 2020-11-09
Migrate cert provider api_bridge.{h|cc} to Value

emaxx-google:
Change the chrome_certificate_provider/api_bridge.{h|cc} files in the
C++ Example App to use the toolchain-independent Value class, instead
of Native Client specific pp::Var class and related.

This contributes to the effort of migrating the
//example_cpp_smart_card_client_app App to support WebAssembly (as
tracked by #220).

### DIFF
--- a/example_cpp_smart_card_client_app/src/chrome_certificate_provider/api_bridge.cc
+++ b/example_cpp_smart_card_client_app/src/chrome_certificate_provider/api_bridge.cc
@@ -17,18 +17,15 @@
 #include <thread>
 #include <utility>
 
-#include <ppapi/cpp/var.h>
-
 #include <google_smart_card_common/global_context.h>
 #include <google_smart_card_common/logging/function_call_tracer.h>
 #include <google_smart_card_common/logging/logging.h>
-#include <google_smart_card_common/pp_var_utils/construction.h>
-#include <google_smart_card_common/pp_var_utils/debug_dump.h>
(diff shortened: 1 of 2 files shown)


example_cpp_smart_card_client_app/src/chrome_certificate_provider/api_bridge.cc
```diff
@@ -61,9 +58,11 @@ void ProcessCertificatesRequest(
...
   GOOGLE_SMART_CARD_CHECK(locked_certificates_request_handler);
   if (locked_certificates_request_handler->HandleRequest(&certificates)) {
-    // TODO(#220): Build `Value` directly, without converting from `pp::Var`.
-    result_callback(gsc::GenericRequestResult::CreateSuccessful(
-        gsc::ConvertPpVarToValueOrDie(gsc::MakeVarArray(certificates))));
+    gsc::Value response(gsc::Value::Type::kArray);
+    response.GetArray().push_back(gsc::MakeUnique<gsc::Value>(
+        gsc::ConvertToValueOrDie(std::move(certificates))));
+    result_callback(
+        gsc::GenericRequestResult::CreateSuccessful(std::move(response)));
```
emaxx-google, line 65: Probably I should introduce a helper that transforms the given sequence of C++ objects into the resulting array `Value`, or even into a `GenericRequestResult`. But I'd prefer to postpone that cleanup till later.

Fabian-Sommer (approved):
LGTM

PR #250, opened 2020-11-09
ifdef NaCl code in scard_structs_serialization

emaxx-google:
Surround the Native Client specific code with #ifdef's in
scard_structs_serialization.{h|cc}, so that it's only compiled in Native
Client builds. This way, these files can be used both by code that is
fully migrated to Emscripten/WebAssembly and the code that is still
using Native Client specific primitives.

This is a temporary step in order to accomplish the migration of the
//example_cpp_smart_card_client_app App (tracked by #220) earlier,
before the Smart Card Connector App gets migrated as well. Later,
once everything is migrated, all these #ifdef's together with the
code inside them will be removed.

### DIFF
--- a/third_party/pcsc-lite/naclport/common/src/google_smart_card_pcsc_lite_common/scard_structs_serialization.h
+++ b/third_party/pcsc-lite/naclport/common/src/google_smart_card_pcsc_lite_common/scard_structs_serialization.h
@@ -24,18 +24,14 @@
 // THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 // This file contains structure and function definitions that help to perform
-// (de)serialization of PC/SC-Lite API structures from/to Pepper values.
+// (de)serialization of PC/SC-Lite API structures from/to `Value`s.
 //
-// One of the reasons of defining separate intermediate structures is that the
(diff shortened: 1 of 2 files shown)


third_party/pcsc-lite/naclport/common/src/google_smart_card_pcsc_lite_common/scard_structs_serialization.h
```diff
@@ -24,18 +24,14 @@
 // THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 // This file contains structure and function definitions that help to perform
-// (de)serialization of PC/SC-Lite API structures from/to Pepper values.
+// (de)serialization of PC/SC-Lite API structures from/to `Value`s.
 //
-// One of the reasons of defining separate intermediate structures is that the
-// original PC/SC API functions do not belong to the google_smart_card
```
emaxx-google, line 30: (This issue with namespaces isn't valid anymore due to the way how the new `value_conversion.h` is implemented, so I just  deleted this.)

Fabian-Sommer (approved):
LGTM

emaxx-google:
Thank you! That was the last PR for today's batch :-)

PR #252, opened 2020-11-09
Not install pcsc-lite/naclport/cpp_client headers

emaxx-google:
This commit disables installation of "public" headers from the
//third_party/pcsc-lite/naclport/cpp_client library into a common
include directory. Instead, the path to this library is simply
added into the include search path in needed makefiles.

More background on this refactoring is in #132.


Fabian-Sommer (approved):
LGTM

emaxx-google:
Thanks for the quick review.

PR #254, opened 2020-11-10
Implement message channel for Emscripten JS loader

emaxx-google:
Extend the JavaScript "EmscriptenModule" class to have a bidirectional
message pipe with the executable. This message pipe conforms to the
Closure Library's goog.messaging.MessageChannel interface. Messages to
the module are sent by calling the "postMessage()" method on the
GoogleSmartCardModule class exposed by it via Embind. Messages in the
reverse direction are received via the callback passed to the class'
constructor.

This contributes to the WebAssembly migration effort, as tracked
by #220.


Fabian-Sommer (approved):
LGTM

emaxx-google:
Thank you for the quick reviews!

PR #255, opened 2020-11-10
Make PostMessageToJs() consume the Value object

emaxx-google:
Change the GlobalContext::PostMessageToJs()'s signature to take the
Value object by value (instead of const-ref) and remove the boolean
return value.

The reason is that on the Emscripten/WebAssembly platform we'll need to
schedule the message sending job to a different thread (in case we're not
on the main thread), which means that we want to grab the ownership of
the value until it's actually sent. Moreover, it means that we cannot
reliably tell the caller whether the delivery was successful (because it'll
happen asynchronously), therefore the return value is changed to "void".

This is a preparatory commit for the abovementioned change. The overall
WebAssembly migration effort of this library is tracked by #185.

### DIFF
--- a/common/cpp/src/google_smart_card_common/global_context.h
+++ b/common/cpp/src/google_smart_card_common/global_context.h
@@ -27,10 +27,9 @@ class GlobalContext {
  public:
   virtual ~GlobalContext() = default;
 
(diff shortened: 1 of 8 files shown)


Fabian-Sommer (approved):
LGTM

PR #256, opened 2020-11-10
Add forgotten Emscripten-Value conversion tests

emaxx-google:
Add the "value_emscripten_val_conversion_unittest.cc" file into build
scripts, so that it's executed with all other tests.

Previously it was forgotten (in both #211 and #218), likely due to some
merge conflict. The tracking issue is #185.


Fabian-Sommer (approved):
LGTM

PR #257, opened 2020-11-10
Fix setCertificates() JS error when there's no API

emaxx-google:
Fix the JavaScript exception being raised during the
chrome.certificateProvider.setCertificates() method call, in case the
whole chrome.certificateProvider API is unavailable.

This is expected to be a rare corner case - for example, when running
the Example C++ App in a desktop Chrome browser - nevertheless, it's
better to fix it to simplify testing and debugging it. Note that the
consequence of the bug was not only a JS error in the console, but also
hanging of the setCertificates() call.


Fabian-Sommer (approved):
LGTM

PR #258, opened 2020-11-10
Delete NaCl conv from cert provider types.{h|cc}

emaxx-google:
Delete the code for conversions to/from Native Client pp::Var types from
chrome_certificate_provider/types.{h|cc} files in
//example_cpp_smart_card_client_app. There's no code that uses them
anymore, and the conversions to/from the toolchain-independent Value
class are used instead.

This commit makes it possible to compile these files in
Emscripten/WebAssembly builds, contributing to the migration effort
tracked by #220.


Fabian-Sommer (approved):
LGTM

PR #261, opened 2020-11-11
Introduce top-level Makefile

emaxx-google:
This commit adds a new makefile, located in the project's root
directory. This makefile can be used to build the whole project in the
specified configuration, namely the TOOLCHAIN (can be "pnacl" [default]
or "emscripten") and the CONFIG (can be "Release" [default] or "Debug").

The main benefit is that the makefile syntax is more suitable than shell
scripts for describing rules like "target foo should always be compiled,
but target bar only under toolchain X" and similar more complex ones.
The makefile also naturally describes dependencies and prerequisites, as
opposed to hardcoding the "magically working" order of targets in shell
scripts.

Currently, only a few targets can be built in the TOOLCHAIN=emscripten
configuration. Follow-up changes will migrate the remaining targets, so
that everything can be compiled in all configurations.

Potentially, the //make-all.sh, //clean-all.sh, //test-all.sh scripts
will be rewritten to use this new makefile. This will make those scripts
simpler and keep the information about targets and their dependencies in
a single place.

This commit contributes to the WebAssembly migration effort, tracked
by #177.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,6 +133,6 @@ endef
 # "make test" will compile and execute all unit and integration tests.
 # Implementation note: A single rule with a sequence of commands in the recipe
 # is used, so that tests don't run in parallel (otherwise they might break).
+.PHONY: test
 test: all
 	$(foreach target,$(TEST_TARGETS),$(call DO_RUN_TEST,$(target)))
-.PHONY: test
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,138 @@
+# Copyright 2020 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,138 @@
...
+TEST_TARGETS += \
+	common/integration_testing/build \
+	example_cpp_smart_card_client_app/build/integration_tests \
+	third_party/libusb/naclport/build/tests \
+
+example_cpp_smart_card_client_app/build/integration_tests: common/cpp/build
+example_cpp_smart_card_client_app/build/integration_tests: common/integration_testing/build
+example_cpp_smart_card_client_app/build/integration_tests: example_cpp_smart_card_client_app/build
+third_party/libusb/naclport/build/tests: common/cpp/build
+third_party/libusb/naclport/build/tests: third_party/ccid/naclport/build
```
emaxx-google, line 95: self-nit: The prerequisite actually needs to be "third_party/libusb/naclport/build".

Fabian-Sommer (approved):
LGTM

PR #262, opened 2020-11-11
Emscripten: rebuild on library/dependency change

emaxx-google:
This commit adds automatic detection of static library or any other
specified dependency changes, triggering the rebuild of the resulting
Emscripten .wasm/.js files.

This resolves one of the feature gaps between our Emscripten builds
scripts and the Native Client ones. This effort is tracked by #177.

The commit also drops DEPS dependencies from the
//example_cpp_smart_card_client_app, since the referenced targets aren't
implemented in Emscripten builds. Moreover, our new approach will be to
drop these DEPS wherever possible, and only leave the dependencies
described in the root //Makefile; this will make the build scripts simpler
and more reliable.

### DIFF
--- a/common/make/internal/executable_building_emscripten.mk
+++ b/common/make/internal/executable_building_emscripten.mk
@@ -186,10 +186,15 @@ endef
 # L: Add the specified directory into the .a library search paths.
 # l: Link the specified .a library.
 define LINK_EXECUTABLE_RULE
+# Dependency on the input object files:
 $(BUILD_DIR)/$(TARGET).js $(BUILD_DIR)/$(TARGET).wasm $(BUILD_DIR)/$(TARGET).worker.js: $(foreach src,$(1),$(call OBJ_FILE_NAME,$(src)))
+# Dependency on the input static libraries:
 $(BUILD_DIR)/$(TARGET).js $(BUILD_DIR)/$(TARGET).wasm $(BUILD_DIR)/$(TARGET).worker.js: $(foreach lib,$(2),$(LIB_DIR)/lib$(lib).a)
+# Dependency on additionally specified targets:
 $(BUILD_DIR)/$(TARGET).js $(BUILD_DIR)/$(TARGET).wasm $(BUILD_DIR)/$(TARGET).worker.js: $(3)
+# Order-only dependency on the destination directory stamp:
 $(BUILD_DIR)/$(TARGET).js $(BUILD_DIR)/$(TARGET).wasm $(BUILD_DIR)/$(TARGET).worker.js: | $(BUILD_DIR)/dir.stamp
+# The recipe that performs the linking and creates the resulting files:
 $(BUILD_DIR)/$(TARGET).js $(BUILD_DIR)/$(TARGET).wasm $(BUILD_DIR)/$(TARGET).worker.js &:
 	emcc \
 		-o $(BUILD_DIR)/$(TARGET).js \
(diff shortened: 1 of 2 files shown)


common/make/internal/executable_building_emscripten.mk
```diff
@@ -186,7 +186,11 @@ endef
 # L: Add the specified directory into the .a library search paths.
 # l: Link the specified .a library.
 define LINK_EXECUTABLE_RULE
-$(BUILD_DIR)/$(TARGET).js $(BUILD_DIR)/$(TARGET).wasm $(BUILD_DIR)/$(TARGET).worker.js &: $(foreach src,$(1),$(call OBJ_FILE_NAME,$(src))) | $(BUILD_DIR)/dir.stamp
+$(BUILD_DIR)/$(TARGET).js $(BUILD_DIR)/$(TARGET).wasm $(BUILD_DIR)/$(TARGET).worker.js: $(foreach src,$(1),$(call OBJ_FILE_NAME,$(src)))
+$(BUILD_DIR)/$(TARGET).js $(BUILD_DIR)/$(TARGET).wasm $(BUILD_DIR)/$(TARGET).worker.js: $(foreach lib,$(2),$(LIB_DIR)/lib$(lib).a)
+$(BUILD_DIR)/$(TARGET).js $(BUILD_DIR)/$(TARGET).wasm $(BUILD_DIR)/$(TARGET).worker.js: $(3)
+$(BUILD_DIR)/$(TARGET).js $(BUILD_DIR)/$(TARGET).wasm $(BUILD_DIR)/$(TARGET).worker.js: | $(BUILD_DIR)/dir.stamp
+$(BUILD_DIR)/$(TARGET).js $(BUILD_DIR)/$(TARGET).wasm $(BUILD_DIR)/$(TARGET).worker.js &:
```
emaxx-google, line 193: I've decided to split this into separate lines, so that it's possible to read the prerequisites without scrolling a lot.

The alternative would be to use backslashes, like describe there: https://stackoverflow.com/questions/38086542/break-long-dependencies-in-makefile-into-several-lines/38096607
Do you think it'd be more readable that way?

Fabian-Sommer: I think the current way is fine - though I am confused as to what the last line does

common/make/internal/executable_building_emscripten.mk
```diff
@@ -186,7 +186,11 @@ endef
 # L: Add the specified directory into the .a library search paths.
 # l: Link the specified .a library.
 define LINK_EXECUTABLE_RULE
-$(BUILD_DIR)/$(TARGET).js $(BUILD_DIR)/$(TARGET).wasm $(BUILD_DIR)/$(TARGET).worker.js &: $(foreach src,$(1),$(call OBJ_FILE_NAME,$(src))) | $(BUILD_DIR)/dir.stamp
+$(BUILD_DIR)/$(TARGET).js $(BUILD_DIR)/$(TARGET).wasm $(BUILD_DIR)/$(TARGET).worker.js: $(foreach src,$(1),$(call OBJ_FILE_NAME,$(src)))
+$(BUILD_DIR)/$(TARGET).js $(BUILD_DIR)/$(TARGET).wasm $(BUILD_DIR)/$(TARGET).worker.js: $(foreach lib,$(2),$(LIB_DIR)/lib$(lib).a)
+$(BUILD_DIR)/$(TARGET).js $(BUILD_DIR)/$(TARGET).wasm $(BUILD_DIR)/$(TARGET).worker.js: $(3)
```
Fabian-Sommer, line 191: Does this still need to be here? In the call in the file below, this has only two parameters.

common/make/internal/executable_building_emscripten.mk
emaxx-google: PTAL - I've added a comment for each of the lines.

Fabian-Sommer (approved):
LGTM

emaxx-google:
Thanks!

PR #263, opened 2020-11-11
Emscripten: rebuild on included header change

emaxx-google:
This commit adds automatic detection of .h header file change,
triggering the recompilation of the source .c/.cc files that included
it (directly or transitively).

This resolves one of the feature gaps between our Emscripten builds
scripts and the Native Client ones. This effort is tracked by #177.


Fabian-Sommer (approved):
LGTM

PR #265, opened 2020-11-11
Inherit JS EmscriptenModule from ExecutableModule

emaxx-google:
Make the GoogleSmartCard.EmscriptenModule JavaScript class inherit
from the GoogleSmartCard.ExecutableModule one, and implement all missing
abstract methods.

This makes it possible to use the EmscriptenModule instances in the
toolchain-agnostic JavaScript code that doesn't care about differences
between Native Client and Emscripten/WebAssembly. This commit
contributes to the migration effort tracked by #220.


Fabian-Sommer (approved):
LGTM

PR #268, opened 2020-11-12
Catch Emscripten logs and handle crashes

emaxx-google:
This commit adds an observer that's called when the Emscripten module
crashes. It also adds observers for stdout/stderr messages received from
the module, forwarding these messages to the JavaScript logger (making
them discoverable in DevTools and also exportable via functionality like
"Export logs").

This contributes to the Emscripten/WebAssembly migration effort tracked
by #220.

### DIFF
--- a/common/js/src/executable-module/emscripten-module.js
+++ b/common/js/src/executable-module/emscripten-module.js
@@ -180,6 +180,8 @@ EmscriptenModule.prototype.onModuleCrashed_ = function() {
  * @private
  */
 EmscriptenModule.prototype.onModuleStdoutReceived_ = function(message) {
+  // TODO(#185): Consider downgrading to the "fine" level once structured log
+  // messages are received from the module.
   this.fromModuleMessagesLogger_.info(message);
 };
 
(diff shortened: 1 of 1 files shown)


common/js/src/executable-module/emscripten-module.js
```diff
@@ -146,6 +146,52 @@ EmscriptenModule.prototype.load_ = async function() {
...
+  this.fromModuleMessagesLogger_.info(message);
+};
+
+/**
+ * Handles an stderr message received from the Emscripten module.
+ * @param {string} message
+ * @private
+ */
+EmscriptenModule.prototype.onModuleStderrReceived_ = function(message) {
+  this.fromModuleMessagesLogger_.info(message);
```
Fabian-Sommer, line 192: nit: Is info appropriate for a message from stderr?

emaxx-google: Currently, all macros like `GOOGLE_SMART_CARD_LOG_` emit messages to stderr, not stdout: https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/blob/11f3044fc54071654bcf3ea0fce48aa14d6ca2be/common/cpp/src/google_smart_card_common/logging/logging.cc#L144
So with the current approach we do need to log stderr by default.

Having said that, I'm planning to eventually adopt a more fine-grained approach that we had for NaCl in the past, which was to send the log messages to JavaScript as form of tuples `(log_severity, log_message)`. After doing that, we'll be able to reestimate how useful stderr logs are.

Let me add a TODO into the CL to point this out.

Fabian-Sommer (approved):
LGTM

PR #270, opened 2020-11-12
Free C++ object on Emscripten module disposal

emaxx-google:
Make sure the destructor of the GoogleSmartCardModule C++ class is
called when the JavaScript class GoogleSmartCard.EmscriptenModule is
disposed of. This makes sure the C++ memory is freed in this flow.

This change is a small cleanup that isn't extremely useful for
real-world use cases, since a smart card client app never shuts down the
module (unless it crashes, but in that case the whole app is
self-reloaded). But the code is more sane this way, and also might make
integration tests consume less memory. The tracking issue is #220.


Fabian-Sommer (approved):
LGTM

MahmoudAGawad (approved):
Good catch! LGTM.

emaxx-google:
Thanks for the fast reviews!

PR #271, opened 2020-11-12
Make Emscripten C++ object untyped in JavaScript

emaxx-google:
Mark the reference to the C++ "GoogleSmartCardModule" object in
JavaScript as having unknown type ("?" in the Closure Compiler
notation), so that the compiler doesn't emit warnings on accessing
unknown properties of it.

This is small cleanup for the effort tracked by #220.


Fabian-Sommer (approved):
LGTM

PR #272, opened 2020-11-12
.gitignore Emscripten temporary build artifacts

emaxx-google:
Add directories that are produced by Emscripten/WebAssembly builds in
this project into .gitignore files, as these are temporary files that
shouldn't be tracked by the version control system.

This is a minor cleanup for the effort tracked by #177.


Fabian-Sommer (approved):
LGTM

PR #274, opened 2020-11-12
Put built JS files into dir depending on toolchain

emaxx-google:
Make compiled JavaScript files be stored in different locations
depending on the TOOLCHAIN environment variable.

This fixes the problem that JS files can stay non-recompiled after
switching the TOOLCHAIN variable. E.g., in the following scenario
resulting JS files would still refer to emscripten if built without this
commit:

  TOOLCHAIN=emscripten make && TOOLCHAIN=pnacl make

The underlying reason is that when both the input files are unchanged
and the resulting file paths don't depend on the toolchain, Make
doesn't see any reason to trigger recompilation. We don't have
prerequisites that depend on the compilation flags, and this would
be tricky to implement.

The solution in this commit makes sure that JS files are put into
different locations for pnacl and emscripten, which bypasses the problem
with missing recompilation. The only downside is a bit of uglyness for
the JS files that don't depend on the toolchain, but it's only a problem
for intermediate build files.

This commit contributes to the WebAssembly migration effort tracked
by #177.

### DIFF
--- a/common/js/src/executable-module/executable-module.js
+++ b/common/js/src/executable-module/executable-module.js
@@ -54,8 +54,7 @@ ExecutableModule.Toolchain = {
 };
 
 /** @define {string} */
-const TOOLCHAIN = goog.define(
-    'GoogleSmartCard.ExecutableModule.TOOLCHAIN', 'pnacl');
+const TOOLCHAIN = goog.define('GoogleSmartCard.ExecutableModule.TOOLCHAIN', '');
 GSC.Logging.check(
     Object.values(GSC.ExecutableModule.Toolchain).includes(TOOLCHAIN),
     'Unexpected value of GoogleSmartCard.ExecutableModule.TOOLCHAIN: ' +
--- a/common/make/executable_building.mk
+++ b/common/make/executable_building.mk
@@ -16,13 +16,8 @@
 # can be run in web-based applications.
 #
 # The actually used technology depends on the TOOLCHAIN environment variable.
-# Currently, the only supported value is "pnacl", which corresponds to the
-# Google Native Client technology.
 #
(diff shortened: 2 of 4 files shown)


common/js/src/executable-module/executable-module.js
```diff
@@ -54,8 +54,7 @@ ExecutableModule.Toolchain = {
 };
 
 /** @define {string} */
-const TOOLCHAIN = goog.define(
-    'GoogleSmartCard.ExecutableModule.TOOLCHAIN', 'pnacl');
```
emaxx-google, line 58: (To explain, this default here isn't needed anymore, since common.mk now makes sure that the variable is always set to something.)

common/make/executable_building.mk
```diff
@@ -16,14 +16,9 @@
 # can be run in web-based applications.
 #
 # The actually used technology depends on the TOOLCHAIN environment variable.
-# Currently, the only supported value is "pnacl", which corresponds to the
-# Google Native Client technology.
 #
 # common.mk must be included before including this file.
 
-# Default to using the "pnacl" toolchain.
-TOOLCHAIN ?= pnacl
```
emaxx-google, line 25: (Ditto.)

common/make/js_building_common.mk
```diff
@@ -123,13 +127,6 @@ JS_BUILD_COMPILATION_FLAGS += \
 	--use_types_for_optimization=false \
 	--warning_level=VERBOSE \
 
-ifneq ($(TOOLCHAIN),)
```
emaxx-google, line 126: (Ditto.)

Fabian-Sommer (approved):
LGTM

PR #275, opened 2020-11-13
Delete old references to NaCl from Example C++ App

emaxx-google:
Rename variables/functions and reword comments to use generic terms
instead of referring to Native Client, since the code can be compiled in
two modes now: Native Client and Emscripten/WebAssembly.

This is a pure refactoring CL; the tracking issue is #220.


Fabian-Sommer (approved):
LGTM

PR #280, opened 2020-11-16
Delete header installation scripts

emaxx-google:
This commit deletes the makefile rules that implement the
installation of libraries' headers into the comment include directory.
All usages of these rules have been removed by now.

Also fix one remaining usage of common installed headers - the
//common/integration_testing library - which was incorrectly specifying
the include directories.

This completes the refactoring tracked by #132.


Fabian-Sommer (approved):
LGTM

PR #281, opened 2020-11-16
Suppress JS error on message port connection error

emaxx-google:
This commit avoids emitting the following JavaScript error in the
DevTools console in case the cross-extension message port fails to
connect:

  "Unchecked runtime.lastError: Could not establish connection.
  Receiving end does not exist."

This could happen when, for example, the middleware app tries to talk to
the Connector app, but the latter is not installed.

This commit is only a cosmetic improvement: instead of logging something
about unhandled errors, which might be confusing, an "info" log message
is printed. The actual error handling isn't affected by this commit: as
before, the connection error causes the message channel to be shut down.


Fabian-Sommer (approved):
LGTM

PR #282, opened 2020-11-30
libusb: Drop unused variable

emaxx-google:
This commit is a small cleanup: an unused variable is deleted from
LibusbOverChromeUsb::WrapLibusbTransferCallback().

This is a preparation for migrating the libusb library port to
Emscripten, since the latter's compiler detects the unused lambda
capture as a warning (causing a compilation error).


Fabian-Sommer (approved):
LGTM

Issue #233, opened 2020-11-04
Add WebAssembly support into Smart Card Connector app


emaxx-google:
The Smart Card Connector app (//smart_card_connector) and its dependencies (//third_party/libusb, //third_party/ccid, //third_party/pcsc-lite and others) need to be refactored and extended in order to support working under WebAssembly. Currently it only compiles under Native Client.

Depends on: #185 (migration of the //common/cpp library to WebAssembly).

PR #283, opened 2020-11-30
libusb: Migrate to Emscripten

emaxx-google:
This commit allows to build our port of the libusb library under
Emscripten.

For the background, our libusb implementation is header-compatible with
the third-party libusb library, but its implementation is completely
different from the latter: instead of talking to drivers and OS-level
components it redirects all requests to the chrome.usb Extensions
API. Before this commit, our port could only be compiled under
Native Client.

This commit contributes to the effort of migrating the Smart Card
Connector app to Emscripten/WebAssembly, as tracked by #233.

### DIFF
--- a/third_party/libusb/naclport/src/libusb_over_chrome_usb_unittest.cc
+++ b/third_party/libusb/naclport/src/libusb_over_chrome_usb_unittest.cc
@@ -39,17 +39,17 @@
 #include "chrome_usb/api_bridge_interface.h"
 #include "chrome_usb/types.h"
 
-#ifndef __EMSCRIPTEN__
+#ifdef __native_client__
 // Native Client's version of Google Test uses a different name of the macro for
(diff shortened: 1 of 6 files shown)


third_party/libusb/naclport/src/libusb_over_chrome_usb_unittest.cc
```diff
@@ -39,12 +39,17 @@
 #include "chrome_usb/api_bridge_interface.h"
 #include "chrome_usb/types.h"
 
-// Google Mock doesn't provide the C++11 "override" specifier for the mock
-// method definitions
+#ifndef __EMSCRIPTEN__
```
Fabian-Sommer, line 42: nit: `#ifdef __NATIVE_CLIENT__`

Fabian-Sommer (approved):
LGTM % nit

PR #284, opened 2020-11-30
ccid: Enable compilation under Emscripten

emaxx-google:
This commit enables compiling the CCID Free Software Driver in the
Emscripten/WebAssembly mode.

The code in this driver is a pure C code that doesn't use itself
low-level OS interfaces (like direct USB access), which is why it can
be compiled using Emscripten without code adaptations.

This commit contributes to the Emscripten/WebAssembly migration
effort tracked by #233.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,7 @@ example_cpp_smart_card_client_app/build: third_party/pcsc-lite/naclport/cpp_demo
 TEST_TARGETS := \
 	common/cpp/build/tests \
 	common/js/build/unittests \
+	third_party/ccid/naclport/build \
 	third_party/libusb/naclport/build/js_unittests \
 	third_party/libusb/naclport/build/tests \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -82,7 +82,6 @@ smart_card_connector_app/build: third_party/libusb/naclport/build
 smart_card_connector_app/build: third_party/pcsc-lite/naclport/common/build
 smart_card_connector_app/build: third_party/pcsc-lite/naclport/server/build
 smart_card_connector_app/build: third_party/pcsc-lite/naclport/server_clients_management/build
-third_party/ccid/naclport/build: third_party/libusb/naclport/build
```
emaxx-google, line 85: This dependency is not needed anymore, now that we don't install libusb's headers anymore.

Fabian-Sommer (approved):
LGTM

PR #285, opened 2020-11-30
pcsc-lite: Migrate server/daemon to Emscripten

emaxx-google:
This commit allows to compile the "pcsc_lite_server" component
(which is a webport of the PC/SC-Lite daemon) under Emscripten.

Brief overview of the changes:
* Replace the usage of Native Client specific pp::Instance class with
  the toolchain-agnostic GlobalContext class;
* Replace the usage of Native Client specific pp::Var* classes with
  the toolchain-independent Value class;
* Drop unneeded dependency on the compilation of "libusb" (it's not
  required since the change that disabled installation of libusb headers
  into a common directory);
* Change the top-level makefile to compile the server regardless of the
  chosen toolchain.

as tracked by #233.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,7 @@ TARGETS := \
 	third_party/pcsc-lite/naclport/common/build \
 	third_party/pcsc-lite/naclport/cpp_client/build \
 	third_party/pcsc-lite/naclport/cpp_demo/build \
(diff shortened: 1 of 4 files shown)


Fabian-Sommer (approved):
LGTM

PR #286, opened 2020-12-02
Enable more Emscripten debugging capabilities

emaxx-google:
This commit enables several Emscripten debugging features when the code
is compiled in the CONFIG=Debug mode.

These debugging features (more extensive debug symbols, runtime memory
access checks, etc.) will aid developers to debug the applications.
There's no change for the Release builds, since many of these checks
introduce a significant runtime overhead.

This commit contributes to the WebAssembly migration effort, as tracked
by #177.


Fabian-Sommer (approved):
LGTM

PR #288, opened 2020-12-02
Make clean-all.sh preserve env by default

emaxx-google:
Change the //clean-all.sh script to keep the //env intact by default.
This means that the SDKs downloaded by the //env/initialize.sh script
(Emscripten and Native Client SDK) are preserved.

This makes the clean-all.sh script more convenient to use, since
developers usually need to delete intermediate build artifacts but not
SDKs, and reinitializing SDKs is a time-consuming step.

The "-f" option is also added that reverts to the old behavior, cleaning
both build artifacts and the env.

### DIFF
--- a/clean-all.sh
+++ b/clean-all.sh
@@ -15,7 +15,8 @@
 # limitations under the License.
 
 # This script removes the temporary files and output files produced by building
-# of the libraries and Apps of this project, and clears the env.
+# of the libraries and Apps of this project. Optionally, if the "-f" flag is
+# specified, it clears the SDKs stored under //env/ as well.
 #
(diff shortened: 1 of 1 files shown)


Fabian-Sommer (commented):
This looks good, but I think the option should be documented somewhere.
(e.g. in the file comment, which also speaks about clearing the env)

emaxx-google:
> This looks good, but I think the option should be documented somewhere.
> (e.g. in the file comment, which also speaks about clearing the env)

Done (comment).

Fabian-Sommer (approved):
LGTM

PR #290, opened 2020-12-02
Add conversion of Value to/from nullptr

emaxx-google:
Implement a special case in the to/from Value conversions that converts
a null value into nullptr and vice versa.

This is a "fake" conversion in the sense that it's only used to denote
the type being null - there's no other payload involved. This special
case will be used in follow-up commits in order to support functions
whose input or output argument must always be null (such functions exist
in the PC/SC-Lite API). Previously, the Native Client's pp::Var::Null
type has been used for such arguments, but this commit adds a
toolchain-independent alternative that can be used in WebAssembly
builds.

tracked by #185.

### DIFF
--- a/common/cpp/src/google_smart_card_common/value_conversion.cc
+++ b/common/cpp/src/google_smart_card_common/value_conversion.cc
@@ -17,6 +17,7 @@
 #include <inttypes.h>
 #include <stdint.h>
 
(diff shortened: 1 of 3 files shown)


Fabian-Sommer (approved):
LGTM

PR #291, opened 2020-12-02
Migrate clients_manager.cc to Value

emaxx-google:
Change the implementation in clients_manager.cc to use the
toolchain-agnostic Value class instead of Native Client specific
primitives.

This contributes to the Emscripten/WebAssembly migration effort tracked
by #233.


Fabian-Sommer (approved):
LGTM

PR #292, opened 2020-12-02
Migrate client_request_processor.{h,cc} to Value

emaxx-google:
Change the implementation of the PC/SC-Lite server_clients_management
client_request_processor.{h,cc} files to use the toolchain-independent
Value class instead of the Native Client specific primitives.

This contributes to the Emscripten/WebAssembly migration effort, as
tracked by #233.


Fabian-Sommer (approved):
LGTM

PR #294, opened 2020-12-03
Support optional in RemoteCallArgumentsExtractor

emaxx-google:
Extend the RemoteCallArgumentsExtractor class to support extracting
optional arguments: for such arguments, the Value can be equal to null,
which denotes that the optional argument is unset.

This capability will be used in follow-up commits in order to support
receiving remote calls for functions with optional arguments.

as tracked by #233.


Fabian-Sommer (approved):
LGTM

PR #296, opened 2020-12-03
Extract NaCl-specific entry point of Connector app

emaxx-google:
Refactor the Smart Card Connector app's core code into a Native Client
specific entry_point_nacl.cc and a toolchain-independent
application.{h,cc} piece.

This commit is a preparation for implementing another - Emscripten
specific - entry point for WebAssembly builds. This migration effort is
tracked by #233.


Fabian-Sommer (approved):
LGTM

PR #297, opened 2020-12-03
Build PCSC server_clients_management in Emscripten

emaxx-google:
This commit allows building the PC/SC-Lite server_clients_management
library in the Emscripten mode. Previously it was built only for
Native Client builds, but due to the landed series of commits it can be
compiled using any of the two toolchains now.

tracked by #233.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,7 @@ TARGETS := \
 	third_party/pcsc-lite/naclport/cpp_client/build \
 	third_party/pcsc-lite/naclport/cpp_demo/build \
 	third_party/pcsc-lite/naclport/server/build \
+	third_party/pcsc-lite/naclport/server_clients_management/build \
 
 example_cpp_smart_card_client_app/build: common/cpp/build
 example_cpp_smart_card_client_app/build: third_party/pcsc-lite/naclport/common/build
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -75,15 +76,13 @@ else ifeq ($(TOOLCHAIN),pnacl)
...
 	smart_card_connector_app/build \
-	third_party/pcsc-lite/naclport/server_clients_management/build \
 
 smart_card_connector_app/build: common/cpp/build
 smart_card_connector_app/build: third_party/ccid/naclport/build
 smart_card_connector_app/build: third_party/libusb/naclport/build
 smart_card_connector_app/build: third_party/pcsc-lite/naclport/common/build
 smart_card_connector_app/build: third_party/pcsc-lite/naclport/server/build
 smart_card_connector_app/build: third_party/pcsc-lite/naclport/server_clients_management/build
-third_party/pcsc-lite/naclport/server_clients_management/build: third_party/pcsc-lite/naclport/server/build
```
emaxx-google, line 86: This dependency is not needed anymore, now that we don't install headers of the "/server/" library into a shared headers directory.

Fabian-Sommer (approved):
LGTM

PR #299, opened 2020-12-03
Support Emscripten in Connector's background JS

emaxx-google:
Add support of Emscripten modules into the Smart Card Connector's
background page JavaScript code. With this commit, the JS behavior will
depend on the TOOLCHAIN constant (which, in turn, is filled based on the
environment variable), loading either the Emscripten or the Native
Client module.

as tracked by #233.

### DIFF
--- a/smart_card_connector_app/src/background.js
+++ b/smart_card_connector_app/src/background.js
@@ -17,6 +17,8 @@
 goog.provide('GoogleSmartCard.ConnectorApp.BackgroundMain');
 
 goog.require('GoogleSmartCard.ConnectorApp.Background.MainWindowManaging');
+goog.require('GoogleSmartCard.EmscriptenModule');
+goog.require('GoogleSmartCard.ExecutableModule');
 goog.require('GoogleSmartCard.Libusb.ChromeLoginStateHook');
 goog.require('GoogleSmartCard.Libusb.ChromeUsbBackend');
 goog.require('GoogleSmartCard.LogBufferForwarder');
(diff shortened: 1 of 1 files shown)


smart_card_connector_app/src/background.js
```diff
@@ -44,37 +46,61 @@ const JS_LOGS_HANDLER_MESSAGE_TYPE = 'js_logs_handler';
...
-// Note that this object needs to be created as early as possible, in order to
-// not miss any important log.
-const logBufferForwarderToNaclModule = new GSC.LogBufferForwarder(
-    GSC.Logging.getLogBuffer(), JS_LOGS_HANDLER_MESSAGE_TYPE);
+if (GSC.ExecutableModule.TOOLCHAIN === GSC.ExecutableModule.Toolchain.PNACL) {
+  // Used to forward logs collected on the JS side to the NaCl module's stdout,
+  // in order to simplify accessing them in some configurations.
+  // Note that this object needs to be created as early as possible, in order to
+  // not miss any important log.
+  var logBufferForwarderToNaclModule = new GSC.LogBufferForwarder(
```
emaxx-google, line 54: Note: I wasn't able to use "let"/"const", since it'd be scoped to that "if". And declaring the variable with "let" before the "if" didn't work either - the compiler complained about assigning to an alias twice (which sounds like a compiler bug, but maybe I'm unaware about some assumption made in the Closure Compiler).

smart_card_connector_app/src/background.js
```diff
@@ -44,37 +46,61 @@ const JS_LOGS_HANDLER_MESSAGE_TYPE = 'js_logs_handler';
...
-  // immediately post the messages that have been accumulated so far.
-  logBufferForwarderToNaclModule.startForwarding(
-      naclModule.getMessageChannel());
-}, () => {});
+/**
+ * The binary executable module that contains the actual smart card client code.
+ * @type {!GSC.ExecutableModule}
+ */
+const executableModule = createExecutableModule();
+executableModule.addOnDisposeCallback(naclModuleDisposedListener);
```
Fabian-Sommer, line 86: nit: Rename "naclModuleDisposedListener" and change its message depending on TOOLCHAIN (unless this callback cannot happen for Wasm)

Fabian-Sommer (approved):
LGTM % nit

PR #300, opened 2020-12-03
Build Connector app in Emscripten

emaxx-google:
This commit allows building the while Smart Card Connector app in the
Emscripten mode. Previously it was built only for Native Client builds,
but due to the landed series of commits it can be compiled using any
of the two toolchains now.

Note that the Emscripten build isn't fully functional yet - this will be
addressed by follow-up commits.

tracked by #233.


Fabian-Sommer (approved):
LGTM

PR #301, opened 2020-12-07
Enable C++ exceptions in Emscripten builds

emaxx-google:
Enable the "DISABLE_EXCEPTION_CATCHING=0" flag when compiling C++ files
under Emscripten, so that the C++ exception support is enabled.

This flag was disabled by default since it affects performance and
binary size, but we're enabling this, as smart card middleware apps
reportedly rely on C++ exceptions. We don't set the flag only for
example_cpp_smart_card_client_app, since the flag must be used
consistently, which includes the common libraries that are linked into
it.

tracked by #220.


Fabian-Sommer (approved):
LGTM

PR #302, opened 2020-12-07
Preload needed static files for Emscripten builds

emaxx-google:
This commit adds the missing static file support into Emscripten builds,
by using the --preload-file flag to package ("preload") the files into a
.data file. This is used in order to configure a virtual file system
with the data files that are needed at runtime for the Smart Card
Connector app's executable module.

This commit also changes the paths that have been used in Native Client
builds, so that the same path works and is natural in both Native Client
and Emscripten configurations. The previously used "crx" directory
doesn't reflect the new approach, since under Emscripten we don't load
files directly from the extension/app's CRX file anymore.

This commit contributes to the Emscripten/WebAssembly migration, as
tracked by #233.


emaxx-google:
Mahmoud: PTAL as well. (I don't know, though, whether you already had experience with the `--preload-file` flag.)

MahmoudAGawad (approved):
LGTM. Thanks!

Fabian-Sommer (approved):
LGTM

PR #303, opened 2020-12-07
Fix shutdown bug in PcscLiteServerClientsManager

emaxx-google:
Fix the crash that's happening in case
PcscLiteServerClientsManager::Detach() was used, with the crash
potentially happenning during the app shutdown.

The bug is that the destructor will be calling Detach() for the second
time, meanwhile that method is asserting against being called multiple
times. The fix is to replace the assert with an early exit.

tracked by #233.


Fabian-Sommer (approved):
LGTM

PR #304, opened 2020-12-07
Omit "NaCl module" log prefix in Emscripten builds

emaxx-google:
This commit is a cosmetic improvement of log messages from the
Emscripten modules: the "NaCl module" prefix won't be printed
anymore in non-NaCl builds.

For the background, the reason why this prefix is added under Native
Client is that the NaCl logs are redirected to Chrome's own stderr logs.
Therefore, to simplify reading the logs, we added this distinguished
prefix in NaCl builds. In Emscripten, on the contrary, such prefixes are
useless as Chrome doesn't embed Emscripten logs into its own logs.

as tracked by #185.


Fabian-Sommer (approved):
LGTM

PR #305, opened 2020-12-15
Enable C++ exceptions in Emscripten Release builds

emaxx-google:
Set the "DISABLE_EXCEPTION_CATCHING=0" flag in Emscripten builds not
only in Debug but also in Release mode. The C++ exception support is
needed unconditionally.

tracked by #220.


Fabian-Sommer (approved):
LGTM

PR #306, opened 2021-01-08
Cherry-pick PC/SC-Lite bugfix for thread function

emaxx-google:
Cherry-pick the following commit from the upstream PC/SC-Lite
repository:

  199f8455745654d3856251a949ad630c4d1a3c13
  Fix compiler warning

  hotplug_libusb.c: In function ‘HPSearchHotPluggables’:
  hotplug_libusb.c:498:4: warning: cast between incompatible function
  types from ‘void (*)(int *)’ to ‘void * (*)(void *)’
  [-Wcast-function-type]
    498 |    (PCSCLITE_THREAD_FUNCTION( )) HPEstablishUSBNotifications, pipefd);
        |    ^

This cherry-picked fix is necessary for the WebAssembly build of
PC/SC-Lite to work (tracking issue #233), since it fixes the wrong
function signature that's causing stack corruption when run under
Emscripten. (For some reason, this bug wasn't causing any issues
when run as a native binary or under Native Client.)

The cherry-picked patch file can be deleted once the upstream PC/SC-Lite
repository makes a new release that includes this commit.

### DIFF
--- a/docs/updating-pcsc-lite.md
+++ b/docs/updating-pcsc-lite.md
@@ -55,22 +55,27 @@ somewhat outdated version of the PC/SC-Lite daemon.
    (Hint: Pay attention to the directory structure; e.g., you should have the
    `AUTHORS` file located at `//third_party/pcsc-lite/src/AUTHORS` in the end.)
 
-4. Edit the URL and the version in the `//third_party/pcsc-lite/README.google`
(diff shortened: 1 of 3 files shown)


Fabian-Sommer (approved):
LGTM

PR #307, opened 2021-01-11
Fix unordered move in socketpair_emulation

emaxx-google:
Fix a potential bug in PC/SC-Lite webport's socketpair_emulation.cc that
could lead to reading from a null shared pointer. The bug is that
std::move() and dereference are occurring within the same
sub-expression, for which C++ doesn't give guarantee about the ordering
between these two operations.

We don't have any evidence that this bug could occur in practice
(especially in the NaCl builds this code seemed to work reliable), but
it's important to fix this potential unspecified behavior anyway.


emaxx-google:
P.S. I wish the clang-tidy's bugprone-use-after-move diagnostic would've discovered that, but I just tried it for the whole repository and apparently it doesn't detect this issue.

Fabian-Sommer (approved):
LGTM

ayaNader:
LGTM

ayaNader (approved):
LGTM

PR #308, opened 2021-01-11
Fix missing Info.plist file build error

emaxx-google:
Make sure the Info.plist file is produced by the CCID webport's makefile
by default, since this file is used by other makefiles in this project.

This is a follow-up to change #302 (linked to the issue #233 that tracks
the Emscripten/WebAssembly migration), which mistakenly removed the
Info.plist file from being produced by default.


Fabian-Sommer (commented):
LGTM

Fabian-Sommer (approved):

PR #311, opened 2021-01-11
Cleanup returning const objects

emaxx-google:
Make sure that non-trivial local C++ variables that are returned by
value aren't marked as "const", since this is a pessimization that
prevents move'ing the returned object. This is mostly a refactoring,
since the performance difference must be negligible in these cases.

This was found via clang-tidy ("performance-no-automatic-move"
diagnostic).


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #312, opened 2021-01-11
Explicitly cast unowned release() to void

emaxx-google:
Add explicit casts to void in places that call
std::unique_ptr::release() without storing its result anywhere. This
clearly shows the programmer's intention, as opposed to potential places
where this could be done by mistake. This is a pure refactoring change,
made in order to improve readability and to silence the clang-tidy
diagnostic.

This was found by the "bugprone-unused-return-value" clang-tidy
diagnostic.


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #313, opened 2021-01-11
Make methods static when possible

emaxx-google:
Change a few C++ class methods to be static methods or global-namespace
functions, in cases when they don't use any of the class internal state.
This is a pure refactoring change.

This was found by clang-tidy (the
"readability-convert-member-functions-to-static" diagnostic).

### DIFF
--- a/third_party/libusb/naclport/src/libusb_over_chrome_usb_unittest.cc
+++ b/third_party/libusb/naclport/src/libusb_over_chrome_usb_unittest.cc
@@ -428,7 +428,6 @@ class LibusbOverChromeUsbTransfersTest
 
   static void SetUpTransferCallbackMockExpectations(
       size_t transfer_index,
-      bool /*is_output*/,
       bool is_canceled,
       MockFunction<void(libusb_transfer_status)>* transfer_callback) {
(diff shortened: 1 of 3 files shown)


Fabian-Sommer (approved):
LGTM

third_party/libusb/naclport/src/libusb_over_chrome_usb_unittest.cc
```diff
@@ -435,7 +435,7 @@ class LibusbOverChromeUsbTransfersTest
                 Call(GetExpectedTransferStatus(transfer_index, is_canceled)));
   }
 
-  void SetUpTransferCallbackMockExpectations(
+  static void SetUpTransferCallbackMockExpectations(
       size_t transfer_index,
       bool /*is_output*/,
```
ayaNader, line 440: could we remove this parameter since it's not used anyway?

emaxx-google: Indeed! Done.
(I guess I added them initially in order to follow the interface of other similar methods, but that's probably not very useful.)

ayaNader (approved):
LGTM

PR #316, opened 2021-01-11
Avoid unnamed namespace in a header file

emaxx-google:
Fix the pcsc_lite_over_requester.h header to not contain the unnamed
namespace - such namespace usage isn't canonical and can lead to symbol
duplication (across every .cc file that included it). Also, while we're
here, move the string constant from the .h file to the .cc file. This is
a pure refactoring change.

This was found by clang-tidy ("google-build-namespaces" and
"misc-definitions-in-headers" diagnostics).


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #317, opened 2021-01-11
Fix copy constructor signature

emaxx-google:
Fix the RemoteCallArgumentsExtractor's class copy constructor signature
to the canonical form: const-ref instead of a non-const-ref. This is a
pure refactoring change.

This was found by the "google-runtime-references" clang-tidy diagnostic.


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #318, opened 2021-01-11
Add missing include for size_t

emaxx-google:
Add inclusion of stddef.h into tuple_unpacking.h, since the latter uses
size_t, which is not a built-in type but rather a definition in one of
the standard header files. This is a pure refactoring change.

This issue was highlighted by the build error when running clang-tidy.

### DIFF
--- a/common/cpp/src/google_smart_card_common/tuple_unpacking.h
+++ b/common/cpp/src/google_smart_card_common/tuple_unpacking.h
@@ -15,6 +15,8 @@
 #ifndef GOOGLE_SMART_CARD_COMMON_TUPLE_UNPACKING_H_
 #define GOOGLE_SMART_CARD_COMMON_TUPLE_UNPACKING_H_
 
+#include <stddef.h>
+
 namespace google_smart_card {
 
 // Helper template class that holds a compile-time integer sequence as its


common/cpp/src/google_smart_card_common/tuple_unpacking.h
```diff
@@ -15,6 +15,8 @@
 #ifndef GOOGLE_SMART_CARD_COMMON_TUPLE_UNPACKING_H_
 #define GOOGLE_SMART_CARD_COMMON_TUPLE_UNPACKING_H_
 
+#include <stddef.h>
```
Fabian-Sommer, line 18: Is there a reason to prefer this over `#include <cstddef>`?

Fabian-Sommer: Answering myself: The style guide does not recommend one over the other, as long as it is consistent. Looking at a few adjacent files, the majority uses this style.
However, I found `#include <cstdarg>` in chromeos_smart_card_connector/common/cpp/src/google_smart_card_common/formatting.h, so perhaps that could be changed.

Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

common/cpp/src/google_smart_card_common/tuple_unpacking.h
emaxx-google: I believe, technically, cstddef provides only `std::size_t` and not `size_t`, meanwhile the code in this file uses the latter. I personally find it a bit annoying to write `std::` in front of such "almost" built-in type, but this is not a strong argument :)
WDYT?

Fabian-Sommer: I don't have a strong opinion for either here.

emaxx-google: Let's go with ".h" then, for now.

PR #321, opened 2021-01-25
Update Emscripten to 2.0.10

emaxx-google:
Bump the Emscripten dependency to the latest available version - 2.0.10.
This brings a few bugfixes for some issues in the Emscripten toolchain.


emaxx-google:
Actually, abandoning this - I realized that there was a regression in 2.0.10 that would make our Debug builds broken (https://github.com/emscripten-core/emscripten/issues/13191). The issue was already fixed upstream, so let's just wait for the newer release.

PR #322, opened 2021-01-25
Parallelize clean-all.sh

emaxx-google:
Speed up the clean-all.sh script by running make with "-j10", which
makes it process directories in parallel. This is the same number of
parallel processes that is used in the make-all.sh script.


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #324, opened 2021-02-06
Fix env/initialize.sh env error with present NaCl

emaxx-google:
Fix the "unbound error" occurring in env/initialize.sh in case when the
existing NaCl SDK is present but the environment doesn't contain the
NACL_SDK_ROOT variable.

This is a follow-up to the PR #178 where the problem was introduced,
attributed to tracking issue #177.

### DIFF
--- a/env/initialize.sh
+++ b/env/initialize.sh
@@ -65,6 +65,7 @@ initialize_depot_tools() {
 }
 
 initialize_nacl_sdk() {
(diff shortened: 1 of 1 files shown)


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #325, opened 2021-02-06
Don't pass Emscripten linker flags to compiler

emaxx-google:
This commit attempts to avoid passing useless flags to the Emscripten
compiler and linker (emcc), by maintaining compilation and linking flags
separately instead of putting all of them together. This makes the build
logs slightly more readable (as the number of command-line flags becomes
less on average) and also avoids nonsensical no-op flags (like
EXPORT_NAME pointing to the name of a static library).

The main information about which Emscripten flags are needed on which
stage is present in this file:
https://github.com/emscripten-core/emscripten/blob/master/src/settings.js

This commit is expected to bring no functional differences on the
resulting build program. The tracking issue is #177.

### DIFF
--- a/common/make/internal/executable_building_emscripten.mk
+++ b/common/make/internal/executable_building_emscripten.mk
@@ -65,53 +65,73 @@ define DEP_FILE_NAME
 $(BUILD_DIR)/$(basename $(subst ..,__,$(1))).d
 endef
 
-# Flags passed to the Emscripten compiler/linker tools.
+# Flags passed to the Emscripten compiler and linker tools.
 #
 # Explanation:
 # bind: Enables "Embind" (the technology used by our C++ wrappers for talking
 #   with JavaScript).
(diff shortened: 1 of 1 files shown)


common/make/internal/executable_building_emscripten.mk
```diff
@@ -65,53 +65,73 @@ define DEP_FILE_NAME
 $(BUILD_DIR)/$(basename $(subst ..,__,$(1))).d
 endef
 
-# Flags passed to the Emscripten compiler/linker tools.
+# Flags passed to the Emscripten compiler and linker tools.
```
Fabian-Sommer, line 68: nit: "Flags passed to both ..." to make it even more clear

Fabian-Sommer (approved):
LGTM % nit

ayaNader (approved):
LGTM

PR #327, opened 2021-02-06
Rename nacl to executable_module in Connector App

emaxx-google:
This commit renames the "nacl_module" directory into
"executable_module", as well as the resulting build artifact. The new
name is more correct, since we support compiling under either NaCl or
Emscripten now. This is a non-function change.

This commit contributes to the WebAssembly/Emscripten migration effort
tracked by #233.


Fabian-Sommer (approved):
LGTM

ayaNader (commented):
do we need to rename all instances of nacl or only some?

emaxx-google:
> do we need to rename all instances of nacl or only some?

Yes, there's still a bunch of places that need to be renamed, since they're actually reused between NaCl and Emscripten. This commit is just one step on that way (it's better to do it progressively, rather than in a single huge commit that's hard to review).

ayaNader (approved):
LGTM

PR #330, opened 2021-02-08
Fix JS exception in ArrayBuffer substitution test

emaxx-google:
Fix the JS error that was occurring in the
"testSubstituteArrayBuffersRecursively" JavaScript test, caused by some
global object not being mocked. The fix is to silence this error by
adding a check of that global object, which also makes the production
code slightly more bullet-proof.

This is a follow-up to PR #226, which introduced this test.


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #332, opened 2021-02-18
Import PC/SC-Lite 1.9.1

emaxx-google:
This commit also drops the patch that's not needed anymore, as it went
into the upstream.

Additionally, this commit initializes the global "Add_Interface_In_Name"
variable to 0, as it's being used by the upstream code in
hotplug_libusb.c since
https://github.com/LudovicRousseau/PCSC/commit/3e2106ebb9be1e030619ee70b2c8901e2f019e97


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #333, opened 2021-02-22
Allow dynamic memory growth in Emscripten modules

emaxx-google:
Allow the Emscripten module to use more memory than the default limit
(which currently is 16 MiB), by dynamically increasing the memory at
runtime when needed thanks to the "ALLOW_MEMORY_GROWTH=1" flag.

This functionality emits a link-time warning that it can slow down the
program due to it using Pthreads as well, but it should be of a minor
concern in our case (because it only seems to affect JS-to-C++
communication).

This commit also specifies the "ABORTING_MALLOC=1" flag, since it's
safer to fail fast and the default when using "ALLOW_MEMORY_GROWTH=1" is
the opposite.

This change contributes to the WebAssembly/Emscripten migration effort,
tracked by #177.

### DIFF
--- a/common/make/internal/executable_building_emscripten.mk
+++ b/common/make/internal/executable_building_emscripten.mk
@@ -85,6 +85,11 @@ EMSCRIPTEN_COMPILER_FLAGS := \
 # `EMSCRIPTEN_COMMON_FLAGS`.
 #
 # Explanation:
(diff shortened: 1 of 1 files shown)


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #334, opened 2021-02-22
Use g instead of g4 with Emscripten

emaxx-google:
Replace the "-g4" compiler/linker flag with "-g" when using
Emscripten. Apparently, "-g" started to produce more useful debug
information than "-g4" in recent Emscripten versions, in particular
regarding the recently added DWARF debug format support:
https://github.com/emscripten-core/emscripten/issues/13534

Note that the debug source maps are only produced at "g4", but we
never supported them properly in this project (in particular, due to
DevTools not being able to load them in Extensions/Apps - see
https://crbug.com/974543).

This contributes to the WebAssembly/Emscripten migration effort, as
tracked by #177.

### DIFF
--- a/common/make/internal/executable_building_emscripten.mk
+++ b/common/make/internal/executable_building_emscripten.mk
@@ -119,7 +119,7 @@ else ifeq ($(CONFIG),Debug)
 #
 # Explanation:
 # O0: Disable optimizations.
-# g3: Preserve debug information, including DWARF data.
+# g: Preserve debug information, including DWARF data.
 EMSCRIPTEN_COMMON_FLAGS += \
   -O0 \
   -g \
--- a/common/make/internal/executable_building_emscripten.mk
+++ b/common/make/internal/executable_building_emscripten.mk
@@ -119,10 +119,10 @@ else ifeq ($(CONFIG),Debug)
 #
 # Explanation:
 # O0: Disable optimizations.
-# g4: Preserve maximum debug information, including source maps.
+# g3: Preserve maximum debug information, including DWARF data.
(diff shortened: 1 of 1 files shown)


Fabian-Sommer (approved):
LGTM % nit in commit message, line 2: Replace "less" with "more"?

common/make/internal/executable_building_emscripten.mk
```diff
@@ -119,10 +119,10 @@ else ifeq ($(CONFIG),Debug)
 #
 # Explanation:
 # O0: Disable optimizations.
-# g4: Preserve maximum debug information, including source maps.
+# g3: Preserve maximum debug information, including DWARF data.
 EMSCRIPTEN_COMMON_FLAGS += \
   -O0 \
-  -g4 \
+  -g3 \
```
RReverser, line 125: I'd probably recommend to just use `-g` since that's the default equivalent to `-g3` and doesn't require looking up particular level in docs for the next person.

emaxx-google: Thanks for chiming in. Fair point, done.

common/make/internal/executable_building_emscripten.mk
```diff
@@ -119,10 +119,10 @@ else ifeq ($(CONFIG),Debug)
 #
 # Explanation:
 # O0: Disable optimizations.
-# g4: Preserve maximum debug information, including source maps.
+# g3: Preserve debug information, including DWARF data.
```
RReverser, line 122:
```suggestion
# g: Preserve debug information, including DWARF data.
```

leftover

ayaNader (approved):
LGTM

PR #335, opened 2021-02-26
Fix error message when Emscripten env is unset

emaxx-google:
Make the Makefile scripts display the correct error message when the
Emscripten build is requested without the required environment variables
being set.

The problem was a syntax error in the $(error) function invocation, as
the multi-line message was missing the backslash escape.

This change contributes to the WebAssembly migration effort tracked
by #177.


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #338, opened 2021-03-31
Shorten log prefix for incoming PC/SC commands

emaxx-google:
Replace the "PC/SC-Lite client handler for ..." text in log messages
with just "PC/SC from ...". This makes the log messages shorter a bit
while preserving all information, which helps readability.

by #146.


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #339, opened 2021-03-31
Remove relative time from JS logs

emaxx-google:
Stop printing relative times (like "[  0.007s]") in the JavaScript logs.
These times aren't really necessary now that we're printing absolute
times since #108.

The only remaining benefit of relative times was that they allowed
to distinguish logs from multiple instances in case they work
concurrently and their logs are squashed together. However, in
practice this was only relevant in a specific scenario (reading
Chrome OS system logs from a device that is locked and has the
NaCl-based smart card applications installed both in-session
and on the Login Screen). Assuming this a rare scenario that will
anyway disappear with the NaCl deprecation, it seems fine to
reduce the log clutter by removing the relative times.

by #146.

### DIFF
--- a/common/js/src/logging/log-formatting.js
+++ b/common/js/src/logging/log-formatting.js
@@ -34,6 +34,7 @@ const GSC = GoogleSmartCard;
  */
 const textFormatter = new goog.debug.TextFormatter();
 textFormatter.showAbsoluteTime = true;
+textFormatter.showRelativeTime = false;
 textFormatter.showSeverityLevel = true;
 
 /**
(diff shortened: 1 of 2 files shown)


Fabian-Sommer (approved):
LGTM

emaxx-google:
I actually remembered there was one reason why I kept these relative times originally. PTAL again - I've updated the commit description; I still think it's beneficial to get rid of these times.

Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #340, opened 2021-04-01
Remove prefix from the syslog log messages

emaxx-google:
Delete the "[syslog]" prefix printed in every log message that's created
via the syslog() call. This prefix is not really useful - the original
idea was to visually disambiguate the logs printed by C++ code from the
logs emitted by third-party C code that uses syslog. In reality, there's
not much ambiguity there, so this prefix was just cluttering the logs.

by #146.


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #341, opened 2021-04-01
Log function call indices

emaxx-google:
Print a function call index (a shared monotonically incrementing
counter) when logging function call tracer.

This should simplify investigating log messages collected in the field,
since it'll be possible to correlate the "called..." and the "returning"
log messages, therefore knowing which input parameters were passed to
the completed call and being able to distinguish between multiple calls
with the same parameters.

### DIFF
--- a/common/cpp/src/google_smart_card_common/logging/function_call_tracer.cc
+++ b/common/cpp/src/google_smart_card_common/logging/function_call_tracer.cc
@@ -17,7 +17,6 @@
 #include <stdint.h>
 
 #include <atomic>
-#include <cstdint>
 
 namespace google_smart_card {
 
--- a/common/cpp/src/google_smart_card_common/logging/function_call_tracer.cc
+++ b/common/cpp/src/google_smart_card_common/logging/function_call_tracer.cc
@@ -14,12 +14,25 @@
 
 #include <google_smart_card_common/logging/function_call_tracer.h>
 
+#include <atomic>
+#include <cstdint>
+
 namespace google_smart_card {
+
+namespace {
+
+int64_t GenerateFunctionCallId() {
+  static std::atomic<int64_t> next_function_call_id;
+  return ++next_function_call_id;
+}
+
+}  // namespace
 
 FunctionCallTracer::FunctionCallTracer(const std::string& function_name,
                                        const std::string& logging_prefix,
                                        LogSeverity log_severity)
-    : function_name_(function_name),
+    : function_call_id_(GenerateFunctionCallId()),
+      function_name_(function_name),
       logging_prefix_(logging_prefix),
       log_severity_(log_severity) {}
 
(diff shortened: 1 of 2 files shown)


common/cpp/src/google_smart_card_common/logging/function_call_tracer.cc
```diff
@@ -14,12 +14,26 @@
...
+
+#include <atomic>
+
 namespace google_smart_card {
 
+namespace {
+
+int64_t GenerateFunctionCallId() {
+  static std::atomic<int64_t> next_function_call_id;
+  return ++next_function_call_id;
```
Fabian-Sommer, line 27: Not familiar with std::atomic - but I would think that this does not quite guarantee uniqueness of ids, only that the behavior of races is well-defined. At least, this looks a lot like what https://en.cppreference.com/w/cpp/atomic/atomic/fetch_add seems to be intended for.

emaxx-google: Hmm, my understanding is that it is actually unique. This is a complex topic though...

Descriptions in the Standard (or at cppreference.com) are very cryptic, but a couple of points:
* https://en.cppreference.com/w/cpp/atomic/atomic/fetch_add says: "The operation is read-modify-write operation."
* It also says that the `memory_order_seq_cst` order is used by default.
* https://arxiv.org/pdf/1803.04432.pdf says: "<...> there is a single total order S of all sequentially consistent operations. An operation B that performs a load on an object M will observe the result
of the last modification A of M that precedes B in S <...>. From this it follows that there is always a happens-before relation between two memory_order_seq_cst operations operating on the same object."

My conclusion from all this is that two different `operator++()` calls are actually executed in the intuitively expected manner, that is, there'll never be the case when two concurrent calls read the same value. WDYT?

Fabian-Sommer: My thinking was that line 27 has a read, an addition, and a write, and that they are not atomic operations.
But I found that std::atomic has an overwrite for operator++(): https://en.cppreference.com/w/cpp/atomic/atomic/operator_arith 
So this is correct indeed.
It appears that both
++next_function_call_id;
and
next_function_call_id += 1;
are atomic, but
next_function_call_id = next_function_call_id + 1;
is not.

Fabian-Sommer (approved):
LGTM

common/cpp/src/google_smart_card_common/logging/function_call_tracer.cc
emaxx-google: Ah, yup, `operator++` is equivalent to `fetch_add`. I misinterpreted your first message - I thought you're concerned that even `fetch_add` can return duplicate values. Alright, resolving this thread then.

ayaNader (approved):
LGTM

PR #342, opened 2021-04-01
Double the log buffer capacity

emaxx-google:
Make the number of log messages retained for the Log Export to be twice
as much as before: in Release builds, it becomes 2000 messages now.
This should help in investigating some customer issues, because
apparently sometimes real-world logs contained information just around
the zone that was cut off.

This change is still a conservative change, because we need to take
memory consumption in mind. Extra 1000 messages added here would mean
100-200 KB of extra memory, which sounds tolerable.


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #344, opened 2021-04-06
Silence Emscripten Pthreads performance warning

emaxx-google:
Suppress the link-time warning emitted by Emscripten due to our usage of
multi-threading together with dynamic memory growth:

  warning: USE_PTHREADS + ALLOW_MEMORY_GROWTH may run non-wasm code
  slowly, see https://github.com/WebAssembly/design/issues/1271

We took this warning into consideration and decided to trade the
performance in favor of lower memory consumption. (In reality, the
performance loss is likely small, since a big chunk of the run time is
spent on interprocess communication.)

So reduce the build log spam by silencing this warning, which was
emitted many times - once for every target binary or static library.
This change is a small improvement to the WebAssembly build
support that is tracked by #177.


Fabian-Sommer (approved):
LGTM

PR #347, opened 2021-04-07
Suppress the "default browser" popups in test runs

emaxx-google:
This suppresses the first-run dialogs when running Chrome for executing
tests. In particular, this allows to skip the blocking popup "make
Chrome the default browser" that was introduced in recent Chrome
versions, and which required a manual mouse click to proceed to the
tests.

### DIFF
--- a/common/make/common.mk
+++ b/common/make/common.mk
@@ -232,8 +232,16 @@ CHROME_ENV ?=
 #
 # This is intended to be customized by the user in case of some special needs.
 #
+# Explanation of arguments:
+# no-first-run: Suppress first-run dialogs, like about making Chrome the default
+#   browser.
+# password-store: Disable using the system-wide encryption storage backend, to
+#   avoid notifications about system backend initialization issues.
+#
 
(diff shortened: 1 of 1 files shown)


common/make/common.mk
```diff
@@ -232,8 +232,16 @@ CHROME_ENV ?=
 #
 # This is intended to be customized by the user in case of some special needs.
 #
+# Explanation of arguments:
+# no-first-run: Suppress first-run dialogs, like about making Chrome the default
```
Fabian-Sommer, line 236: nits: "Suppresses", "like about" -> "such as"

common/make/common.mk
```diff
@@ -232,8 +232,16 @@ CHROME_ENV ?=
...
+#   browser.
+# password-store: Disable using the system-wide encryption storage backend, to
```
Fabian-Sommer, line 238: nit: "Disables"

Fabian-Sommer (approved):
LGTM % nits

ayaNader (approved):
LGTM

PR #349, opened 2021-04-16
Remove NaCl specifics from external_logs_printer

emaxx-google:
Refactor the common external_logs_printer.cc file to stop using
conversions to NaCl-specific structures, and switch to using the
toolchain-independent "Value" class.

Even though this external_logs_printer.cc file is compiled only under
NaCl, it's still beneficial to switch it away from the NaCl-specific
conversion routines, since it unblocks deleting the latter from the
project.

This refactoring contributes to the WebAssembly/Emscripten migration
effort tracked by #185.

### DIFF
--- a/common/cpp/src/google_smart_card_common/external_logs_printer.cc
+++ b/common/cpp/src/google_smart_card_common/external_logs_printer.cc
@@ -17,14 +17,10 @@
 #include <iostream>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #351, opened 2021-04-30
Strengthen int conversions to Emscripten val

emaxx-google:
Add the missing safety check that prevents a huge integer in a Value
object from being converted to a double Emscripten/JavaScript number
with a loss of precision.

This improvement is part of the WebAssembly migration effort, tracked
by #185.


Fabian-Sommer (approved):
LGTM

PR #352, opened 2021-05-01
Fix comment about non-portable Emscripten code

emaxx-google:
Correct the comment about the allegedly non-portable code in
global_context_impl_emscripten.cc that converts pointers to integers and
back. Apparently, it's a valid C++, so it's not a hack that only works
under Emscripten.

This is a pure cleanup commit; it contributes to the WebAssembly
migration effort tracked by #185.


Fabian-Sommer (approved):
LGTM

PR #353, opened 2021-06-07
Detect NaCl crash loops before Chrome throttles us

emaxx-google:
Improve the crash loop detector for the case when the crash happens in
the C/C++ code compiled under Native Client. Before this commit, the
crash loop detection would be triggered too late, after Chrome Native
Client system already began the throttling. This means that instead of
useful logs that contain information before the actual crash, the only
information would be like this:

>   Failed to load NaCl module with the following message: NaCl module
>   load failed: ServiceRuntime: failed to start

With the following information in Chrome logs:

>  NaCl process launch failed: Process creation was throttled due to
>  excessive crashes

This commit avoids this situation by tightening the crash loop detection
constants, so that the detector is triggered earlier and therefore
avoids this dead loop and preserves valuable logs.

P.S. Meanwhile this commit is applied to both NaCl and WebAssembly
builds, it's presumably OK, since all we do is tightening the crash loop
detection constants. If we discover later that WebAssembly contains a
similar, but even more aggressive throttling, we can tighten the
constants further; otherwise they don't harm.

### DIFF
--- a/common/js/src/logging/crash-loop-detection.js
+++ b/common/js/src/logging/crash-loop-detection.js
@@ -49,9 +49,14 @@ const CRASH_LOOP_WINDOW_MILLISECONDS = 60 * 1000;
 //    seconds after the launch":
 //    <https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/extensions/api/runtime/chrome_runtime_api_delegate.cc;l=172-173;drc=d17745f350d4956a07cb1113ee19e9cbc4be699f>.
 // 2. Chrome's Native Client system disables the module when it crashes too
-//    frequently, with the criteria currently being "3 crashes within 2
-//    minutes":
+//    frequently, with the criteria currently being "forbid subsequent load
+//    attempts after 3 crashes within 2 minutes":
 //    <https://source.chromium.org/chromium/chromium/src/+/main:components/nacl/browser/nacl_browser.cc;l=113-114;drc=d17745f350d4956a07cb1113ee19e9cbc4be699f>.
+//
+// We're choosing the largest possible value that satisfies these restrictions
+// in order to reduce the risk of false positives, i.e., 3. Note that despite
+// this value is exactly equal to restriction#2, it's still fine, since
+// throttling breaks subsequent load attempts, which is exactly what we avoid.
 const CRASH_LOOP_THRESHOLD_COUNT = 3;
 
 /** @type {boolean} */
--- a/common/js/src/logging/crash-loop-detection.js
+++ b/common/js/src/logging/crash-loop-detection.js
@@ -38,11 +38,19 @@ const CRASH_LOOP_WINDOW_MILLISECONDS = 60 * 1000;
 // crash loop.
 //
 // Note that |CRASH_LOOP_WINDOW_MILLISECONDS| and |CRASH_LOOP_THRESHOLD_COUNT|
-// have to be chosen in a way that the crash loop is detected earlier than the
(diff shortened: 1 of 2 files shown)


common/js/src/logging/crash-loop-detection.js
```diff
@@ -38,11 +38,21 @@ const CRASH_LOOP_WINDOW_MILLISECONDS = 60 * 1000;
...
+// have to be chosen in a way that the crash loop is detected earlier than
+// Chrome starts to throttle us, since that would leave the application in a
+// state without any useful logs.
+//
+// As of 2021-06-07, two subsystems in Chrome have to be taken into account as
+// follows:
+// 1. Chrome's Extension system disables the extension when it reloads too
+//    frequently, with the criteria currently being "5 reloads, each within 10
+//    seconds after the launch":
+//    <https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/extensions/api/runtime/chrome_runtime_api_delegate.cc;l=172-173;drc=d17745f350d4956a07cb1113ee19e9cbc4be699f>.
```
Fabian-Sommer, line 50: nit: As far as I can see, the drc parameter is not necessary in the URL

common/js/src/logging/crash-loop-detection.js
```diff
@@ -38,11 +38,21 @@ const CRASH_LOOP_WINDOW_MILLISECONDS = 60 * 1000;
...
+// 2. Chrome's Native Client system disables the module when it crashes too
+//    frequently, with the criteria currently being "3 crashes within 2
+//    minutes":
+//    <https://source.chromium.org/chromium/chromium/src/+/main:components/nacl/browser/nacl_browser.cc;l=113-114;drc=d17745f350d4956a07cb1113ee19e9cbc4be699f>.
```
Fabian-Sommer, line 54: nit: ditto

common/js/src/logging/crash-loop-detection.js
```diff
@@ -38,11 +38,21 @@ const CRASH_LOOP_WINDOW_MILLISECONDS = 60 * 1000;
...
+const CRASH_LOOP_THRESHOLD_COUNT = 3;
```
Fabian-Sommer, line 55: By setting this to 3, we would only detect the crash loop when the NaCl process is already throttled. I think this should be just 2.

Fabian-Sommer (changes requested):

common/js/src/logging/crash-loop-detection.js
emaxx-google: Ack - It's more future-proof with `drc`, since it specifies the commit hash. Without it, if the file gets moved later, the link would become dead.

common/js/src/logging/crash-loop-detection.js
emaxx-google: This is exactly on the boundary, but I believe should be correct. I've added more clarifications into the comment - PTAL.

Fabian-Sommer: I see. What confused me was that before the constant was 4, while the only known condition was 5.

Fabian-Sommer (approved):
LGTM

common/js/src/logging/crash-loop-detection.js
emaxx-google: Yeah, I previously used 5 as a no-brainer (maybe 6 would work there too). BTW, I also tested the "3" constant yesterday manually (by inducing an intentional crash loop in the NaCl module) and it worked as expected.

PR #356, opened 2021-06-10
Workaround for chrome.usb.getConfigurations issue

emaxx-google:
Change the USB wrapper (used in the webport of libusb, which is
transitively used for the CCID driver and therefore the Smart Card
Connector app) to stop asserting that chrome.usb.getConfigurations() can
only return a single active configuration. Instead, just proceed with
the first configuration among those that are marked as active.

Meanwhile multiple active configurations is most likely a bug in the
chrome.usb API (https://crbug.com/1218141), we need this workaround
until that issue is fixed in Chrome, since currently the Connector app
ends up in a crash loop when such device is attached.

### DIFF
--- a/third_party/libusb/webport/src/libusb_over_chrome_usb.cc
+++ b/third_party/libusb/webport/src/libusb_over_chrome_usb.cc
@@ -388,7 +388,7 @@ int LibusbOverChromeUsb::LibusbGetActiveConfigDescriptor(
   for (const auto& chrome_usb_config : chrome_usb_configs) {
     if (chrome_usb_config.active) {
       if (*config) {
-        // Normally there should be only one active coniguration, but the
+        // Normally there should be only one active configuration, but the
         // chrome.usb API implementation misbehaves on some devices: see
         // <https://crbug.com/1218141>. As a workaround, proceed with the first
         // received configuration that's marked as active.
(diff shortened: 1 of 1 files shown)


third_party/libusb/webport/src/libusb_over_chrome_usb.cc
```diff
@@ -387,9 +387,16 @@ int LibusbOverChromeUsb::LibusbGetActiveConfigDescriptor(
...
-      GOOGLE_SMART_CARD_CHECK(!*config);
+      if (*config) {
+        // Normally there should be only one active coniguration, but the
+        // chrome.usb API implementation misbehaves on some devices: see
+        // <https://crbug.com/1218141>. As a workaround, proceed with the first
+        // received configuration that's marked as active.
+        GOOGLE_SMART_CARD_LOG_WARNING
+            << "Unexpected result from the chrome.usb.getConfigurations() API: "
+               "received multiple active configurations";
+        continue;
```
emaxx-google, line 398: Actually, maybe we should do `break` here, since emitting multiple logs here is just a log spam. WDYT?

Fabian-Sommer: Agree with break - I don't think we are interested in just how many active configs there are.

Fabian-Sommer (approved):
LGTM

PR #358, opened 2021-06-11
Only pass MODULARIZE to Emscripten linker in tests

emaxx-google:
Fix the unit test build scripts to only pass the MODULARIZE flag to the
Emscripten linker, and stop passing it to the compiler. The latter
has been ignoring this flag anyway, and also newer Emscripten versions
started to emit errors when a useless flag like this is supplied.

This commit unblocks updating to a newer Emscripten version, and
contributes to the WebAssembly migration effort tracked by #185.


Fabian-Sommer (approved):
LGTM

PR #360, opened 2021-06-11
Fix env/initialize.sh Python 2 errors

emaxx-google:
Python 2 isn't maintained anymore, so more and more stuff stops working.
But as NaCl env build scripts are using it (and no one is interested in
migrating them to Pytnon 3), this commit fixes the issues that are
happening right now:

* pip should be fetched from a different URL (a special URL for a
  2.7-compatible version),
* some module called "astroid" should be installed with the version
  before 2.0, otherwise it fails with syntax errors.


ayaNader (approved):
LGTM

Fabian-Sommer (approved):
LGTM

Issue #264, opened 2020-11-11
Modernize JavaScript code


emaxx-google:
Currently, the JavaScript code in this repository follows pretty old guidelines (in some aspects, a few-years-old one). The code doesn't conform to the current Google JavaScript Style Guide (<https://google.github.io/styleguide/jsguide.html>) in several aspects, most notably:

- Instead of modules, the `goog.require`+`goog.provide` combination is used.
- Instead of `const`/`let`, `var` is used.
- Instead of JavaScript classes, functions with the `@constructor` annotations are used.

emaxx-google:
Besides just cleaning up the code, fixing this will also allow us to update to recent versions of Closure Compiler and Closure Library. Currently we're blocked, since the recent versions of Closure Compiler require the usage of JS modules.

PR #363, opened 2021-07-23
Add missing JS imports

emaxx-google:
Add missing goog.require() statements into the JS code where needed.
Older versions of Closure Compiler were somewhat relaxed in requiring
these imports, but newer do require all of them to be explicitly
require()'ed.

This contributes to the JS modernization effort tracked by #264.


ayaNader (approved):
LGTM

Fabian-Sommer (approved):
LGTM

PR #364, opened 2021-07-25
Use const/let instead of var in JS code

emaxx-google:
Modernize the JavaScript code by using "let"/"const" instead
of "var".

This is intended to be a non-functional change that contributes
to the effort tracked by #264.


emaxx-google:
Apologies for the large commit, but I thought it doesn't make much sense to split this mechanical changes into separate ones.

Unfortunately, I also accidentally made the diff even bigger by applying clang-tidy, as I couldn't easily "undo" the clang-tidy change when it was done in the same block of code as the one that I originally modified for this CL.

third_party/libusb/webport/src/chrome_usb/chrome-login-state-hook-unittest.js
```diff
@@ -118,104 +114,94 @@ function makeTest(fakeProfileType, fakeSessionState, testCallback) {
...
-      function(callback) { callback(fakeNewSessionState); });
+      chrome.loginState, 'getSessionState', function(callback) {
+        callback(fakeNewSessionState);
+      });
   goog.array.forEach(sessionStateListeners, function(listener) {
     listener(fakeNewSessionState);
   });
 }
 
+goog.exportSymbol('test_ChromeLoginStateHook_NoApi', function() {
```
Fabian-Sommer, line 125: Starting from here, this file looks weird to me. Maybe I'm just not parsing the git diff correctly, but I believe that with this patch there are two test_ChromeLoginStateHook_NoApi, but no test_ChromeLoginStateHook_DoesNotFilterInSession.

Fabian-Sommer (changes requested):
LGTM for all files except third_party/libusb/webport/src/chrome_usb/chrome-login-state-hook-unittest.js

third_party/libusb/webport/src/chrome_usb/chrome-login-state-hook-unittest.js
emaxx-google: Great catch! Seems like an artifact of an unlucky interactive git add :(
Now I've checked out the original version of this file and redid the const-related changes without any unrelated reformatting.

Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #368, opened 2021-07-27
Update Closure Compiler to 20210601

emaxx-google:
Roll our pinned version of the Closure Compiler to the latest stable
version. Do the needed adjustments in the compilation flags passed to
the compiler:
* the "STRICT"/"LOOSE" dependency_mode values got renamed to
  "PRUNE"/"PRUNE_LEGACY";
* some of warning classes for jscomp_error got renamed/deleted/added;
* temporarily disable the "lintChecks" warning class, since it started
  to require significant refactoring of our code base (e.g., to use
  modules instead of good.provide).

This commit unblocks the 1.5-year-old frozen roll; the unblocking became
possible thanks to some JS modernization work done in previous commits.


emaxx-google:
I only edited the following files:
* common/make/js_building_common.mk
* third_party/closure-compiler-binary/README.google
* third_party/closure-compiler/README.google

The rest is a verbatim copy from upstream (from the Closure Compiler) release, so shouldn't be reviewed here.

Fabian-Sommer (approved):
LGTM.
I wonder if there is a way to turn on all jscomp_errors, instead of listing all of them one by one.

emaxx-google:
> LGTM.
> I wonder if there is a way to turn on all jscomp_errors, instead of listing all of them one by one.

Thanks. It's possible to pass an asterisk, but unfortunately we need all but some of them (see the --jscomp_off parameters), and I haven't seen a way to achieve that.

emaxx-google:
> > > LGTM.
> > > I wonder if there is a way to turn on all jscomp_errors, instead of listing all of them one by one.
> > 
> > 
> > Thanks. It's possible to pass an asterisk, but unfortunately we need all but some of them (see the --jscomp_off parameters), and I haven't seen a way to achieve that.
> 
> Actually, let me try that. If --jscomp_off takes precedence, then we can turn on everything via an asterisk, and then suppress the unwanted ones by listing them.

It actually worked, thanks a lot for the idea!

On a side note, apparently, there are some undocumented warning classes which get triggered by the asterisk as well. I had to suppress some of them as they started to complain on our code (mostly reasonably!); later we'll need to clean that up.

Fabian-Sommer (approved):
LGTM

common/make/js_building_common.mk
```diff
@@ -64,86 +64,71 @@ $(eval $(call CLEAN_RULE,$(JS_BUILD_DIR_ROOT_PATH)))
...
+# jscomp_error *: Enable all warning classes and treat them as errors by
+#   default. Note: To be on the safe side, we're specifying all possible warning
+#   classes, except those that don't work for our codebase, which we explicitly
+#   list via jscomp_off.
+# jscomp_off extraRequire: Suppress warnings about unused goog.require()
+#   statements. TODO: Fix the violations and promote this to jscomp_error.
+# jscomp_off jsdocMissingConst: Suppress warnings about variables that could be
+#   marked as const. TODO: Fix the violations and promote this to jscomp_error.
+# jscomp_off lintChecks: Suppress complaints about code style. TODO: Fix the
+#   code (e.g., switch to modules instead of good.provide) and promote this to
```
ayaNader, line 93: typo: goog*

common/make/js_building_common.mk
```diff
@@ -64,86 +64,71 @@ $(eval $(call CLEAN_RULE,$(JS_BUILD_DIR_ROOT_PATH)))
...
 
 
-#
 # Closure Compiler compilation flags that should be additionally specified when
 # compiling JavaScript files with tests.
 #
-
+# Explanation:
+# dependency_mode=PRUNE_LEGACY: In addition to the standard PRUNE behavior (see
+#   above), also include all JS files that don't have a good.provide. This
```
ayaNader, line 123: typo: goog*

ayaNader (approved):
LGTM

common/make/js_building_common.mk
emaxx-google: Done, thanks for catching this! I'm making this typo all the time - probably my hidden excitement of Closure Compiler :)

PR #369, opened 2021-07-27
Replace usages of goog.isArray

emaxx-google:
Refactor JavaScript code to use Array.isArray() instead of
goog.isArray().

goog.isArray() got deprecated and removed from the recent
versions of the Closure Library.

This contributes to the refactoring effort tracked by #264.


Fabian-Sommer (approved):
LGTM

PR #370, opened 2021-07-27
Update Closure Library to 20200830

emaxx-google:
Note that we're picking a 1-year-old version, because newer versions of
Closure Library deprecate a bunch of stuff, and the recommended
replacements didn't exist in very old versions of the library. That's why
we're doing the update stepwise, like this: refactor => uprev =>
refactor => uprev => etc.


emaxx-google:
Hmm, something is wrong with this change - all unit tests get ignored with it. I'll investigate.

emaxx-google:
> Hmm, something is wrong with this change - all unit tests get ignored with it. I'll investigate.

Ah, no, they're still run, it's jut that the Closure Library stopped displaying their names in the UI. OK.

Fabian-Sommer (approved):

PR #371, opened 2021-07-27
Replace usages of goog.isFunction

emaxx-google:
Refactor JavaScript code to use goog.functions.isFunction()
instead of goog.isFunction().

goog.isFunction() got deprecated and removed from the recent
versions of the Closure Library; for whatever reason, they decided to
add a synonym "goog.functions.isFunction".

This change contributes to the refactoring effort tracked by #264.


Fabian-Sommer (approved):
LGTM

PR #374, opened 2021-07-28
Add missing @const

emaxx-google:
Refactor JavaScript code to specify "@const" for all class attributes
that are never reassigned. Also promote the "jsdocMissingConst" Closure
Compiler diagnostic, which complained about private properties that can
be marked as @const, to be an error.

This is intended to be a non-function change, contributing to the


ayaNader (approved):
LGTM

Fabian-Sommer (approved):
LGTM

PR #375, opened 2021-07-28
Add missing @private

emaxx-google:
Fix a couple of occasions in the JavaScript code where a private member
wasn't marked as such (via @private).

This is expected to be a non-function change, which contributes to the


Fabian-Sommer (approved):
LGTM

PR #376, opened 2021-07-29
Migrate DeferredProcessor to ES6 classes

emaxx-google:
Stop using the old-style Closure Compiler classes and switch to modern
ES6 classes. Code was *not* reformatted, in order to make the diff
readable; later there'll be a separate commit that'll apply clang-format.

A brief overview of the changes done here:
* Instead of declaring the class by marking a function with
  "@constructor", the "class" ES6 keyword is used.
* Instead of using goog.inherits() for declaring the inheritance, the
  "extends" ES6 keyword is used.
* Static properties are first defined as file-global constants and only
  then copied into the class; this allows using them in the class code,
  since the old-style "this.someConstant" access doesn't work anymore.
  (A side note: ES6 class static properties aren't supported by Closure
  Compiler yet.)
* Calls to the parent class' methods are done via super() instead of the
  Closure Compiler's base() method.
* Nested classes are moved outside to be separate classes, since
  nesting isn't supported in ES6 classes.

This is expected to be a non-functional change. It contributes to the

### DIFF
--- a/common/js/src/deferred-processor.js
+++ b/common/js/src/deferred-processor.js
@@ -38,6 +38,26 @@ const suppressUnhandledRejectionError =
     GSC.PromiseHelpers.suppressUnhandledRejectionError;
 
 /**
+ * This structure is used to store the jobs in a queue.
+ * @struct
+ */
+class Job {
+/**
+ * @param {function()} jobFunction The job function that needs to be executed.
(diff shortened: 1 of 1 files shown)


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #378, opened 2021-09-08
Smartcard migration

ayaNader:
This PR is the UI migration of the Smart Card Connector from apps to extensions.
Changes:
1. the main window (being displayed as an extension rather than a separate app window)
2. the about and help links work in the new extension
3. popup window is shown using the new extension-related APIs
Many parts of the code have been commented out due to irrelevancy or due to causing errors, these parts will either be removed completely or replaced in future PRs.

### DIFF
--- a/common/js/src/popup-window/client.js
+++ b/common/js/src/popup-window/client.js
@@ -28,6 +28,7 @@ goog.provide('GoogleSmartCard.PopupWindow.Client');
 
 goog.require('GoogleSmartCard.DebugDump');
 goog.require('GoogleSmartCard.Logging');
+goog.require('GoogleSmartCard.Packaging');
 goog.require('goog.events');
 goog.require('goog.events.EventType');
 goog.require('goog.events.KeyCodes');
@@ -50,14 +51,13 @@ GSC.PopupWindow.Client.getData = function() {
   return goog.object.get(window, 'passedData', {});
 };
 
-//#revisitcode
 /**
- * Shows the window.
+ * Shows the app window.
  */
-// GSC.PopupWindow.Client.showWindow = function() {
-//   logger.fine('Showing the window...');
-//   chrome.app.window.current().show();
-// };
+GSC.PopupWindow.Client.showWindow = function() {
+  goog.log.fine(logger, 'Showing the window...');
+  chrome.app.window.current().show();
+};
 
 /**
  * Resolves the modal dialog, passing the specified result to the caller in the
@@ -100,7 +100,8 @@ GSC.PopupWindow.Client.prepareAndShowAsModalDialog = function() {
   GSC.PopupWindow.Client.setWindowHeightToFitContent();
   GSC.PopupWindow.Client.setupClosingOnEscape();
   GSC.PopupWindow.Client.setupRejectionOnWindowClose();
-  // GSC.PopupWindow.Client.showWindow();
+  if (GSC.Packaging.MODE == GSC.Packaging.Mode.APP)
+    GSC.PopupWindow.Client.showWindow();
 };
 
 /**
--- a/common/js/src/popup-window/client.js
+++ b/common/js/src/popup-window/client.js
@@ -50,13 +50,14 @@ GSC.PopupWindow.Client.getData = function() {
   return goog.object.get(window, 'passedData', {});
 };
 
+//#revisitcode
 /**
  * Shows the window.
  */
-GSC.PopupWindow.Client.showWindow = function() {
-  goog.log.fine(logger, 'Showing the window...');
-  chrome.app.window.current().show();
-};
+// GSC.PopupWindow.Client.showWindow = function() {
+//   logger.fine('Showing the window...');
+//   chrome.app.window.current().show();
+// };
 
 /**
  * Resolves the modal dialog, passing the specified result to the caller in the
@@ -99,7 +100,7 @@ GSC.PopupWindow.Client.prepareAndShowAsModalDialog = function() {
   GSC.PopupWindow.Client.setWindowHeightToFitContent();
   GSC.PopupWindow.Client.setupClosingOnEscape();
   GSC.PopupWindow.Client.setupRejectionOnWindowClose();
-  GSC.PopupWindow.Client.showWindow();
+  // GSC.PopupWindow.Client.showWindow();
 };
 
 /**
--- a/example_cpp_smart_card_client_app/src/window.js
+++ b/example_cpp_smart_card_client_app/src/window.js
@@ -24,6 +24,7 @@ goog.provide('SmartCardClientApp.WindowMain');
 goog.require('GoogleSmartCard.I18n');
 goog.require('GoogleSmartCard.Logging');
 goog.require('GoogleSmartCard.ObjectHelpers');
+goog.require('GoogleSmartCard.Packaging');
 goog.require('GoogleSmartCard.PopupWindow.Client');
 goog.require('goog.dom');
 goog.require('goog.events');
@@ -46,13 +47,12 @@ const logger = GSC.Logging.getLogger(
 
 goog.log.info(logger, 'The main window is created');
 
-//#revisitcode
 // Obtain the message channel that is used for communication with the executable
 // module.
-// const executableModuleMessageChannel =
-//     /** @type {!goog.messaging.MessageChannel} */
-//     (GSC.ObjectHelpers.extractKey(
-//          GSC.PopupWindow.Client.getData(), 'executableModuleMessageChannel'));
+const executableModuleMessageChannel =
+    /** @type {!goog.messaging.MessageChannel} */
+    (GSC.ObjectHelpers.extractKey(
+        GSC.PopupWindow.Client.getData(), 'executableModuleMessageChannel'));
 
 // Load the localized strings into the HTML elements.
 GSC.I18n.adjustAllElementsTranslation();
@@ -60,15 +60,16 @@ GSC.I18n.adjustAllElementsTranslation();
 // Called when the "close" button is clicked. Closes the window.
 function onCloseWindowClicked(e) {
   e.preventDefault();
-  // chrome.app.window.current().close();
+  if (GSC.Packaging.MODE == GSC.Packaging.Mode.APP)
+    chrome.app.window.current().close();
 }
 
 // Called when the "run test" button is clicked. Sends a command to the
 // executable module to start the test.
-// function onRunTestClicked(e) {
-//   e.preventDefault();
-//   executableModuleMessageChannel.send('ui_backend', {command: 'run_test'});
-// }
+function onRunTestClicked(e) {
+  e.preventDefault();
+  executableModuleMessageChannel.send('ui_backend', {command: 'run_test'});
+}
 
 // Called when a "output_message" message is received from the executable
 // module. Adds a line into the "output" <pre> element in the UI.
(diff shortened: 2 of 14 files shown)


common/make/app_building_common.mk
```diff
@@ -41,6 +41,7 @@ APP_RUN_CHROME_ARGS := \
 	--no-first-run \
 	--user-data-dir=$(APP_RUN_USER_DATA_DIR_PATH) \
 	--load-and-launch-app=$(OUT_DIR_PATH) \
+	--whitelisted-extension-id=khpfeaanjngmcnplbdlpegiifgpfgdco \
```
emaxx-google, line 44: This is a temporary debugging change and shouldn't belong to the PR.

common/js/src/popup-window/client.js
```diff
@@ -99,7 +100,7 @@ GSC.PopupWindow.Client.prepareAndShowAsModalDialog = function() {
   GSC.PopupWindow.Client.setWindowHeightToFitContent();
   GSC.PopupWindow.Client.setupClosingOnEscape();
   GSC.PopupWindow.Client.setupRejectionOnWindowClose();
-  GSC.PopupWindow.Client.showWindow();
+  // GSC.PopupWindow.Client.showWindow();
```
emaxx-google, line 103: There's a lot of code commented out in the PR, but I'm not sure it's the right approach to land it like that. Based on the previous performance, it'll probably take multiple weeks before everything gets fixed up. Also, based on discussion with Marcus yesterday, the rollout will be tricky - most likely we won't be able to roll out the extension in the next few months until sufficient number of users update to newer Chrome OS versions.

To sum up, my proposal would be to restructure the PR so that it doesn't comment out big blocks of code. Moreover, I'm proposing that we introduce two build modes: one for the App and another for the Extension, so that we still have a working App version at any point of time. Let me create a PR that would add build-time constants that we can use in the code to guard app- and extension- specific code.

example_cpp_smart_card_client_app/src/window.js
```diff
@@ -46,28 +46,29 @@ const logger = GSC.Logging.getLogger(
 
 goog.log.info(logger, 'The main window is created');
 
+//#revisitcode
```
emaxx-google, line 49: Ditto for not landing with these temporary comments. Let's keep the codebase clean at any moment of time.

smart_card_connector_app/src/manifest.json.template
```diff
@@ -1,16 +1,20 @@
...
   "description": "__MSG_appDesc__",
-  "version": "1.3.5.0",
-  "app": {
-    "background": {
-      "persistent": false,
-      "scripts": [
-        "background.js"
-      ]
-    }
+  "version": "1.3.4.0",
```
emaxx-google, line 7: The version shouldn't be decreased - just keep 1.3.5.0 that we have on master.

smart_card_connector_app/src/manifest.json.template
```diff
@@ -1,16 +1,20 @@
 {
+  "key": "MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAlc6hy5++uljMusXcpXspmA3UH5t9ppVHj3FfxvVMPOFX0RLrDvXtfH3f/TkdyPsAzk1QDDQ1++lDatRa3xc4rOGa2ynXljOR9y6rWoqB5wwMOjB42AWSVZWNwpccRsYkMhML5s7uur+B4SERWeMasXuJUhTPweS48Mz0mvp9wZMQB9+xPNm93iUeUZvVtB63ksibx5B3UJUCBz1wSUXLnEoTjy6TR9XsQ+boZlRsk/jY79A+iMQZw72QBBXvWSfJRaeDj266Dz+Jyp0h4lxZJjg/pnL0Rp5s0zlhlSNm6zyrQ6lWMzagjP/2G95LWMNvLTE95oKqhP4hNaxYrI/pUQIDAQAB",
```
emaxx-google, line 2: Ditto for a temporary debugging change that shouldn't end up landed.

smart_card_connector_app/src/window-about-showing.js
```diff
@@ -46,7 +46,7 @@ const openAboutElement =
 function openAboutClickListener(e) {
   e.preventDefault();
 
-  chrome.app.window.create(ABOUT_WINDOW_URL, ABOUT_WINDOW_OPTIONS);
+  chrome.windows.create({url: ABOUT_WINDOW_URL, type: "popup", width: 700, height: 500});
```
emaxx-google, line 49: nit: This line is longer than 80 characters, which is not allowed by the style guide - please reformat it (you can just run the "format-code.sh" script I added recently).

emaxx-google, line 49: nit: Let's introduce constants for "700" and "500", so that these magic numbers aren't scattered through the code, but are rather near the top of the file where they're easily discoverable.

smart_card_connector_app/src/window-apps-displaying.js
```diff
@@ -43,11 +43,14 @@ const KnownApp = PermissionsChecking.KnownApp;
...
-
-const knownAppsRegistry = new PermissionsChecking.KnownAppsRegistry();
+//#revisitcode
+/**
+ * @type {!Element}
+ * @const
+ */
+// var appListElement = /** @type {!Element} */ (goog.dom.getElement('app-list'));
+//revisitcode
+// const knownAppsRegistry = new PermissionsChecking.KnownAppsRegistry();
```
emaxx-google, line 53: Hmm, do you remember why this had to be commented out? That's strange if that class doesn't work in extensions, but I might be missing something.

smart_card_connector_app/src/window-main.js
```diff
@@ -45,19 +45,21 @@ const logger = GSC.Logging.getScopedLogger('ConnectorApp.MainWindow');
...
-GSC.ConnectorApp.Window.DevicesDisplaying.initialize();
+// GSC.ConnectorApp.Window.AppsDisplaying.initialize();
+// GSC.ConnectorApp.Window.DevicesDisplaying.initialize();
 GSC.ConnectorApp.Window.HelpShowing.initialize();
 GSC.ConnectorApp.Window.LogsExporting.initialize();
 
 GSC.I18n.adjustAllElementsTranslation();
 
-GSC.PopupWindow.Client.showWindow();
+// GSC.PopupWindow.Client.showWindow();
```
emaxx-google, line 62: BTW, now that we don't explicitly make the window visible and it's shown from the very beginning, does this introduce visible flickering? I.e., are you able to see the window being initially opened empty, and in the next moment populated with data?

If that's happening indeed, that'd be unfortunate. I don't have an immediate proposal about how to fix it though.

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker.js
```diff
@@ -278,39 +278,36 @@ UserPromptingChecker.prototype.promptUserForUnknownApp_ = function(
...
-        promiseResolver.reject(new Error(
-            'Rejected permission because of the user cancellation of the prompt ' +
-            'dialog'));
-      },
-      this);
+  
+  chrome.windows.create({url: USER_PROMPT_DIALOG_URL, width: 300, height: 180, type: "popup"});
+
+  // #toclean
+  // var dialogPromise = GSC.PopupWindow.Server.runModalDialog(
```
emaxx-google, line 285: Now that we know that we're going to fix PopupWindow.Server to work in extensions too, is there really a need in commenting out all of that?

Maybe we should start from a PR that would start this update of PopupWindow.Server?

emaxx-google (commented):
Thanks.

smart_card_connector_app/src/window-main.js
ayaNader: Nope, no flickering. I believe this function was for the app, but as an extension the window is opened right away when clicking on the icon

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker.js
ayaNader: I thought I'd start with the UI migration and then the backend part, WDYT?

emaxx-google: Sure, but we don't have to break the code base on Github while doing that. It's a common principle in the software development: try to maintain the central repository in a good shape even in the middle of complex multi-commit changes. For example, see Abseil's "Live at Head" philosophy: https://abseil.io/about/philosophy#upgrade-support

example_cpp_smart_card_client_app/src/window.js
```diff
@@ -59,7 +60,8 @@ GSC.I18n.adjustAllElementsTranslation();
 // Called when the "close" button is clicked. Closes the window.
 function onCloseWindowClicked(e) {
   e.preventDefault();
-  chrome.app.window.current().close();
+  if (GSC.Packaging.MODE == GSC.Packaging.Mode.APP)
```
emaxx-google, line 63: Perhaps this is not needed, if we don't show the close button in the extension mode at all?

smart_card_connector_app/src/window-about-showing.js
```diff
@@ -46,7 +46,16 @@ const openAboutElement =
...
   e.preventDefault();
 
-  chrome.app.window.create(ABOUT_WINDOW_URL, ABOUT_WINDOW_OPTIONS);
+  if (GSC.Packaging.MODE == GSC.Packaging.Mode.APP) {
+    chrome.app.window.create(ABOUT_WINDOW_URL, ABOUT_WINDOW_OPTIONS);
+  } else if (GSC.Packaging.MODE == GSC.Packaging.Mode.EXTENSION) {
+    chrome.windows.create({
+      url: ABOUT_WINDOW_URL,
+      type: 'popup',
+      width: ABOUT_WINDOW_OPTIONS['innerBounds']['width'],
```
emaxx-google, line 55: This looks weird - let's just create constants ABOUT_WINDOW_WIDTH/ABOUT_WINDOW_HEIGHT instead of ABOUT_WINDOW_OPTIONS.

smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -202,9 +203,15 @@ GSC.ConnectorApp.Window.DevicesDisplaying.initialize = function() {
...
   // Stop tracking the current list of readers when our window gets closed.
-  chrome.app.window.current().onClosed.addListener(function() {
-    readerTrackerUnsubscriber(onReadersChanged);
-  });
+  if (GSC.Packaging.MODE == GSC.Packaging.Mode.APP) {
+    chrome.app.window.current().onClosed.addListener(function() {
+      readerTrackerUnsubscriber(onReadersChanged);
+    });
+  } else if (GSC.Packaging.MODE == GSC.Packaging.Mode.EXTENSION) {
+    chrome.windows.onRemoved.addListener(function() {
```
emaxx-google, line 211: The documentation of chrome.windows.onRemoved is a bit unclear... Is it actually triggered when our window (which is the browser action popup) closed?

smart_card_connector_app/src/window-main.js
```diff
@@ -45,19 +46,22 @@ const logger = GSC.Logging.getScopedLogger('ConnectorApp.MainWindow');
...
+if (GSC.Packaging.MODE == GSC.Packaging.Mode.APP) {
+  goog.events.listen(
+      goog.dom.getElement('close-window'), goog.events.EventType.CLICK,
+      closeWindowClickListener);
+}
 
 GSC.ConnectorApp.Window.AboutShowing.initialize();
-GSC.ConnectorApp.Window.AppsDisplaying.initialize();
-GSC.ConnectorApp.Window.DevicesDisplaying.initialize();
+// GSC.ConnectorApp.Window.AppsDisplaying.initialize();
```
emaxx-google, line 56: These shouldn't be commented-out either.

smart_card_connector_app/src/window.css
```diff
@@ -30,6 +30,12 @@ button {
   font-weight: 400;
 }
 
+html,
+body {
+  width: 530px;
```
emaxx-google, line 35: BTW, does this affect the window's appearance in the App mode? Ideally we should avoid that.

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker.js
```diff
@@ -278,6 +279,17 @@ UserPromptingChecker.prototype.promptUserForUnknownApp_ = function(
  */
 UserPromptingChecker.prototype.runPromptDialog_ = function(
     clientAppId, userPromptDialogData, promiseResolver) {
+  if (GSC.Packaging.MODE == GSC.Packaging.Mode.EXTENSION) {
+    chrome.windows.create({
```
emaxx-google, line 283: This code should live in GSC.PopupWindow.Server, not here. Here we just want to show the dialog without taking care of such low-level details.

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker.js
```diff
@@ -278,6 +279,17 @@ UserPromptingChecker.prototype.promptUserForUnknownApp_ = function(
...
+      url: USER_PROMPT_DIALOG_URL,
+      width:
+          USER_PROMPT_DIALOG_WINDOW_OPTIONS_OVERRIDES['innerBounds']['width'],
```
emaxx-google, line 286: Ditto width/height constants.

emaxx-google:
Is this PR still relevant, or were all of its parts landed in separate PRs?

ayaNader:
> Is this PR still relevant, or were all of its parts landed in separate PRs?

Yeah all were landed in separate PRs

PR #381, opened 2021-09-10
Enable launch functionality only in the app mode

emaxx-google:
Add the chrome.app.runtime.onLaunched event listener only in the app
mode. Skip that in the extension mode, since launching is an
app-specific concept, and browser actions are used for extensions
instead.


emaxx-google:
PTAL - This is an example of how to use the new variable in the JS code, so that we can avoid breaking the app mode while developing the extension mode.

ayaNader (approved):
LGTM

Issue #379, opened 2021-09-10
Add "extension" build mode


emaxx-google:
Add build-time configuration option to switch between the "Chrome App" and "Extension" build options.

This will allow us to start migrating from Apps to Extensions without breaking ability to compile and publish new versions of the App.

PR #382, opened 2021-09-13
Rename {app_building => packaging}_common.mk

emaxx-google:
Rename the common/make/app_building_common.mk file to
packaging_common.mk, since we're going to add support for other
packaging modes besides the (legacy) Chrome App.

task #379.


Fabian-Sommer (approved):
LGTM

Issue #383, opened 2021-09-13
Fix fatal errors in the "extension" mode of the Smart Card Connector


emaxx-google:
This issue tracks a specific milestone of the app-to-extension migration of the Smart Card Connector, namely making it launchable in the PACKAGING=extension mode without fatal errors.

This specific milestone is *not* aiming to make a fully functioning version.

PR #384, opened 2021-09-13
Avoid exceptions on missing USB API

emaxx-google:
Improve the ChromeUsbBackend JavaScript code to avoid throwing
exceptions when the USB API is actually unavailable.

This should never happen in production, but it does happen in the
current (early) "extension" version of the Smart Card Connector.
Also extra safety in the code won't harm anyway.

task #383.


Fabian-Sommer (approved):
LGTM

PR #385, opened 2021-09-16
app/extension conditions in manifest.json.template

emaxx-google:
Add the support of conditionals into the Connector's
manifest.json.template, allowing to switch between extension- and
app-specific blocks in this file depending on the PACKAGING build
variable.

Do this by introducing a simple "${if ...} ... ${endif}" syntax into
create_app_manifest.py, with the regexp-based implementation. The
implementation is very basic and, e.g., doesn't support nested
conditions.

task #379.


Fabian-Sommer (approved):
LGTM

emaxx-google:
Thank you the quick review!

PR #386, opened 2021-09-16
Differentiate app-/extension-specific permissions

emaxx-google:
Change the Connector's manifest.json.template to only request the
App-specific permissions in the "app" packaging mode. This allows to
avoid warnings in Chrome when loading the "extension" version.

task #383.


Fabian-Sommer (approved):
LGTM

PR #387, opened 2021-09-16
Fix data passing from background to popup window

emaxx-google:
Make the data passing (connected readers and connected client
applications) between the Connector's background page and the popup
window work in both "app" and "extension" packaging modes.

This is achieved by changing the mechanism for passing data: instead of
manipulating the popup's global variables (in the
chrome.app.window.create() callback - which was obviously an
App-specific mechanism), let the popup read them from the background
page's globals.

task #383.


Fabian-Sommer (approved):
LGTM

PR #389, opened 2021-09-16
Add extension browser action for the Connector

ayaNader:
This adds a condition in the manifest.json.template which includes the browser_action property in case the PACKAGING build
variable is "extension".

task #379.

### DIFF
--- a/smart_card_connector_app/src/manifest.json.template
+++ b/smart_card_connector_app/src/manifest.json.template
@@ -16,6 +16,13 @@ ${if PACKAGING=app}
     }
 ${endif}
   },
+${if PACKAGING=extension}
+  "browser_action": {
+    "default_icon": "icon.png",
+    "default_title": "Smart Card Connector",
+    "default_popup": "window.html"
+  },
+${endif}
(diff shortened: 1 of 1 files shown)


smart_card_connector_app/src/manifest.json.template
```diff
@@ -16,6 +16,13 @@ ${if PACKAGING=app}
     }
 ${endif}
   },
+${if PACKAGING=extension}
+  "browser_action": {
+    "default_icon": "icon.png",
+    "default_title": "Smart Card Connector",
```
emaxx-google, line 22: This is a hardcoded text in English, but ideally we should target i18n from day one.

Would `__MSG_appShortName__` work here? At least it should, according to this doc: https://developer.chrome.com/docs/extensions/reference/browserAction/#tooltip

emaxx-google:
Also let's rename the PR to reflect what we're trying to achieve, as opposed to how. E.g., "Add extension browser action for the Connector".

emaxx-google (approved):

PR #391, opened 2021-09-17
Fix log collection in Connector's extension mode

emaxx-google:
This fixes the log collection and exporting functionality in the
"extension" packaging mode of the Smart Card Connector. The goal is to
collect the logs from all windows and pages in the background page, and
also to allow exporting them when the user clicks "Export logs" in the
UI.

Implementation-wise, this is achieved by switching to a different
mechanism for communicating between the background page and other
windows. The old app-specific mechanism worked like this: the background
page holds a log buffer that it injects into every opened window by
setting a global property on it. The new mechanism used in this commit
is the inverse of this: every window calls
chrome.runtime.getBackgroundPage() and grabs the log buffer.

Some additional complexity in this change is caused by the fact that
getBackgroundPage() is asynchronous, so that we need to preserve and
then move all logs we've accumulated when it completes. Also we need to
be careful to register the log buffer in the window's logger, which
means using the window's Closure Library instance.

task #390.

### DIFF
--- a/common/js/src/logging/log-buffer.js
+++ b/common/js/src/logging/log-buffer.js
@@ -85,6 +85,36 @@ goog.inherits(LogBuffer, goog.Disposable);
 goog.exportSymbol('GoogleSmartCard.LogBuffer', LogBuffer);
 
 /**
+ * Attaches the given log buffer to the given logger, by letting the buffer
+ * collect all logs sent to the logger. Additionally, all collected logs are
+ * attributed with the specified document location, which simplifies
+ * interpreting the collected logs.
+ * @param {!goog.log.Logger} logger
+ * @param {string} documentLocation
+ */
+LogBuffer.attachBufferToLogger = function(logBuffer, logger, documentLocation) {
+  // Note: It's crucial that we're in a static method and calling a global
+  // function (`goog.log.addHandle()`), because when `logBuffer` comes from a
+  // different page (e.g., the background page), we're dealing with two
+  // instances of Closure Library: one in our page and another in the buffer's
+  // page. We want the current page's instance to register the buffer, since
+  // it's where the `logger`s messages are handled.
+  goog.log.addHandler(
+      logger, logBuffer.onLogRecordObserved_.bind(logBuffer, documentLocation));
+};
+
+goog.exportProperty(
+    LogBuffer, 'attachBufferToLogger', LogBuffer.attachBufferToLogger);
+
+/** @override */
+LogBuffer.prototype.disposeInternal = function() {
+  this.formattedLogsPrefix_.length = 0;
+  this.observers_.length = 0;
+  this.formattedLogsSuffix_.length = 0;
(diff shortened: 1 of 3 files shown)


common/js/src/logging/log-buffer.js
```diff
@@ -85,26 +85,44 @@ goog.inherits(LogBuffer, goog.Disposable);
...
 
 goog.exportProperty(
-    LogBuffer.prototype, 'getCapacity', LogBuffer.prototype.getCapacity);
+    LogBuffer, 'attachBufferToLogger', LogBuffer.attachBufferToLogger);
+
+/** @override */
+LogBuffer.prototype.disposeInternal = function() {
+  this.formattedLogsPrefix_.length = 0;
+  this.observers_.length = 0;
+  this.formattedLogsSuffix_.length = 0;
```
emaxx-google, line 113: Actually this should've been `clear()`. Not sure why the compiler didn't complain about writing to a non-existing property in `CircularBuffer`.

common/js/src/logging/logging.js
```diff
@@ -314,24 +317,45 @@ function setupRootLoggerLevel() {
...
+    logBuffer.copyToOtherBuffer(backgroundLogBuffer);
+    // From now, start using the background page's buffer for collecting data
+    // from our page's loggers. Dispose of our log buffer to avoid storing new
+    // logs twice.
+    GSC.LogBuffer.attachBufferToLogger(
+        backgroundLogBuffer, rootLogger, document.location.href);
+    logBuffer.dispose();
+    // Switch all references to the background page's log buffer.
+    logBuffer = backgroundLogBuffer;
+    goog.global[GLOBAL_LOG_BUFFER_VARIABLE_NAME] = logBuffer;
```
Fabian-Sommer, line 357: Is this ever accessed if we are not the background page? If yes, why not use getLogBuffer() instead? If not, can we just set this to null?

common/js/src/logging/log-buffer.js
```diff
@@ -85,26 +85,44 @@ goog.inherits(LogBuffer, goog.Disposable);
 goog.exportSymbol('GoogleSmartCard.LogBuffer', LogBuffer);
 
 /**
- * @return {number}
+ * Attaches the given log buffer to the given logger, by letting the buffer
+ * collect all logs sent to the logger. Additionally, all collected logs are
+ * attributed with the specified document location, which simplifies
+ * interpreting the collected logs.
+ * @param {!goog.log.Logger} logger
```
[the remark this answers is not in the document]
emaxx-google: Done. I'm surprised the compiler didn't complain about that...

common/js/src/logging/logging.js
emaxx-google: Done - removed this property, as it's not needed indeed.

emaxx-google (commented):
Thanks!

PR #395, opened 2021-09-20
Fix ArrayBuffer bug in single-message-based comm

### LINKED ISSUE #393: SCardGetStatusChange returns wrong data in message JSON
Since version 1.3.5.0 (the bug does not occur in 1.3.3.0) we are getting a wrong behavior on the JSON received from the Smart Card Connector, regarding the response of a `SCardGetStatusChange` request.

The `atr` attribute should be a list, either empty (`[]`) when no card is inserted, or a list with the bytes from the inserted card's ATR. This was the case until 1.3.3.0, but since 1.3.5.0, the `atr` attribute is always an empty object (`{}`), which causes problems from that point on. 

We send the connector a `SCardGetStatusChange` request:

```
{
    "function_name": "SCardGetStatusChange",
    "arguments": [
        175319793,
        0,
        [
            {
                "reader_name": "Gemalto PC Twin Reader 00 00",
                "current_state": 0
            }
        ]
    ]
}
```
...and get back this response:
```
{
    "type": "pcsc_lite_function_call::response",
    "data": {
        "request_id": 2,
        "payload": [
            0,
            [
                {
                    "atr": {},
                    "current_state": 0,
                    "event_state": 34,
                    "reader_name": "Gemalto PC Twin Reader 00 00"
                }
            ]
        ]
    }
}
```
Notice the empty object (`{}`) returned as the `atr` attribute.
The log file contains the correct reply from the smart card, showing that the problem is after that point, in the data which is sent as a Google Chrome message to the extension:
```
[210918 20:08:54.61] [NaclModule] [INFO] [PC/SC from "ckhdehpbimdohoflpgoohhkffeipofmp" (id 1)] SCardGetStatusChange#8(hContext=0x0A732AF1, dwTimeout=0, rgReaderStates=0xFEB368B0[{szReader="Gemalto PC Twin Reader 00 00", pvUserData=NULL, dwCurrentState=SCARD_STATE_UNAWARE}]): called...

[210918 20:08:54.61] [NaclModule] [INFO] [PC/SC from "ckhdehpbimdohoflpgoohhkffeipofmp" (id 1)] SCardGetStatusChange#8: returning 0x00000000 ["Command successful."], rgReaderStates=0xFEB368B0[{szReader="Gemalto PC Twin Reader 00 00", pvUserData=NULL, dwCurrentState=SCARD_STATE_UNAWARE, dwEventState=SCARD_STATE_CHANGED|SCARD_STATE_PRESENT, cbAtr=25, rgbAtr=<0x3B 0xFF 0x96 0x00 0x00 0x81 0x31 0xFE 0x43 0x80 0x31 0x80 0x65 0xB0 0x85 0x04 0x01 0x20 0x12 0x0F 0xFF 0x82 0x90 0x00 0xD0>}]
```
But the `rbgAtr` values are not received in the message from the Smart Card Connector.
The correct response should be (taken from 1.3.3.0):
```
{
    "type": "pcsc_lite_function_call::response",
    "data": {
        "request_id": 1,
        "payload": [
            0,
            [
                {
                    "atr": [
                        59,
                        255,
                        150,
                        0,
                        0,
                        129,
                        49,
                        254,
                        67,
                        128,
                        49,
                        128,
                        101,
                        176,
                        133,
                        4,
                        1,
                        32,
                        18,
                        15,
                        255,
                        130,
                        144,
                        0,
                        208
                    ],
                    "current_state": 0,
                    "event_state": 196642,
                    "reader_name": "Gemalto PC Twin Reader 00 00"
                }
            ]
        ]
    }
}
```
Or, with no smart card inserted:
```
{
    "type": "pcsc_lite_function_call::response",
    "data": {
        "request_id": 1,
        "payload": [
            0,
            [
                {
                    "atr": [],
                    "current_state": 0,
                    "event_state": 131090,
                    "reader_name": "Gemalto PC Twin Reader 00 00"
                }
            ]
        ]
    }
}
```
Notice the empty list (`[]`) and not empty object in the `atr`attribute.
I'm including a text file with all `SCardEstablishContext`, `SCardListReaders` and `SCardGetStatusChange`, for all cases (version 1.3.3.0, with or without card -- working OK, and version 1.3.5.0, with or without card -- not working OK), including JSON requests and responses, as well as corresponding log messages from the Smart Card Connector.

Thanks

[chromeos-smartcardconnector-bug.txt](https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/files/7190546/chromeos-smartcardconnector-bug.txt)
!

emaxx-google:
Fix the issue with incorrect serialization of byte fields when using
single-message-based communication channels. The root cause is that
we did internal migration in the Smart Card Connector app to start using
ArrayBuffers for representing such fields, however due to the
cross-extension messages not supporting them we needed to transform this
into integer arrays before sending this. While #226 did this for
port-based communications, the single-message-based communication left
broken, resulting in some fields, like "atr", being received as "{}"
instead of the array of integer-bytes.

This fixes #393.

### DIFF
--- a/common/js/src/messaging/single-message-based-channel-unittest.js
+++ b/common/js/src/messaging/single-message-based-channel-unittest.js
@@ -205,6 +205,48 @@ goog.exportSymbol('testSingleMessageBasedChannelSending', function() {
   return testCasePromiseResolver.promise;
 });
 
+// Test that array buffers in sent messages are substituted with byte arrays.
+goog.exportSymbol(
+    'testSingleMessageBasedChannelArrayBufferSending', function() {
+      const MESSAGE_TYPE = 'foo';
+      const MESSAGE_DATA = {x: (new Uint8Array([1, 255])).buffer};
+      const EXPECTED_TRANSMITTED_DATA = {x: [1, 255]};
+      const EXPECTED_TRANSMITTED_MESSAGE =
+          new TypedMessage(MESSAGE_TYPE, EXPECTED_TRANSMITTED_DATA);
+
+      const testCasePromiseResolver = goog.Promise.withResolver();
+      const propertyReplacer = new goog.testing.PropertyReplacer;
+
+      const mockedSendMessage =
+          setupSendMessageMock(testCasePromiseResolver, propertyReplacer);
+      setupSendMessagePingResponding(mockedSendMessage);
+
+      chrome.runtime.sendMessage(
+          verifyChannelIdMatcher,
+          new goog.testing.mockmatchers.ObjectEquals(
+              EXPECTED_TRANSMITTED_MESSAGE.makeMessage()));
+      mockedSendMessage.$once();
+      mockedSendMessage.$replay();
+
+      function onChannelEstablished() {
+        globalChannel.send(MESSAGE_TYPE, MESSAGE_DATA);
+
+        mockedSendMessage.$verify();
+        testCasePromiseResolver.resolve();
+        globalChannel.dispose();
+      }
+
+      function onChannelDisposed() {
+        testCasePromiseResolver.reject();
+      }
+
+      globalChannel =
+          new GSC.SingleMessageBasedChannel(EXTENSION_ID, onChannelEstablished);
+      globalChannel.addOnDisposeCallback(onChannelDisposed);
+
+      return testCasePromiseResolver.promise;
+    });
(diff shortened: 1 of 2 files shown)


Fabian-Sommer (approved):
LGTM

PR #396, opened 2021-09-20
Don't call getBackgroundPage in non-app/extensions

emaxx-google:
Skip calling chrome.runtime.getBackgroundPage() when setting up logging
(originally added in #391), so that this code doesn't throw an exception
when this method is unavailable. This should never happen in production
currently, but might happen if the code is running outside an app/extension.

task #390.


Fabian-Sommer (approved):
LGTM

PR #380, opened 2021-09-10
Add build-time PACKAGING environment variable

emaxx-google:
This variable can be specified in the command line when running "make"
and specify the needed mode. By default, the "app" mode is used. The
value specified in the command line is exposed to the JavaScript code as
the GoogleSmartCard.Packaging.MODE constant.

This contributes to the app-to-extension migration effort, in particular
task #379.

PR #399, opened 2021-09-20
Fix GitHub binary preparation script for packaging

emaxx-google:
Update the prepare-binaries-for-github-release.py script to use the
correct directory name for input artifacts. Since #380, the directory
name has the pattern "$(PACKAGING)_$(TOOLCHAIN)_$(CONFIG)" (e.g.,
"app_pnacl_Release"), meanwhile it used to be just
"$(TOOLCHAIN)_$(CONFIG)".


Fabian-Sommer (approved):
LGTM

PR #388, opened 2021-09-16
Rename server.js and client.js and corresponding calls

ayaNader:
Rename the files "server.js" and "client.js" to give more context of what their job is.
Server creates the window.
Client sets up everything and shows it.

This a refactoring PR.

PR #400, opened 2021-09-20
Add app/extension modes for main window

ayaNader:
This PR adds functionality to handle both modes of the Smart Card Connector (i.e. app mode and extension mode), for the main window.
This is done by adding conditions to use app specific APIs if the app mode is chosen and extension specific APIs if the extension mode is chosen.

P.S. other PRs will be created to handle other functionalities for both modes.


common/js/src/popup-window/in-popup-main-script.js
```diff
@@ -99,7 +100,8 @@ GSC.InPopupMainScript.prepareAndShowAsModalDialog = function() {
   GSC.InPopupMainScript.setWindowHeightToFitContent();
   GSC.InPopupMainScript.setupClosingOnEscape();
   GSC.InPopupMainScript.setupRejectionOnWindowClose();
-  GSC.InPopupMainScript.showWindow();
+  if (GSC.Packaging.MODE == GSC.Packaging.Mode.APP)
```
emaxx-google, line 103: nit: Triple-equality (the identity operator) is preferred. See https://google.github.io/styleguide/jsguide.html#features-equality-checks 

P.S. I used the double-equality in my PR by mistake.

common/js/src/popup-window/in-popup-main-script.js
```diff
@@ -99,7 +100,8 @@ GSC.InPopupMainScript.prepareAndShowAsModalDialog = function() {
...
+    GSC.InPopupMainScript.showWindow();
```
emaxx-google, line 104: nit: Let's add a comment that explains why we're doing this in the App mode only. Something along the lines of:
```
// In the App packaging mode popup windows are initially opened hidden (see the flags
// passed to chrome.app.window.create() in popup-opener.js), to avoid flickering while
// the UI is being initialized. In the Extension packaging mode, there's no way to open an
// initially-hidden window, so there's no need to change the visibility here.
```

common/js/src/popup-window/in-popup-main-script.js
```diff
@@ -115,7 +117,11 @@ GSC.InPopupMainScript.setWindowHeightToFitContent = function() {
           typeof wholeContentHeight === 'number');
   goog.log.fine(
       logger, 'Resizing the window size to ' + wholeContentHeight + 'px');
-  chrome.app.window.current().innerBounds.height = wholeContentHeight;
+  if (GSC.Packaging.MODE == GSC.Packaging.Mode.APP) {
+    chrome.app.window.current().innerBounds.height = wholeContentHeight;
+  } else if (GSC.Packaging.MODE == GSC.Packaging.Mode.EXTENSION) {
+    chrome.windows.height = wholeContentHeight;
```
emaxx-google, line 123: Hmm, does it work? Seems odd - this, basically, creates a property called `height` in the `chrome.windows` namespace. I don't think that's how you set the opened window height.

common/js/src/popup-window/popup-opener.js
```diff
@@ -25,6 +25,7 @@ goog.provide('GoogleSmartCard.PopupOpener');
 goog.require('GoogleSmartCard.DebugDump');
 goog.require('GoogleSmartCard.Logging');
 goog.require('goog.Promise');
+goog.require('GoogleSmartCard.Packaging');
```
emaxx-google, line 28: nit: Please keep the imports sorted.

smart_card_connector_app/src/window-main.js
```diff
@@ -29,6 +29,7 @@ goog.require('GoogleSmartCard.ConnectorApp.Window.LogsExporting');
 goog.require('GoogleSmartCard.I18n');
 goog.require('GoogleSmartCard.Logging');
 goog.require('GoogleSmartCard.InPopupMainScript');
```
emaxx-google, line 31: BTW, looks like when #388 renamed the "Client"/"Server" modules, these lexicographic orders were broken. Let's sort the imports now.

common/js/src/popup-window/popup-opener.js
```diff
@@ -66,9 +67,11 @@ GSC.PopupOpener.createWindow = function(url, createWindowOptions, opt_data) {
...
   /** @preserveTry */
   try {
-    chrome.app.window.create(
-        url, createWindowOptions,
-        createWindowCallback.bind(null, createdWindowExtends));
+    if (GSC.Packaging.MODE == GSC.Packaging.Mode.APP) {
+      chrome.app.window.create(
+          url, createWindowOptions,
+          createWindowCallback.bind(null, createdWindowExtends));
+    }
```
emaxx-google, line 74: So this PR doesn't add the code for opening the popup in the extension mode? Was it forgotten, or is the plan to put it into a separate PR? (which would also be fine)

emaxx-google (changes requested):
Thanks.
+Fabian for another pair of eyes.

common/js/src/popup-window/popup-opener.js
ayaNader: yeah leaving that for a separate PR

Fabian-Sommer (approved):
LGTM

emaxx-google (approved):

PR #403, opened 2021-09-22
Add functionality to main window UI 

ayaNader:
This PR adds some functionality to parts of the extension's main window such as events for the About and Help buttons which cause corresponding windows to be opened. As well as adding the extension version of tracking list of readers.

### DIFF
--- a/smart_card_connector_app/src/window-about-showing.js
+++ b/smart_card_connector_app/src/window-about-showing.js
@@ -34,9 +34,16 @@ const ABOUT_WINDOW_URL = 'about-window.html';
 const ABOUT_WINDOW_WIDTH = 700;
 const ABOUT_WINDOW_HEIGHT = 500;
 
-const ABOUT_WINDOW_OPTIONS = {
+const APP_ABOUT_WINDOW_OPTIONS = {
   'id': 'about-window',
   'innerBounds': {'width': ABOUT_WINDOW_WIDTH, 'height': ABOUT_WINDOW_HEIGHT}
+};
+
+const EXTENSION_ABOUT_WINDOW_OPTIONS = {
+  url: ABOUT_WINDOW_URL,
+  type: 'popup',
+  width: ABOUT_WINDOW_WIDTH,
+  height: ABOUT_WINDOW_HEIGHT
 };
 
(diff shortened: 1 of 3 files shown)


smart_card_connector_app/src/window-about-showing.js
```diff
@@ -33,6 +34,8 @@ const ABOUT_WINDOW_OPTIONS = {
   'id': 'about-window',
   'innerBounds': {'width': 700, 'height': 500}
```
emaxx-google, line 35: nit: Let's use the new constants here as well.

emaxx-google (changes requested):

smart_card_connector_app/src/window-about-showing.js
```diff
@@ -46,7 +51,16 @@ const openAboutElement =
 function openAboutClickListener(e) {
   e.preventDefault();
 
-  chrome.app.window.create(ABOUT_WINDOW_URL, ABOUT_WINDOW_OPTIONS);
+  if (GSC.Packaging.MODE === GSC.Packaging.Mode.APP) {
+    chrome.app.window.create(ABOUT_WINDOW_URL, ABOUT_WINDOW_OPTIONS);
```
emaxx-google, line 55: nit: Actually, let's maybe delete the ABOUT_WINDOW_OPTIONS constant and just inline it here? It's a weird asymmetry that we have a constant for the App-specific API but no constant or the Extension-specific API.

emaxx-google (approved):
LGTM % nit

smart_card_connector_app/src/window-about-showing.js
ayaNader: wouldn't inlining it be too crowded?

emaxx-google: A matter of taste I guess, but for this case - where the constant is only useful for one conditional branch but not for the other - I think it's fine.

If you don't like it, then what about doing it the other way round, i.e., extracting the chrome.windows.create's argument into a global constant? And clearing naming one as a app-specific and another as an extension-specific constant?

emaxx-google (approved):
LGTM!

PR #404, opened 2021-09-23
Create permission popup UI for extension

ayaNader:
This PR creates the permissions popup UI for the extension mode which is shown when a third party app/extension requests access to the Smart Card Connector.


ayaNader:
@emaxx-google making the code symmetrical here when creating the popup wasn't the same as previously done as the window options for the app are sent as an argument, so using it to create the window in the extension mode required that I just use the needed values from the window options, WDYT should be done here?

emaxx-google:
> @emaxx-google making the code symmetrical here when creating the popup wasn't the same as previously done as the window options for the app are sent as an argument, so using it to create the window in the extension mode required that I just use the needed values from the window options, WDYT should be done here?

I see, that's a good question. While this might work for now, this is somewhat hacky: we're just taking some of the options and ignoring others.

I think the root cause is that the input argument (`createWindowOptions` passed to `GSC.PopupOpener.createWindow()`) uses an App-specific type (`chrome.app.window.CreateWindowOptions`). This can't work well if we want the same code to work in the Extensions mode as well. The proper solution would probably be to stop using this `chrome.app.window.CreateWindowOptions` type, and pass all bits of data via separate arguments. I.e., add arguments `windth`, `height`, etc.

common/js/src/popup-window/popup-opener.js
```diff
@@ -71,6 +71,13 @@ GSC.PopupOpener.createWindow = function(url, createWindowOptions, opt_data) {
       chrome.app.window.create(
           url, createWindowOptions,
           createWindowCallback.bind(null, createdWindowExtends));
+    } else if (GSC.Packaging.MODE === GSC.Packaging.Mode.EXTENSION){
+      chrome.windows.create({
+        url: url,
+        type: 'popup',
+        width: createWindowOptions['innerBounds']['width'],
+        height: createWindowOptions['innerBounds']['height']
```
emaxx-google, line 79: Do all callers actually provide width and height? E.g., user-prompting-checker.js only specifies the width (see the `USER_PROMPT_DIALOG_WINDOW_OPTIONS_OVERRIDES` constant). Or does it work because non-existing property evaluates to `undefined`, and `chrome.windows.create()` ignores properties whose values are undefined?

Another question is whether we even want to pass `height` anywhere? You already saw the code in `GSC.InPopupMainScript.setWindowHeightToFitContent()` - it automatically resizes the window to fit the content. If that's indeed working as intended, then perhaps no caller should provide the height? It'd be a good thing to try out.

ayaNader: > Do all callers actually provide width and height?
some calls to GSC.PopupOpener.createWindow use both width and height, so it's needed

> Another question is whether we even want to pass height anywhere?
you're right, GSC.InPopupMainScript.setWindowHeightToFitContent() should be enough to handle the height and indeed with trying to remove/keep height property didn't cause any difference in the window appearance

common/js/src/popup-window/popup-opener.js
```diff
@@ -113,7 +141,7 @@ GSC.PopupOpener.runModalDialog = function(
     Object.assign(dataWithDialogCallbacks, opt_data);
 
   GSC.PopupOpener.createWindow(
-      url, createWindowOptions, dataWithDialogCallbacks);
+      url, DEFAULT_DIALOG_CREATE_WINDOW_OPTIONS, dataWithDialogCallbacks);
```
emaxx-google, line 144: Is this change intentional? We're calculating `createWindowOptions` above but are never using it.

If you mean that `opt_createWindowOptionsOverrides` is always unused now, then let's kill this argument and the `createWindowOptions` variable. Otherwise, we need to respect them.

example_cpp_smart_card_client_app/src/background.js
```diff
@@ -109,8 +109,7 @@ const MAIN_WINDOW_OPTIONS = {
   'hidden': true,
   'id': MAIN_WINDOW_ID,
   'innerBounds': {
```
emaxx-google, line 111: Should it become just `'width': 500`, without nesting inside innerBounds? Your new typedef doesn't have innerBounds anymore.

common/js/src/popup-window/popup-opener.js
```diff
@@ -32,6 +32,22 @@ goog.require('goog.object');
 
 goog.scope(function() {
 
+const GSC = GoogleSmartCard;
+
+/**
+ * The window options to be used for the resulting window.
```
emaxx-google, line 38: Please document which options are supported in which mode (as I see that some of them are ignored in the Extension mode, presumably intentionally).

common/js/src/popup-window/popup-opener.js
```diff
@@ -50,34 +64,48 @@ const logger = GSC.Logging.getScopedLogger('PopupWindow.PopupOpener');
...
+            'visibleOnAllWorkspaces': windowOptions['visibleOnAllWorkspaces']
+          },
           createWindowCallback.bind(null, createdWindowExtends));
+    } else if (GSC.Packaging.MODE === GSC.Packaging.Mode.EXTENSION) {
+      chrome.windows.create({
+        url: url,
+        type: 'popup',
+        width: windowOptions['width'],
+      });
     }
```
emaxx-google, line 103: We need to throw an error when the `MODE` is unexpected, since otherwise `createWindow()` will do nothing, which is a surprising effect (especially if the caller then waits for a response from the window).

smart_card_connector_app/src/background-main-window-managing.js
```diff
@@ -40,7 +40,9 @@ const WINDOW_OPTIONS = {
   'frame': 'none',
   'hidden': true,
   'id': WINDOW_ID,
-  'innerBounds': {'width': 530, 'height': 430},
+  'innerBounds': {
```
emaxx-google, line 43: Ditto

emaxx-google (changes requested):
Thanks!

common/js/src/popup-window/popup-opener.js
ayaNader: oh no this is actually an unintentional change

ayaNader: oh no this is actually an unintentional change

example_cpp_smart_card_client_app/src/background.js
ayaNader: so innerBounds doesn't need to be defined?

common/js/src/popup-window/popup-opener.js
ayaNader: you mean like a nested try/catch?

emaxx-google (changes requested):

common/js/src/popup-window/popup-opener.js
```diff
@@ -32,6 +32,24 @@ goog.require('goog.object');
...
+ *            hidden:(boolean|undefined),
+ *            id:(string|undefined),
+ *            width:(number|undefined),
+ *            resizeable:(boolean|undefined),
+ *            visibleOnAllWorkspaces:(boolean|undefined)
+ *          }}
+ */
+let WindowOptions;
+
 const DEFAULT_DIALOG_CREATE_WINDOW_OPTIONS = {
```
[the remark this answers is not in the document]
ayaNader: i looked at other examples using typedef and the format was like so, the typedef written and then the object declared write below so that it shows the object name of this type, are you suggesting something like so?
// WindowOptions type
let WindowOptions;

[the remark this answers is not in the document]
emaxx-google: I meant
```
/** @type {WindowOptions} */
const DEFAULT_DIALOG_CREATE_WINDOW_OPTIONS = {
```

emaxx-google (changes requested):
Only the last open question about DEFAULT_DIALOG_CREATE_WINDOW_OPTIONS's type remains.

emaxx-google (approved):

PR #405, opened 2021-09-23
Allow vendor control commands for Omnikey 5427 CK

emaxx-google:
By default, the PC/SC protocol doesn't allow sending commands (APDUs) to
a reader that doesn't have a card inserted. There's a legitimate use
case (which exactly, isn't relevant here) for supporting this for the
Omnikey 5427 CK reader.

This reader provides a capability for sending APDUs, but only via a
vendor control command. By default, vendor control commands are
forbidden, and can be enabled for all readers via setting the
"ifdDriverOptions" parameter in the Info.plist config to have the bit
"1" (DRIVER_OPTION_CCID_EXCHANGE_AUTHORIZED). However, we don't want to
enable this parameter, since for some readers it might be a security
risk (see
https://ludovicrousseau.blogspot.com/2011/10/featureccidesccommand.html).
Hence the solution in this pull request is to special-case the Omnikey
5427 CK reader and allow any application to send control commands. It
was confirmed with the reader's vendor that this doesn't impose a
security risk here.


Fabian-Sommer (approved):
LGTM

emaxx-google:
Thanks. I'm only waiting to see whether the same change can be accepted upstream, in which case I'd grab the upstream's patch instead of my own (in order to simplify future upreds).

PR #406, opened 2021-09-23
Import CCID 1.4.36

emaxx-google:
PTAL. The only files that I manually edited are:

- third_party/ccid/README.google
- third_party/ccid/webport/include.mk


Fabian-Sommer (approved):
LGTM

PR #407, opened 2021-09-23
Import PC/SC-Lite 1.9.3

emaxx-google:
This commit also drops the patch that's not needed anymore,
as it went into the upstream release.


emaxx-google:
The only files I modified manually were:

- third_party/pcsc-lite/README.google

Fabian-Sommer (approved):
LGTM

Issue #377, opened 2021-08-06
Add support for using the smartcard connector app from a web page


xtralogic:
Currently the connector app accepts connections only from Chrome apps or extension and implements client whitelisting based on the extension id.
To accept connection from web pages the connector app's manifest.json needs to include "externally_connectable" key.
For the connections coming from web pages the extension id is null, so whitelisting mechanism needs to be updated to use the client web page origin URL, which is available in sender.url (see chrome.runtime.onMessageExternal.addListener).
The whitelisted url format should allow for wildcards, e.g. *.domain.com

emaxx-google:
Hi, thanks for filing this. This is in plans, although not yet in active development, since we're trying to find a way to limit it only to trusted web pages (for which being on the allowlist might not necessarily be a sufficient criteria alone). We'll share the details once it's fleshed out with security.

hcldan:
Any updates?

emaxx-google:
Nothing to share yet. An internal team in Google is still working on a security model for exposing such "sensitive" APIs.

hcldan:
@emaxx-google Would you (google) like to collaborate? I'd like to help where I can, if I can.

emaxx-google:
@hcldan This effort is still in the design stage, so unless you have specific ideas of how this can be designed, any concrete steps (like pull requests / change lists) aren't possible at the moment, unfortunately.

Basically, the challenge here is that a regular web page is considered to be an unsafe environment for accessing such a security/privacy sensitive API as PC/SC-Lite smart card access. The open web doesn't offer sufficient trust boundaries. Folks are trying to find a reasonable and maintainable architecture that would allow the page to declare (and prove) its conformance to stricter security boundaries, and gate access to sensitive APIs like PC/SC-Lite only to such pages.

hcldan:
"regular web page"?  
I mean there's a lot of development going on for PWA's where google is doing things in a secure way.
An https origin with a trusted certificate chain that's allow-listed by user opt-in or admin policy push isn't enough?

hcldan:
And to that effect... it would be great if this entire extension was merged into chrome-proper so I could use it in a PWA.
Could even prompt to allow access to it like NativeFileSystem does.. etc.

emaxx-google:
[...]

PR #408, opened 2021-09-27
Add helper for checking string being extension ID

emaxx-google:
Add a small helper for checking whether a string is (purely based on its
format) an extension ID.

The plan is to use this helper in follow-ups in order to distinguish
between extension IDs and generic web origins. Specifically, we're
adding the support of receiving smart card commands from web pages,
which means we'll store two kinds of permissions granted by the user:
ones based on Extension IDs and ones based on web origins; when loading
these persisted permissions, we'll need to determine the kind of every
stored item.

### DIFF
--- a/common/js/src/extension-id-unittest.js
+++ b/common/js/src/extension-id-unittest.js
@@ -1,0 +1,39 @@
+/**
+ * @license
+ * Copyright 2021 Google Inc. All Rights Reserved.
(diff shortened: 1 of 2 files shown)


Fabian-Sommer (approved):
LGTM

emaxx-google:
Thanks!

PR #409, opened 2021-09-27
Add messaging origin helpers

emaxx-google:
Introduce a concept of a "messaging origin", which is a unique
identifier of a remote endpoint that sends messages to us. Practically,
it's going to be used for identifying which extension or which website
is sending PC/SC commands to us, that is, to the Smart Card Connector.

An origin is a suitable abstraction that allows to make most of code
agnostic to whether the sender is an extension/app or a web page.


Fabian-Sommer (approved):
LGTM

PR #410, opened 2021-09-28
Generalize PC/SC permission prompt

emaxx-google:
Change the dialog to make it agnostic to whether the client application
is an extension/app or a web page. Also improve the wording.


Fabian-Sommer (approved):
LGTM

PR #411, opened 2021-09-28
Use "mimicking" ID for Connector in Debug builds

emaxx-google:
When building the Smart Card Connector in CONFIG=Debug, automatically
append the "key" property with the real Smart Card Connector's public
key.

This is useful in development and manual testing scenarios, since it
allows to load the just-built application under the production extension
ID ("khpfeaanjngmcnplbdlpegiifgpfgdco").

### DIFF
--- a/smart_card_connector_app/src/create_app_manifest.py
+++ b/smart_card_connector_app/src/create_app_manifest.py
@@ -162,7 +162,7 @@ def main():
       dest='target_file')
   args_parser.add_argument(
       '--enable-condition',
-      nargs='+',
+      action='append',
       dest='enabled_conditions')
   args = args_parser.parse_args()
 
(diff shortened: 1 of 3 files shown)


smart_card_connector_app/src/create_app_manifest.py
```diff
@@ -162,7 +162,7 @@ def main():
       dest='target_file')
   args_parser.add_argument(
       '--enable-condition',
-      nargs='+',
```
emaxx-google, line 165: Turns out, "nargs" didn't work properly with multiple arguments - it was creating a list, but always of size 1, regardless of the number of passed arguments o_O
"action=append" seems to do what we want - build a list of all conditions that were specified in the command line.

ayaNader (approved):
LGTM

PR #412, opened 2021-09-29
Switch known_client_apps.json to origins

emaxx-google:
Switch the JSON config that describes all client applications that are
"trusted" to list origins instead of extension IDs. This change is part
of the general plan to switch to origins, allowing us to support
requests from web pages in the same way as how we supported them from
extensions.

Also rename a few source files and classes to better reflect the new
function: "KnownApp" => "TrustedClientInfo", "KnownAppsRegistry" =>
"TrustedClientsRegistry". We don't rename the JSON config since it'll
break github links in various old docs and conversations.


Fabian-Sommer (approved):
LGTM % nit

PR #413, opened 2021-09-29
Switch PortMessageChannel JS class to origins

emaxx-google:
Refactor the class to use a more generic notion of "origins" instead of
"extension IDs".


Fabian-Sommer (approved):
LGTM

PR #414, opened 2021-09-29
Switch UserPromptingChecker JS class to origins

emaxx-google:
Refactor the class to use a more generic notion of "origins" instead of
"extension IDs".

The important detail here is that we're switching the persistent storage
format: for every client app for which the user clicked "allow" we'll now
store an origin instead of an extension ID. In order to preserve
backwards compatibility and not forget about clients previously
approved by the user, we're performing the migration on-the-flight
when loading the persistent storage.

### DIFF
--- a/third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker-unittest.js
+++ b/third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker-unittest.js
@@ -40,12 +40,15 @@ const UserPromptingChecker = GSC.PcscLiteServerClientsManagement
                                  .PermissionsChecking.UserPromptingChecker;
 const ignoreArgument = goog.testing.mockmatchers.ignoreArgument;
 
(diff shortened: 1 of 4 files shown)


third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker-unittest.js
```diff
@@ -254,85 +258,133 @@ goog.exportSymbol(
...
-             {[FAKE_CLIENT_1_ID]: true}} /* fakeInitialStorageData */,
+        {
+          [STORAGE_KEY]: {[FAKE_CLIENT_1_ORIGIN]: true}
+        } /* fakeInitialStorageData */,
+        null /* expectedStorageDataToBeWritten */,
+        MockedDialogBehavior.NOT_RUN /* mockedDialogBehavior */,
+        function(userPromptingChecker) {
+          return userPromptingChecker.check(FAKE_CLIENT_1_ORIGIN);
+        }));
+goog.exportSymbol(
```
Fabian-Sommer, line 276: nit: I think each test should be preceded by an empty line (here and in a few other places in this file)

Fabian-Sommer (approved):
LGTM % nit

PR #415, opened 2021-09-29
Switch to client_name_for_log in C++ code

emaxx-google:
Change the semantics of the parameter sent from JS to C++ to be not an
"extension ID" but instead a generic "name for log". The idea is that
once we switch to using origins we want to still use something shorter
in C++ logs, so we're renaming the parameter to reflect its actual
function and preparing to send any kind of information there.


Fabian-Sommer (approved):
LGTM

PR #416, opened 2021-09-29
Implement PC/SC handling of requests from web pages

### LINKED ISSUE #377: Add support for using the smartcard connector app from a web page

emaxx-google:
Add implementation for receiving PC/SC commands from web pages,
in addition to the existing capability of that done by extensions/apps.

A web page will need to open a messaging port to us (to the Smart
Card Connector) and send commands according to our protocol (e.g.,
using the example_js_standalone_smart_card_client_library).

This fixes #377: supporting PC/SC requests from web pages.
NOTE: this commit is only adding underlying implementation, but NOT
allowing any real-world page to send commands, as we keep the
"externally_connectable" manifest property empty. The latter is still
blocked on security discussions around web pages' trust modes.


third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/checker.js
```diff
@@ -76,74 +76,73 @@ Checker.prototype.logger = GSC.Logging.getScopedLogger(
...
 
 /**
- * Starts the permission check for the given client app.
+ * Starts the permission check for the given client application.
  *
  * The result is returned asynchronously as a promise (which will be eventually
  * resolved if the permission is granted or rejected otherwise).
- * @param {string|null} clientAppId ID of the client app, or null if the client
- * is our own app.
+ * @param {string|null} clientOrigin Orogin of the client application, or null
```
Fabian-Sommer, line 83: nit: "Origin" instead of "Orogin"

Fabian-Sommer (approved):
LGTM % nits

PR #422, opened 2021-10-25
Fix JS minification issues

emaxx-google:
Fix a few places in the code to correctly work in Release mode with the
recently enabled JavaScript code minification (see #420).


ayaNader (approved):
LGTM

PR #423, opened 2021-10-25
Backend functionality for permission popup in extension mode

ayaNader:
This PR creates the backend functionality of the permission popup window which appears to request access for the Smart Card Connector by the third party app/extension.
This means that it handles the flow after which the user presses "Allow" or "Block" in the permission popup window, and resolves the dialog.


common/js/src/popup-window/popup-opener.js
```diff
@@ -140,15 +149,32 @@ GSC.PopupOpener.runModalDialog = function(
...
+  goog.global[`googleSmartCard_resolveModalDialog${lastUsedPopupId}`] =
+      promiseResolver.resolve;
+  goog.global[`googleSmartCard_rejectModalDialog${lastUsedPopupId}`] =
+      promiseResolver.reject;
+
   if (opt_data !== undefined)
     Object.assign(dataWithDialogCallbacks, opt_data);
 
+  if (GSC.Packaging.MODE === GSC.Packaging.Mode.EXTENSION && opt_data) {
+    url.searchParams.append('passed_data', JSON.stringify(opt_data));
```
emaxx-google, line 175: The bug with modifying the input `url` parameter is still here. Maybe it's still unclear for you what the issue is - let me illustrate it:

```
function f(url) {
  url.searchParams.append('passed_data', 'foo');
}
let someUrl = new URL('http://example.org');
f(someUrl);
console.log(someUrl.toString());
// prints: 'http://example.org/?passed_data=foo'
f(someUrl);
console.log(someUrl.toString());
// prints: 'http://example.org/?passed_data=foo&passed_data=foo'
```

As we can see from these results, the function is only modifying the "local" variable, but the changes end up in the caller's variable, which is totally wrong.

Regarding how to fix it - we need to make a copy of the `url` object. I guess this should work: `const modifiedUrl = new URL(url)`.

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker.js
```diff
@@ -148,6 +150,9 @@ UserPromptingChecker.prototype.check = function(clientOrigin) {
       },
       this);
 
+  delete goog.global[`googleSmartCard_resolveModalDialog${GSC.PopupOpener.lastUsedPopupId}`];
```
emaxx-google, line 153: This deletion should live under /popup-window/ - it shouldn't be that every callsite needs to take care of this cleanup. Also it's technically not fully precise: in case there are two dialogs shown at the same time, `lastUsedPopupId` will be "2", and the first dialog's variables will stay as a garbage.

emaxx-google (changes requested):

common/js/src/popup-window/popup-opener.js
ayaNader: ah ok, actually i understood the idea i just didn't see why we don't want to update the url itself (and not just the local variable), but the idea is that we create a new window with this modified url so that's why we don't need to change the actual current url

common/js/src/popup-window/in-popup-main-script.js
```diff
@@ -61,25 +64,26 @@ GSC.InPopupMainScript.showWindow = function() {
   chrome.app.window.current().show();
 };
 
+const lastUsedPopupId = GSC.InPopupMainScript.getData()['popup_id'];
```
emaxx-google, line 67: nit: A better name would be "currentPopupId" or "ourPopupId".

It's not necessarily a "lastUsedPopupId", since there might be multiple popups opened.

common/js/src/popup-window/popup-opener.js
```diff
@@ -67,9 +67,17 @@ const DEFAULT_DIALOG_CREATE_WINDOW_OPTIONS = {
  */
 const logger = GSC.Logging.getScopedLogger('PopupWindow.PopupOpener');
 
+/**
+ * ID used to differentiate between different popup instances.
+ */
+let lastUsedPopupId = 0;
+
+/** @const */
+GSC.PopupOpener.lastUsedPopupId = lastUsedPopupId;
```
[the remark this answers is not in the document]
emaxx-google: ping on this nit

common/js/src/popup-window/popup-opener.js
```diff
@@ -140,17 +149,40 @@ GSC.PopupOpener.runModalDialog = function(
...
-    'rejectModalDialog': promiseResolver.reject
-  };
+  // Set data for corresponnding SCC mode
+  const modifiedData =
+      (GSC.Packaging.MODE === GSC.Packaging.Mode.APP ?
+           {
+             'resolveModalDialog': promiseResolver.resolve,
+             'rejectModalDialog': promiseResolver.reject
+           } :
+           {'popup_id': lastUsedPopupId.toString()});
```
emaxx-google, line 164: nit: I suspect `.toString()` is not actually needed here.

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker.js
```diff
@@ -148,6 +150,11 @@ UserPromptingChecker.prototype.check = function(clientOrigin) {
       },
       this);
 
+  delete goog.global[`googleSmartCard_resolveModalDialog${
```
emaxx-google, line 153: This code should be moved to popup-opener.js, because this code would be the same for every piece of code that wants to show a popup

emaxx-google (changes requested):
Thanks! We're really close - the last piece of code to be moved to the right location, and some nits.

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker.js
ayaNader: you mean because it's not being deleted per popup?

emaxx-google: Ah, that's another bug indeed. But my comment was referring to the location of this code: the right abstraction layer for this would be the common/js/popup-window/ folder.

common/js/src/popup-window/popup-opener.js
```diff
@@ -70,7 +70,7 @@ const logger = GSC.Logging.getScopedLogger('PopupWindow.PopupOpener');
 /**
  * ID used to differentiate between different popup instances.
  */
-let lastUsedPopupId = 0;
```
emaxx-google: This variable should actually stay "lastUsedPopupId". My comment was about renaming in in-popup-main-script.js.

To reiterate, the point was that the opener might open multiple popups, and its variable "lastUsedPopupId" is a counter; meanwhile the in-popup-script only stores its own id, which is why "currentPopupId" is better there.

emaxx-google (approved):
LGTM

ayaNader:
Previously fetching the popup_id straight away was causing UI glitches due to "popup_id" being null...this change fixed it

emaxx-google:
> Previously fetching the popup_id straight away was causing UI error due to "popup_id" being null...this change fixed it.

Hmm, the changes in the last commit should be a no-op... If there's some issue, let's investigate it - maybe it's not fixed yet? Did you see some errors in the console, or what kind of UI error was that?

ayaNader:
Yeah the UI was completely blank and I found an error that "popup_id" was null.

ayaNader:
> > Previously fetching the popup_id straight away was causing UI error due to "popup_id" being null...this change fixed it.
> 
> Hmm, the changes in the last commit should be a no-op... If there's some issue, let's investigate it - maybe it's not fixed yet? Did you see some errors in the console, or what kind of UI error was that?

Yeah the UI was completely blank and I found an error that "popup_id" was null.

emaxx-google:
> > > Previously fetching the popup_id straight away was causing UI error due to "popup_id" being null...this change fixed it.
> > 
> > 
> > Hmm, the changes in the last commit should be a no-op... If there's some issue, let's investigate it - maybe it's not fixed yet? Did you see some errors in the console, or what kind of UI error was that?
> 
> Yeah the UI was completely blank and I found an error that "popup_id" was null.

Then it means we have a problem and "forget" to pass popup_id in some cases? Every popup should have it.

common/js/src/popup-window/in-popup-main-script.js
```diff
@@ -59,20 +64,32 @@ GSC.InPopupMainScript.showWindow = function() {
...
 GSC.InPopupMainScript.resolveModalDialog = function(result) {
-  const callback = GSC.InPopupMainScript.getData()['resolveModalDialog'];
-  GSC.Logging.checkWithLogger(logger, callback);
   goog.log.fine(
       logger,
       'The modal dialog is resolved with the following result: ' +
           GSC.DebugDump.debugDump(result));
-  callback(result);
+  if (GSC.Packaging.MODE === GSC.Packaging.Mode.APP) {
+    const callback = GSC.InPopupMainScript.getData()['resolveModalDialog'];
```
emaxx-google, line 81: nit: Let's use the new `popUpData` variable here and in other places throughout the file. It'll save on extra calls, but more importantly make the code look less surprising.

emaxx-google (approved):
LGTM % nit

PR #426, opened 2021-11-10
Fix IPC emulation waiting with no timeout

emaxx-google:
Fix the WaitForInMemoryFileCanBeRead() method to actually perform
waiting when the "timeout_milliseconds" parameter is passed as a null
optional.

The bug was caused by the fact that, apparently, std::chrono's max()
method is impractical to use: waiting for such a max() timeout actually
is a no-op, because the standard library adds the timeout to the current
time, causing the integere overflow.

The fix is to revert to the older version that used two separate calls
sleep() and sleep_for() depending on the parameter passed.

This commit contributes to the WebAssembly migration effort, tracked by
issue #233.


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #428, opened 2021-11-11
Pass --bind only to the Emscripten linker

emaxx-google:
The --bind flag causes an error in recent Emscripten releases when
passed to the compiler. It's a linker-only flag, so change our build
scripts to only pass it to the linker.

This change contributes to the WebAssembly migration effort, tracked
by #177.


Fabian-Sommer (approved):
LGTM

Issue #429, opened 2021-11-17
Add WebUSB support


emaxx-google:
We need to switch to WebUSB while migrating the Smart Card Connector from Apps to Extensions, since the previously used chrome.usb API is Apps-only.

PR #430, opened 2021-11-17
[usb] Rename LibusbOverChromeUsbGlobal

emaxx-google:
Rename LibusbOverChromeUsbGlobal to LibusbWebPortService. Also move it
from "google_smart_card_libusb/global.{h,cc}" to
"public/libusb_web_port_service.{h,cc}", and change files that include
it to specify full path - for making the code easier to follow.

This is a pure refactoring change. It's preparatory work for the WebUSB
support effort tracked by #429.


Fabian-Sommer (approved):
LGTM

PR #433, opened 2021-11-19
[usb] Add JsProxyReceiver

emaxx-google:
Add a JavaScript class for handling the Libusb requests sent from the
C++ code. Redirect the C++ code to send requests to this class instead
of the ChromeUsbBackend JS class.

Currently the new JsProxyReceiver class only has a stub implementation
that forwards all requests to ChromeUsbBackend. Follow-up commits will
change it, by introducing an abstraction layer that abstracts away
differences between chrome.usb and WebUSB APIs.

### DIFF
--- a/third_party/libusb/webport/src/libusb-proxy-receiver.js
+++ b/third_party/libusb/webport/src/libusb-proxy-receiver.js
@@ -1,0 +1,84 @@
+/**
+ * @license
+ * Copyright 2021 Google Inc.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


third_party/libusb/webport/src/libusb-proxy-receiver.js
```diff
@@ -0,0 +1,87 @@
...
+goog.require('GoogleSmartCard.Libusb.ChromeUsbBackend');
+goog.require('GoogleSmartCard.Logging');
+goog.require('GoogleSmartCard.RemoteCallMessage');
+goog.require('goog.Promise');
+goog.require('goog.log');
+goog.require('goog.messaging.AbstractChannel');
+
+goog.scope(function() {
+
+const EXECUTABLE_MODULE_REQUESTER_NAME = 'libusb';
```
Fabian-Sommer, line 32: nit: Perhaps add a comment here that this needs to match the string in api_bridge.h

Fabian-Sommer (approved):
LGTM % nit

emaxx-google:
thanks!

PR #434, opened 2021-11-19
Rename Detach() to ShutDown()

emaxx-google:
Rename all Detach() methods in the C++ code to ShutDown(). The new name
is more self-explanatory.

This is a pure refactoring commit. It's done as part of the preparation
for the WebUSB implementation effort, since we'ge going to add a bunch
of new classes and want to use consistent naming.

### DIFF
--- a/smart_card_connector_app/src/application.cc
+++ b/smart_card_connector_app/src/application.cc
@@ -50,8 +50,8 @@ Application::Application(
 
 Application::~Application() {
   // Intentionally leak objects that might still be used by background threads.
-  // Only shut them down (which makes them stop referring to `this` and sending
-  // requests to the JavaScript side).
+  // Only shut them down (which makes them stop referring to `this` and stop
+  // sending requests to the JavaScript side).
   libusb_web_port_service_->ShutDown();
   (void)libusb_web_port_service_.release();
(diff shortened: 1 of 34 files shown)


emaxx-google:
@Fabian-Sommer IIRC you actually suggested the name `ShutDown` some time ago, and I argued it doesn't exactly reflect what the method is doing, so we ended up using `DisableJsCommunication` in some places and the old `Detach` in others. But now when I started to rewrite some of the code, I realized that `ShutDown` actually works better and is simpler to follow; thanks :)

smart_card_connector_app/src/application.cc
```diff
@@ -50,8 +50,9 @@ Application::Application(
 
 Application::~Application() {
   // Intentionally leak objects that might still be used by background threads.
-  // Only detach them from `this` and the JavaScript side.
-  libusb_web_port_service_->Detach();
+  // Only shut them down (which makes them stop referring to `this` and sending
```
ayaNader, line 53: for clarity: "and stop sending"?

ayaNader (approved):
LGTM %nit

Fabian-Sommer (approved):
LGTM

PR #435, opened 2021-11-19
[wasm] Minify the Emscripten JS files

emaxx-google:
Use the Closure Compiler to minify the JavaScript files produced by
Emscripten. This reduces the executable_module.js file size by 55% (from
177KB to 80KB), which should make the .crx file be installed faster
and the JS code to be loaded faster.

### DIFF
--- a/common/make/internal/executable_building_emscripten.mk
+++ b/common/make/internal/executable_building_emscripten.mk
@@ -123,6 +123,13 @@ EMSCRIPTEN_COMMON_FLAGS += \
 EMSCRIPTEN_COMPILER_FLAGS += \
   -DNDEBUG \
 
+# Add linker flags specific to release builds.
+#
+# Explanation:
+# closure=1: Minify the auxilliary .js file produced by Emscripten.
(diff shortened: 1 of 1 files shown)


emaxx-google:
This is blocked on Emscripten bugs - it performs incorrect minification when using our combination of flags, so that the resulting file can't be loaded.

emaxx-google:
> This is blocked on Emscripten fixing the bug - it performs incorrect minification when using our combination of flags, so that the resulting file can't be loaded.

The bug is https://github.com/emscripten-core/emscripten/issues/15585 .

emaxx-google:
> > This is blocked on Emscripten fixing the bug - it performs incorrect minification when using our combination of flags, so that the resulting file can't be loaded.
> 
> The bug is [emscripten-core/emscripten#15585](https://github.com/emscripten-core/emscripten/issues/15585) .

The bug is fixed now (and I verified it on ToT), so we can proceed with this optimization.

ayaNader (approved):
LGTM

Fabian-Sommer (approved):
LGTM

Issue #440, opened 2021-11-21
Permission from third party to access Smart Card Connector not prompted  


ayaNader:
When launching the third party app/extension (such as CSSI or CACKey), a permission prompt must be shown to ask the user whether or not they allow the the third party to access the Smart Card Connector. This happens in the extension mode but in the app mode the permission prompt isn't shown and the third party automatically has access to the Smart Card Connector App.
This issue occurred after the changes made in #423.

TJ4Urservices:
Yes wow this happens every time i do multiple password change like i would go sign in to my Gmail account and find out the password was changed. Then just literally two seconds after I change the password he would change it and then I would change it again. This would be consistently on all the emails going on for hours upon hours emotionally and physically just trying to be faster than him trying.

PR #441, opened 2021-11-21
Resolve permission prompt not showing in App mode

ayaNader:
This PR solves issue #440 where the permission prompt wasn't showing due to permission prompt URL being rejected for security reasons.


common/js/src/popup-window/popup-opener.js
```diff
@@ -92,8 +92,9 @@ GSC.PopupOpener.createWindow = function(url, windowOptions, opt_data) {
   /** @preserveTry */
   try {
     if (GSC.Packaging.MODE === GSC.Packaging.Mode.APP) {
+      const loaclUrl = url.replace("chrome-extension://khpfeaanjngmcnplbdlpegiifgpfgdco", "");
```
emaxx-google, line 95: We shouldn't hardcode the "chrome-extension://" or the "khpfeaanjngmcnplbdlpegiifgpfgdco" parts here - it's intended as a generic code that should work regardless of these concrete details. Also low-level string replacement will break in case the searched pattern occurs in other parts of the URL - say, in the URL parameters.

I think the original root cause issue is that our API in this file became asymmetric in #440: `runModalDialog()` takes a "hostless" URL (I don't know what's the proper term for this), but then passes a URL with host to `createWindow()`. So can we fix `runModalDialog()` to create a "hostless" URL and pass it further? I don't know if JavaScript has a nicer helper for this, but it can be something like

```
modifiedUrl.pathname + modifiedUrl.search + modifiedUrl.hash
```

emaxx-google (changes requested):

emaxx-google (approved):
LGTM, thanks.

PR #443, opened 2021-11-22
[build] Specify output JS language version

emaxx-google:
Explicitly tell the Closure Compiler to make the resulting files to
conform to the JavaScript specification not newer than year 2015,
because our app's minimum_chrome_version is still 48 (released in
2015/2016).

This is an additional safety measure: most likely we aren't using
unsupported constructs anyway, but the explicit flags will make sure of
that.


Fabian-Sommer (approved):
LGTM

TJ4Urservices:
Thank you

PR #445, opened 2021-11-23
[usb] Listing devices via WebUSB

emaxx-google:
Add the WebUSB implementation of the libusb-to-JS adaptor. Add the
"listDevices" method implementation into it.


Fabian-Sommer (approved):
LGTM

PR #446, opened 2021-11-23
[style] Call JS constructors with brackets

emaxx-google:
The Google JavaScript Style Guide requires to always specify brackets
when invoking constructors. Fix a couple of places in our code that
violate that.


Fabian-Sommer (approved):
LGTM

PR #447, opened 2021-11-24
[usb] Keep a map of chrome.usb devices

emaxx-google:
Store all currently available chrome.usb Device objects in a mapping on
the JS side, so that we can pass the needed Device instance to
chrome.usb when the C++ side asks us to perform a request for the
given device ID.


Fabian-Sommer (approved):
LGTM

PR #448, opened 2021-11-24
[usb] Get device configs via new JS proxy

emaxx-google:
Add structures and boilerplate for the device config passing into the
libusb-to-JS adaptor, and provide its implementation for the chrome.usb
API.

### DIFF
--- a/third_party/libusb/webport/src/libusb-proxy-data-model.js
+++ b/third_party/libusb/webport/src/libusb-proxy-data-model.js
@@ -19,6 +19,10 @@
 
 goog.provide('GoogleSmartCard.LibusbProxyDataModel');
 
+goog.scope(function() {
+
+const GSC = GoogleSmartCard;
+
 // The following definitions must match the ones in
 // libusb_js_proxy_data_model.h.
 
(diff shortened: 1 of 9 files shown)


third_party/libusb/webport/src/libusb-proxy-data-model.js
```diff
@@ -33,4 +37,72 @@ goog.provide('GoogleSmartCard.LibusbProxyDataModel');
...
  */
-GoogleSmartCard.LibusbProxyDataModel.LibusbJsDevice;
+GSC.LibusbProxyDataModel.LibusbJsDevice;
+
+const LibusbJsDevice = GSC.LibusbProxyDataModel.LibusbJsDevice;
+
+/**
+ * @enum {string}
+ */
+GSC.LibusbProxyDataModel.LibusbJsDirection = {
```
Fabian-Sommer, line 47: nit: Here we could mention that these need to be kept in sync with third_party/libusb/webport/src/libusb_js_proxy_data_model.cc

Fabian-Sommer (approved):
LGTM % nit

third_party/libusb/webport/src/libusb-proxy-data-model.js
emaxx-google: Done, and also added similar comments for every other struct.

PR #449, opened 2021-11-24
[usb] Open/close devices via new JS proxy

emaxx-google:
Implement opening/closing a device via the new libusb-to-JS adaptor and
its chrome.usb implementation.


ayaNader (approved):
LGTM

PR #450, opened 2021-11-24
[usb] Fix memory leak on close failure

emaxx-google:
Don't leak the libusb_device_handle memory when the handle closing
failed. Such failures should never happen under normal circumstances,
but it's still better to avoid such leaks.


Fabian-Sommer (approved):
LGTM

PR #454, opened 2021-11-25
[usb] Add types for control transfer

emaxx-google:
Add data model and implement the JS side of the USB control transfers in
the new libusb-to-JS adaptor.

The C++ side is not wired up with this new code yet.


Fabian-Sommer (approved):
LGTM

PR #455, opened 2021-11-25
[usb] Decouple USB destination from chrome.usb

emaxx-google:
Refactor the UsbTransferDestination struct to stop using the chrome.usb
entities internally, and also to expose factory methods that don't use
chrome.usb entities.

This is a preparation for switching the callsites away from using
chrome.usb. It's preparatory work for the WebUSB support effort tracked
by #429.

### DIFF
--- a/third_party/libusb/webport/src/usb_transfer_destination.cc
+++ b/third_party/libusb/webport/src/usb_transfer_destination.cc
@@ -26,11 +26,83 @@
 
 #include <tuple>
 
+#include <libusb.h>
+
 namespace google_smart_card {
+
+namespace {
+
+// TODO(#429): Delete this converter once the C++ code gets fully abstracted
+// away from chrome.usb.
+uint8_t GetLibusbRequestType(
+    const chrome_usb::ControlTransferInfo& chrome_usb_control_transfer_info) {
+  uint8_t request_type = 0;
+  switch (chrome_usb_control_transfer_info.request_type) {
+    case chrome_usb::ControlTransferInfoRequestType::kStandard:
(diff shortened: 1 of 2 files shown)


third_party/libusb/webport/src/usb_transfer_destination.cc
```diff
@@ -26,34 +26,123 @@
...
 bool UsbTransferDestination::IsInputDirection() const {
-  return direction_ == chrome_usb::Direction::kIn;
+  // It's only correct to check the flag presence by doing bitwise and against a
+  // non-zero constant.
+  static_assert(LIBUSB_ENDPOINT_IN != 0,
+                "Bad mask: LIBUSB_ENDPOINT_IN is zero");
+
+  if (control_transfer_request_type_) {
+    // For control transfers, the direction is encoded in the request type.
+    return (*control_transfer_request_type_ & LIBUSB_ENDPOINT_IN) != 0;
```
Fabian-Sommer, line 137: As discussed, better to use (*control_transfer_request_type_ & LIBUSB_ENDPOINT_DIR_MASK) == LIBUSB_ENDPOINT_IN here

third_party/libusb/webport/src/usb_transfer_destination.cc
```diff
@@ -26,34 +26,123 @@
...
+  }
+  if (endpoint_address_) {
+    // For all other transfer types, the direction is encoded in the endpoint
+    // address.
+    return (*endpoint_address_ & LIBUSB_ENDPOINT_IN) != 0;
```
Fabian-Sommer, line 142: ditto

Fabian-Sommer (approved):
LGTM % nits

PR #456, opened 2021-11-25
[usb] Implement sync control transfer via async

emaxx-google:
Change our libusb_control_transfer implementation - which performs
synchronous transfer - to redirect to our asynchronous transfer
implementation.

This helps streamlining various USB transfer flows. When migrating to
the new libusb-to-JS adaptor, we'll only need to take care of the
asynchronous code path.


third_party/libusb/webport/src/libusb_js_proxy.cc
```diff
@@ -952,37 +960,40 @@ int LibusbJsProxy::LibusbControlTransfer(libusb_device_handle* dev,
...
+  // Implement the synchronous transfer in terms of asynchronous one.
+  libusb_transfer transfer;
+  memset(&transfer, 0, sizeof(transfer));
+
+  // Libusb requires the control transfer's setup packet (of size
+  // `LIBUSB_CONTROL_SETUP_SIZE`) to precede the data buffer.
+  std::vector<uint8_t> buffer(LIBUSB_CONTROL_SETUP_SIZE + wLength);
+  libusb_fill_control_setup(buffer.data(), bmRequestType, bRequest, wValue,
+                            index, wLength);
+  if (!(bmRequestType & LIBUSB_ENDPOINT_IN)) {
```
Fabian-Sommer, line 972: Did you mean LIBUSB_ENDPOINT_OUT here? If not here, below in line  991?

Fabian-Sommer (approved):
LGTM % question

third_party/libusb/webport/src/libusb_js_proxy.cc
emaxx-google: That'd be a bug ;) LIBUSB_ENDPOINT_OUT is all-zeroes, unlike LIBUSB_ENDPOINT_IN.

https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/blob/91c8e5e7a6d4c4d058107269f5758ead95831d01/third_party/libusb/webport/src/libusb_js_proxy.cc#L630
Let me do that, now that you asked about it.

Fabian-Sommer: Ah, now I understand this - I misread the & as a &&. Yeah, in that case LIBUSB_ENDPOINT_IN makes sense in both places.

Fabian-Sommer: But I agree, l 630 is more readable than either l 972 or 991

PR #458, opened 2021-11-25
[usb] Do sync bulk/interrupt transfers via async

emaxx-google:
Change our libusb_bulk_transfer/libusb_interrupt_transfer implementation -
which perform synchronous transfers - to redirect to our asynchronous
transfer implementation.

Also delete various helpers for sync transfers which became unnecessary
now.

This helps streamlining various USB transfer flows. When migrating to
the new libusb-to-JS adaptor, we'll only need to take care of the
asynchronous code path. It's preparatory work for the WebUSB
support effort tracked by #429.

### DIFF
--- a/third_party/libusb/webport/src/libusb_js_proxy.cc
+++ b/third_party/libusb/webport/src/libusb_js_proxy.cc
@@ -1003,30 +1003,8 @@ int LibusbJsProxy::LibusbBulkTransfer(libusb_device_handle* dev,
                                       int* actual_length,
                                       unsigned timeout) {
   GOOGLE_SMART_CARD_CHECK(dev);
-
-  chrome_usb::GenericTransferInfo transfer_info;
-  CreateChromeUsbGenericTransferInfo(endpoint_address, data, length, timeout,
-                                     &transfer_info);
-  const UsbTransferDestination transfer_destination =
-      UsbTransferDestination::CreateFromChromeUsbGenericTransfer(
-          GetChromeUsbConnectionHandle(*dev), transfer_info);
-  SyncTransferHelper sync_transfer_helper(
-      contexts_storage_.FindContextByAddress(dev->context()),
-      transfer_destination);
-
-  chrome_usb_api_bridge_->AsyncBulkTransfer(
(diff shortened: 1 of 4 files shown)


third_party/libusb/webport/src/libusb_js_proxy.cc
```diff
@@ -1175,4 +1097,47 @@ int LibusbJsProxy::LibusbHandleEventsWithTimeout(libusb_context* context,
...
+
+  // Implement the synchronous transfer in terms of asynchronous one.
+  libusb_transfer transfer;
+  std::memset(&transfer, 0, sizeof(transfer));
+
+  int transfer_completed = 0;
+  if (transfer_type == LIBUSB_TRANSFER_TYPE_BULK) {
+    libusb_fill_bulk_transfer(&transfer, device_handle, endpoint_address, data,
+                              length,
+                              /*callback=*/
```
Fabian-Sommer, line 1118: nit: The formatting here looks weird to me - why is there a newline here?

third_party/libusb/webport/src/libusb_js_proxy.cc
```diff
@@ -1175,4 +1097,47 @@ int LibusbJsProxy::LibusbHandleEventsWithTimeout(libusb_context* context,
...
+                              &OnSyncTransferCompleted,
+                              /*user_data=*/&transfer_completed, timeout);
+  } else if (transfer_type == LIBUSB_TRANSFER_TYPE_INTERRUPT) {
+    libusb_fill_interrupt_transfer(&transfer, device_handle, endpoint_address,
+                                   data, length,
+                                   /*callback=*/
```
Fabian-Sommer, line 1124: ditto

Fabian-Sommer (approved):
LGTM % nit

third_party/libusb/webport/src/libusb_js_proxy.cc
emaxx-google: Done - Was probably the case where clang-tidy fails on the side of respecting original newlines (thinking that they're intentional, meanwhile they're not).

PR #459, opened 2021-11-25
[usb] Use new libusbjs transfer type

emaxx-google:
Migrate the LibusbJsProxy code to switch away from
chrome_usb::TransferResult to LibusbJsTransferResult.

This is expected to be a pure refactoring change. It's preparatory work
for the WebUSB support effort tracked by #429.


Fabian-Sommer (approved):
LGTM

PR #460, opened 2021-11-25
[js] Fix bad "this" in TrustedClientsRegistry

emaxx-google:
The code in TrustedClientsRegistry made an incorrect dereference of
"this" inside a local function. Fix this by changing the function to be
a lambda (a.k.a. "arrow function"), which inherits "this" from the outer
scope.

The bug was introduced in #412. Not sure why it wasn't visible earlier -
probably the recent changes of JS compilation levels made the compiler
produce slightly different code. Also it's unclear why the compiler
didn't emit a warning on this mistake.


Fabian-Sommer (approved):
LGTM

PR #461, opened 2021-11-25
[usb] JS base for bulk/interrupt transfers

emaxx-google:
Add types for the wire format of bulk and interrupt USB transfers in the
libusb-to-JS adaptor, and implement these transfers in the chrome.usb
adaptor.

The C++ side is not wired up with this new code yet.


Fabian-Sommer (approved):
LGTM

PR #463, opened 2021-11-26
[usb] Fix chrome.usb empty output transfer results

emaxx-google:
Contrary to the chrome.usb API docs, the transfer result's "data" field
is always provided, even for output transfers - for them it's set to an
empty array buffer. Fix our code to stop being confused by that, since
we were misinterpreting this as a zero-length write, effectively the
failure to do output transfers.


Fabian-Sommer (approved):
LGTM

PR #464, opened 2021-11-26
[usb] Fix synchronous transfer event loop

emaxx-google:
We need to call libusb_handle_events_completed() in a loop: this
function doesn't promise that the specified transfer (identified by the
"completed" variable) completes, but it only promises to process one
asynchronous event from any transfer in flight.

Before this fix, it could've happened that we returned to the caller
before the transfer actually finishes.


Fabian-Sommer (approved):
LGTM

PR #465, opened 2021-11-26
[usb] Fix missing transfer completion event

emaxx-google:
Make sure that when multiple event loops are running, the loop that was
interested in a particular transfer (via the associated "completed"
variable) exits immediately after that transfer finishes. The
problematic scenario is when there are two threads, both running an
event loop, and one thread populates the result of the transfer that the
other thread was waiting for. When a thread misses a notification like
in that case, it'll wait until the whole transfer timeout passes before
noticing that the "completed" variable is populated.

To fix that, we trigger one additional conditional variable
notification immediately after the transfer is populated.

This bug was likely present in our code since ever, but it became more
visible once we switched synchronous transfers to asynchronous ones
(in #458).


Fabian-Sommer (approved):
LGTM

PR #466, opened 2021-11-26
[usb] Implement transfer timeouts in proxy

emaxx-google:
Add implementation of USB transfer timeouts into the libusb-to-JS
proxy. Two main aspects of this are: (1) waiting for the transfer result
shouldn't exceed the timeout, and (2) data that's read after an input
transfer is timed out should be rerouted to later transfers.

The goal is to stop relying on the JavaScript counterpart to enforce the
timeouts, since, unlike chrome.usb, WebUSB doesn't support timeouts.

Luckily, "2" was already implemented by us in the past (for the
cancellation to work), so implementing timeouts in this commit isn't a
radical change.

### DIFF
--- a/third_party/libusb/webport/src/libusb_opaque_types.cc
+++ b/third_party/libusb/webport/src/libusb_opaque_types.cc
@@ -62,9 +62,9 @@ void libusb_context::WaitAndProcessAsyncTransferReceivedResults(
       const std::chrono::time_point<std::chrono::high_resolution_clock>
           min_transfer_timeout = GetMinTransferTimeout();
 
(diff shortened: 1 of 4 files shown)


Fabian-Sommer (approved):
LGTM % nit - and also, I would cut the "ourselves" from the CL title - after all, almost all CLs land some code that we did ourselves :)

emaxx-google:
> LGTM % nit - and also, I would cut the "ourselves" from the CL title - after all, almost all CLs land some code that we did ourselves :)

Sure, done. And thanks for the fast reviews! Really appreciate that.

PR #467, opened 2021-11-26
[ux] Improve heuristic for spurious USB failures

emaxx-google:
Improve the heuristic that we use for hiding failed devices from the UI.

The previous one - based on interface numbers - turned out to not work
for some popular devices like YubiKeys, so we ended up displaying two
items for them: a successfully initialized and a failed device, because
it's a composite device for which only the third interface is smart card
related.

The new heuristic is based on display names: if we're trying to show
multiple entries that have the same display name but some are failed and
others succeeded, then the failures are likely spurious.

For posterity, the last time we touched this heuristic was #362.

### DIFF
--- a/smart_card_connector_app/src/window-devices-displaying.js
+++ b/smart_card_connector_app/src/window-devices-displaying.js
@@ -72,8 +72,42 @@ const addDeviceElement =
  * @param {!Array.<!GSC.PcscLiteServer.ReaderInfo>} readers
  */
 function onReadersChanged(readers) {
-  displayReaderList(readers);
-  updateAddDeviceButtonText(readers.length);
+  const filteredReaders = removeSpuriousFailures(readers);
+  displayReaderList(filteredReaders);
+  updateAddDeviceButtonText(filteredReaders.length);
+}
+
+/**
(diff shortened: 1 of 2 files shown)


smart_card_connector_app/src/window-devices-displaying.js
```diff
@@ -72,8 +72,42 @@ const addDeviceElement =
...
+  /** @type {!Array<!GSC.PcscLiteServer.ReaderInfo>} */
+  const filteredReaders = readers.filter(reader => {
+    if (reader['status'] !== GSC.PcscLiteServer.ReaderStatus.FAILURE)
+      return true;
+    if (!nonFailedDisplayNames.has(
+            makeReaderNameForDisplaying(reader['name']))) {
+      return true;
+    }
+    goog.log.info(logger, `Hiding spuriously failed reader ${reader['name']}`);
+    return false;
```
Fabian-Sommer, line 108: Optional nit: I think this gets more readable by going 
if (nonFailedDisplayNames.has(...)) {
   log('Hiding failed reader'...)
   return false;
}
return true;

Fabian-Sommer (approved):
LGTM with optional nit

PR #469, opened 2021-11-26
[usb] Fix output transfer length overflow

emaxx-google:
Use 64-bit integers for storing bulk/interrupt transfer length. The
length we used before - 16 bits - is only sufficient for control
transfers, and we incorrectly used it for all types of transfers,
causing the length to be clamped due to integer overflow.

### DIFF
--- a/third_party/libusb/webport/src/libusb_js_proxy_data_model.h
+++ b/third_party/libusb/webport/src/libusb_js_proxy_data_model.h
@@ -135,7 +135,7 @@ struct LibusbJsGenericTransferParameters {
   // Only set for output transfers.
   optional<std::vector<uint8_t>> data_to_send;
   // Only set for input transfers.
-  optional<uint16_t> length_to_receive;
+  optional<int64_t> length_to_receive;
 };
 
 struct LibusbJsTransferResult {


third_party/libusb/webport/src/libusb_js_proxy_data_model.h
```diff
@@ -135,7 +135,7 @@ struct LibusbJsGenericTransferParameters {
   // Only set for output transfers.
   optional<std::vector<uint8_t>> data_to_send;
   // Only set for input transfers.
-  optional<uint16_t> length_to_receive;
+  optional<int64_t> length_to_receive;
```
emaxx-google, line 138: Compiler didn't warn about the integer clamping here... Thinking about ways of preventing this in the future, there's
(a) use int64_t everywhere - I don't like it since for all input structures we'll have the same problem when converting them back into C structures;
(b) add a helper like `AssignWithBoundsCheck`, which would be a template function that takes destination point and source value;
... any other options?..

TJ4Urservices:
Wow what amazing work you all did. Does this mean you can prove i identify of my stalker and how he has been harassing me since Aug 16th 2021

TJ4Urservices:
Secondly i am not sure about policy here is the a way that can communicate in a more messenger form or talking on the phone because I feel like I'm not understanding what I need to do I need to set up a brand new phone and make sure I'm not going to open the door for him to get in

7326189882

third_party/libusb/webport/src/libusb_js_proxy_data_model.h
Fabian-Sommer: No strong opinion on my side

Fabian-Sommer (approved):
LGTM

third_party/libusb/webport/src/libusb_js_proxy_data_model.h
emaxx-google: Actually, maybe a static assert would be a more readable alternative. Let me try that out (in a separate PR).

PR #471, opened 2021-11-29
[wasm] Fix Emscripten load failure handling

emaxx-google:
Fix a corner case that could happen when the Emscripten code fails to
load. We were raising an unnecessary error in that case, which would
slightly pollute logs:

  TypeError: Cannot read properties of null (reading 'delete')
      at GoogleSmartCard.EmscriptenModule.disposeInternal

The fix is to only call the 'delete' method when the
googleSmartCardModule_ member is already non-null. This will protect
against cases when the load failure occurs before this member is
assigned.


Fabian-Sommer (approved):
LGTM

PR #472, opened 2021-11-29
[usb] Mark LibusbToJsApiAdaptor as abstract

emaxx-google:
All subclasses of the LibusbToJsApiAdaptor JS class should implement all
methods, hence mark it as "@abstract" in order to emit compilation
warnings when this is violated.

Temporarily add empty method implementations into the
LibusbToWebusbAdaptor JS class, in order to silence the warnings until
we finish implementing the class.

This is a pure refactoring change. It's part of the WebUSB support
effort tracked by #429.


Fabian-Sommer (approved):
LGTM

PR #473, opened 2021-11-29
[usb] Port loginState hook to new adaptor

emaxx-google:
Port ChromeLoginStateHook to the new libusb-to_JS adaptor, so that
Libusb doesn't see any devices when the devices is locked. For this, add
an abstract definition for the hook class and implement the new
loginState hook via it. Thanks to the better libusb-to-JS API and
extensive use of promises, the hook is now implemented in a more
type-check-friendly way.

### DIFF
--- a/third_party/libusb/webport/src/chrome_usb/chrome-login-state-hook.js
+++ b/third_party/libusb/webport/src/chrome_usb/chrome-login-state-hook.js
@@ -17,149 +17,100 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
  */
 
-goog.provide('GoogleSmartCard.Libusb.ChromeLoginStateHook');
+goog.provide('GoogleSmartCard.LibusbLoginStateHook');
 
-goog.require('GoogleSmartCard.Libusb.ChromeUsbBackend');
-goog.require('GoogleSmartCard.DebugDump');
+goog.require('GoogleSmartCard.LibusbProxyHook');
 goog.require('GoogleSmartCard.Logging');
-goog.require('GoogleSmartCard.RemoteCallMessage');
-goog.require('GoogleSmartCard.RequestReceiver');
-goog.require('goog.Promise');
 goog.require('goog.asserts');
-goog.require('goog.array');
-goog.require('goog.Disposable');
-goog.require('goog.iter');
 goog.require('goog.log');
-goog.require('goog.log.Logger');
-goog.require('goog.messaging.AbstractChannel');
-goog.require('goog.object');
-goog.require('goog.promise.Resolver');
 
 goog.scope(function() {
 
 const GSC = GoogleSmartCard;
 
-/**
- * This class implements a hook that hides USB devices when the app is running
- * for a user who has locked their screen. This causes PCSC Lite to release all
- * owned USB devices so that they can be accessed from the sign-in profile app.
- * @extends goog.Disposable
- * @constructor
- */
-GSC.Libusb.ChromeLoginStateHook = function() {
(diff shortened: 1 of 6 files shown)


third_party/libusb/webport/src/chrome_usb/chrome-login-state-hook.js
```diff
@@ -17,149 +17,100 @@
...
-ChromeLoginStateHook.prototype.logger =
-    GSC.Logging.getScopedLogger('Libusb.LoginStateHook');
+const logger = GSC.Logging.getScopedLogger('LibusbLoginStateHook');
 
-/** @override */
-ChromeLoginStateHook.prototype.disposeInternal = function() {
-  if (chrome.loginState) {
-    chrome.loginState.onSessionStateChanged.removeListener(
-        this.boundOnGotSessionState_);
+// TODO(#429): Add event unsubscription.
```
emaxx-google, line 33: I haven't ported the "dispose" method yet. Let me see how simple it is to add into the current PR...

third_party/libusb/webport/src/libusb-proxy-receiver.js
```diff
@@ -62,7 +64,30 @@ GSC.LibusbProxyReceiver = class {
...
+    this.hooks_.push(hook);
+  }
+
+  /**
+   * @return {!GSC.LibusbToJsApiAdaptor}
+   * @private
+   */
+  getLibusbToJsApiAdaptor_() {
+    if (this.hooks_.length)
+      return this.hooks_[this.hooks_.length - 1];
```
Fabian-Sommer, line 89: Starting with Chrome 92, we could also use this.hooks_.at(-1)
Though I guess we need to care about much older Chrome versions than that...

third_party/libusb/webport/src/chrome_usb/chrome-login-state-hook.js
```diff
@@ -17,149 +17,100 @@
...
+      this.resolveInitializationPromise_ = resolve;
+    });
+    if (chrome && chrome.loginState) {
+      chrome.loginState.getProfileType(this.onGotProfileType_.bind(this));
+    } else {
+      goog.log.warning(
+          logger,
+          'chrome.loginState API is not available. This app might require a ' +
+              'newer version of Chrome.');
+      this.resolveInitializationPromise_();
```
Fabian-Sommer, line 54: I may misunderstand this, but isn't resolveInitializationPromise_ null until initializationPromise_ is called/awaited, which probably hasn't happened here?

Fabian-Sommer (approved):
LGTM % question

third_party/libusb/webport/src/libusb-proxy-receiver.js
emaxx-google: Exactly - currently we claim the compatibility with Chrome >=48: https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/blob/6d8ae8d0da3eac0b9f25ffb38b06d2a37649face/smart_card_connector_app/src/manifest.json.template#L29

Also if we were to drop it, jumping directly to 92 might be too aggressive, given a long tail of devices that stay on outdated OS versions...

third_party/libusb/webport/src/chrome_usb/chrome-login-state-hook.js
emaxx-google: The `Promise` constructor executes the callback synchronously, so we do have the `resolveInitializationPromise_` field initialized already at this point. I know this looks weird, and that's what `goog.Promise.withResolver()` was improving - but they recommend to switch away from `goog.Promise` these days...

PR #474, opened 2021-11-29
Build JS unit tests for the Connector application

emaxx-google:
Add build script boilerplate for compiling and running JavaScript
unit tests from the //smart_card_connector/src/ directory.

Currently there are no JS unit tests there, but they will be added in
follow-up commits.


Fabian-Sommer (approved):
LGTM

PR #475, opened 2021-11-29
[usb] Compile-time checks for struct field sizes

emaxx-google:
Add compile-time assertions into some of the C/C++ code that performs
data conversion to/from libusb structs and to/from libusb-to-JS adaptor
structs. This should prevent bugs with accidental data clamping, similar
to #469.

Also make some of the libusb-to-JS structs have a more precise sizing,
as a few fields were violating the checks added by this commit.

This commit is part of the WebUSB implementation effort, tracked
by #429.

### DIFF
--- a/common/cpp/src/google_smart_card_common/numeric_conversions.h
+++ b/common/cpp/src/google_smart_card_common/numeric_conversions.h
@@ -14,8 +14,6 @@
 
 // This file contains low-level operations for conversions between numeric data
 // types.
-//
-// FIXME(emaxx): Use boost::numeric?
 
 #ifndef GOOGLE_SMART_CARD_COMMON_NUMERIC_CONVERSIONS_H_
 #define GOOGLE_SMART_CARD_COMMON_NUMERIC_CONVERSIONS_H_
@@ -28,6 +26,7 @@
 #include <type_traits>
 
 #include <google_smart_card_common/formatting.h>
+#include <google_smart_card_common/logging/logging.h>
 
 namespace google_smart_card {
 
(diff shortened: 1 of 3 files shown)


ayaNader (approved):
LGTM

Fabian-Sommer (commented):
LGTM (I would have had a nit with one of the error messages, but this is a really small nit)

emaxx-google:
> LGTM (I would have had a nit with one of the error messages, but this is a really small nit)

Please comment - I anyway found a forgotten #include and need a follow-up :)

common/cpp/src/google_smart_card_common/numeric_conversions.h
```diff
@@ -135,6 +134,32 @@ inline bool CastIntegerToDouble(T value,
...
+  GOOGLE_SMART_CARD_CHECK(lhs);
+  static_assert(std::is_integral<FirstType>::value,
+                "This assertion is only applicable to integral types");
+  static_assert(std::is_integral<SecondType>::value,
+                "This assertion is only applicable to integral types");
+  static_assert(sizeof(FirstType) > sizeof(SecondType) ||
+                    (sizeof(FirstType) == sizeof(SecondType) &&
+                     std::is_signed<FirstType>::value ==
+                         std::is_signed<SecondType>::value),
+                "Left-hand-side type is smaller than the right-hand-side type");
```
Fabian-Sommer, line 150: nit: This error does not match the std::is_signed<FirstType>::value == std::is_signed<SecondType>::value check - I think we want a separate error message / assert for that

emaxx-google: Splitting static_assert conditions is somewhat hard, since all static_asserts are always executed (so you can't express "do this static_assert only if some other condition is true", at least not in C++11; "if constexpr" in C++17 could probably help us once we drop NaCl support).

The closest I can come up with is:
```
  static_assert(sizeof(FirstType) >= sizeof(SecondType), ...);
  static_assert(sizeof(FirstType) != sizeof(SecondType) ||
                         std::is_signed<FirstType>::value == std::is_signed<SecondType>::value,
                        ...);
```

emaxx-google: Published #478 - PTAL. Thanks.

PR #476, opened 2021-11-29
[wasm] Use content security policy in extension

emaxx-google:
This commit adds a content_security_policy key into manifest.json file
when the EXTENSION packaging mode is used. This allows the WebAssembly
module to be loaded in that mode, since by default the extension CSP
forbids this.

Note that this is not required in the APP packaging mode, since App's
default CSP is not so strict:
https://source.chromium.org/chromium/chromium/src/+/main:extensions/common/manifest_handlers/csp_info.cc;l=61;drc=1e0c2a218eb71bef788fd87b0e7a4e9e93ecfe1d


Fabian-Sommer (commented):
LGTM

PR #477, opened 2021-11-29
[wasm] Specify COOP/COEP when using WebAssembly

emaxx-google:
Add cross_origin_embedder_policy and cross_origin_opener_policy
properties into manifest.json, which allow to configure the extension's
Cross-Origin-Opener-Policy and Cross-Origin-Embedder-Policy
correspondingly.

These settings will be required by new versions of Chrome for extensions
that use WebAssembly with multi-threading. See, for example:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer/Planned_changes

Normally these settings are configured via HTTP headers, but inside the
extension this isn't possible. But the Chrome Extensions team added
special manifest entries that allow to achieve the same COOP/COEP
configuration as if it was done via headers:
https://groups.google.com/u/0/a/chromium.org/g/chromium-extensions/c/Auo0HwzsyG0/m/TIpqFyIaDAAJ


ayaNader (approved):
LGTM

emaxx-google:
I'm going to temporarily revert this because it causes multiple issues: http://crbug.com/1275140, http://crbug.com/1275145. Let's see how these discussions go and how/when we'll be able to reland this.

PR #480, opened 2021-11-30
[common] Fix circular ref handling in JS debugDump

emaxx-google:
Make the JavaScript DebugDump helpers robust against circular
references in the input data. Achieve this by remembering the set
of already visited objects during the depth-first-search-like
traversal.

Previously, we only had an ad-hoc protection against such issues, by
checking for a few well-known types that typically contain circular
references, like the built-in Document type. Another ad-hoc protection
was catching and suppressing all exceptions thrown during the
traversal in the Release. The current commit doesn't delete these, since
the former makes logs more readable and the latter is a useful safety
belt.

### DIFF
--- a/common/js/src/logging/debug-dump-unittest.js
+++ b/common/js/src/logging/debug-dump-unittest.js
@@ -67,27 +67,45 @@ goog.exportSymbol('testDebugDump', function() {
   assertEquals('<HTMLCollection>', dump(document.body.children));
 });
 
+// Test the `dump()` method against container types with circular references: it
+// should detect the loops and avoid infinite recursion.
 goog.exportSymbol('testDebugDumpWithCircularRefs', function() {
   const circularInsideArray = [];
   circularInsideArray.push(circularInsideArray);
-  assertEquals('[<duplicate>]', dump(circularInsideArray));
+  assertEquals('[<circular>]', dump(circularInsideArray));
 
   const circularInObject = {'a': 'b'};
   circularInObject['c'] = circularInObject;
-  assertEquals('{"a": "b", "c": <duplicate>}', dump(circularInObject));
+  assertEquals('{"a": "b", "c": <circular>}', dump(circularInObject));
 
   const circularInMap = new Map();
   circularInMap.set('a', 'b');
   circularInMap.set('c', circularInMap);
-  assertEquals('Map{"a": "b", "c": <duplicate>}', dump(circularInMap));
+  assertEquals('Map{"a": "b", "c": <circular>}', dump(circularInMap));
 
   const circularInSet = new Set();
   circularInSet.add('foo');
   circularInSet.add(circularInSet);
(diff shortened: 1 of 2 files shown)


Fabian-Sommer (commented):
Perhaps I misunderstand how JS passes parameters, but I think this solution is overly zealous in that it will never dump an object multiple times, even when it is not part of some cycle.
If I do misunderstand this, could we add a test like:
randomObject = {1,2,3}
const array = [randomObject, randomObject]
assertEquals(...) (<- this should print properly, and not skip the second time with <duplicate>)

emaxx-google:
> Perhaps I misunderstand how JS passes parameters, but I think this solution is overly zealous in that it will never dump an object multiple times, even when it is not part of some cycle. If I do misunderstand this, could we add a test like: randomObject = {1,2,3} const array = [randomObject, randomObject] assertEquals(...) (<- this should print properly, and not skip the second time with )

That's correct. That's why I'm printing "<duplicate>" and not "<circular>". Do you see a better alternative though?

BTW, Mozilla shows this as an intended use case for the `WeakSet`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet#use_case_detecting_circular_references

emaxx-google:
> > Perhaps I misunderstand how JS passes parameters, but I think this solution is overly zealous in that it will never dump an object multiple times, even when it is not part of some cycle. If I do misunderstand this, could we add a test like: randomObject = {1,2,3} const array = [randomObject, randomObject] assertEquals(...) (<- this should print properly, and not skip the second time with )
> 
> That's correct. That's why I'm printing `<duplicate>` and not `<circular>`. Do you see a better alternative though?
> 
> BTW, Mozilla shows this as an intended use case for the `WeakSet`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet#use_case_detecting_circular_references

Actually, updated the PR with a new approach. It uses the standard DFS idea for detecting loops; we don't have a real DFS here, since we intentionally want to visit duplicates multiple times, but I think the loop detection should still work fine.

Fabian-Sommer (approved):
LGTM - This is what I had in mind

common/js/src/logging/debug-dump-unittest.js
```diff
@@ -66,4 +66,46 @@ goog.exportSymbol('testDebugDump', function() {
...
+  circularInSet.add('foo');
+  circularInSet.add(circularInSet);
+  assertEquals('Set{"foo", <circular>}', dump(circularInSet));
+
+  const nestedCircular = {};
+  nestedCircular['foo'] = {'bar': nestedCircular};
+  assertEquals('{"foo": {"bar": <circular>}}', dump(nestedCircular));
+});
+
+// Test that the `dump()` method correct handles the case when an object is
```
ayaNader, line 96: nit: correctly*

ayaNader (approved):
LGTM % nit

PR #481, opened 2021-11-30
[common] Optimize ArrayBuffer debug dumping

emaxx-google:
Optimize the JavaScript DebugDump helper that deals with ArrayBuffers
and make the output format more compact (use a single "0x" instead of
prepending "0x" before every byte and using the ", " separator).

The previous implementation was doing lots of overhead by passing every
byte through a bunch of type and boundary checks. All of that is
unnecessary, as we know that after the type casting we're dealing with
an array of unsigned single-byte integers.

We don't have any measurements, but the inefficiency of the old version
could actually be user-visible, since we have a lot of log statements
that are calculated even in the Release mode (unlike when using C/C++,
it's not possible to automatically completely skip log message
calculation in JavaScript, even when we know that we're below the
current log level threshold).

### DIFF
--- a/common/js/src/logging/debug-dump-unittest.js
+++ b/common/js/src/logging/debug-dump-unittest.js
@@ -47,8 +47,7 @@ goog.exportSymbol('testDebugDump', function() {
   assertEquals('[0x01, "foo", [0x02, 0x03], []]', dump([1, 'foo', [2, 3], []]));
 
   assertEquals('ArrayBuffer[]', dump(new ArrayBuffer(0)));
-  assertEquals(
-      'ArrayBuffer[0x01FF]', dump(new Uint8Array([1, 255]).buffer));
+  assertEquals('ArrayBuffer[0x01FF]', dump(new Uint8Array([1, 255]).buffer));
 
   assertEquals(
       'Map{0x01: 0x02, 0x03: "foo", "bar": {}, {}: 0x04}',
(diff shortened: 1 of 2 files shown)


Fabian-Sommer (approved):
LGTM - Personally I think that using 4 characters of clutter (", 0x") for every 2 characters of content is a bit much, especially for longer ArrayBuffers (what about 0x0123456789ABCDEF instead, with 0x and ", " only once every 8 bytes?), but that is just personal preference

emaxx-google:
> LGTM - Personally I think that using 4 characters of clutter (", 0x") for every 2 characters of content is a bit much, especially for longer ArrayBuffers (what about 0x0123456789ABCDEF instead, with 0x and ", " only once every 8 bytes?), but that is just personal preference

That's fair. I've implemented just a single "0x", since grouping by 8 bytes is a bit more work to implement, and also depending on the context can make interpreting logs even harder. Does this work for you?

Fabian-Sommer:
> > LGTM - Personally I think that using 4 characters of clutter (", 0x") for every 2 characters of content is a bit much, especially for longer ArrayBuffers (what about 0x0123456789ABCDEF instead, with 0x and ", " only once every 8 bytes?), but that is just personal preference
> 
> That's fair. I've implemented just a single "0x" now, since grouping by 8 bytes is a bit more work to implement, and also depending on the context can make interpreting logs even harder. Does this work for you?

Sounds good

PR #483, opened 2021-12-01
[usb] Fetch extraData manually when using WebUSB

emaxx-google:
Polyfill the missing "extraData" information that's not returned by
WebUSB but is part of the libusb interface and is used by the CCID
driver. Do this by making a special "GET_DESCRIPTOR" control transfer to
the device and parsing the bytes out of it.

For the background, each USB device exposes a hierarchy of descriptors
(basically, metadata to be used by the host): configuration descriptors
==> interface descriptors ==> endpoint descriptors. Additionally there
might be class-/vendor-specific descriptors, like 0x21 is a smart card
reader specific descriptor type, and it contains stuff like whether
there's a PIN support in the reader. Such class-/vendor-specific
descriptors aren't returned by WebUSB currently, but the libusb library
that we're trying to emulate does expose it as field "extra_data",
and the CCID driver we're using does heavily rely on this data to be
present.

This polyfill wasn't needed in the past, because the chrome.usb API
returned extraData directly.

### DIFF
--- a/third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
+++ b/third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
@@ -380,6 +380,7 @@ async function fetchAndFillConfigurationExtraDataForOpenedDevice(
     const descriptorLength = descriptors.getUint8(offset);
     if (descriptorLength < 2) {
       // Ignore the invalid (too short) descriptor.
+      offset += descriptorLength;
       continue;
     }
     if (offset + descriptorLength > totalLength) {
--- a/third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
+++ b/third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
@@ -21,6 +21,7 @@ goog.provide('GoogleSmartCard.LibusbToWebusbAdaptor');
 
 goog.require('GoogleSmartCard.Logging');
 goog.require('goog.asserts');
+goog.require('goog.log');
 goog.require('goog.object');
 
 goog.scope(function() {
@@ -94,6 +95,8 @@ GSC.LibusbToWebusbAdaptor = class extends GSC.LibusbToJsApiAdaptor {
             webusbDevice, libusbJsConfiguration);
       } catch (exc) {
         // Suppress errors of fetching extraData: they should be non-fatal.
+        // Log only at the "fine" level, as this can quickly flood the logs.
+        goog.log.fine(logger, `Failure fetching extra data: ${exc}`);
       }
       libusbJsConfigurations.push(libusbJsConfiguration);
     }
--- a/third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
+++ b/third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
@@ -441,7 +441,8 @@ async function fetchAndFillConfigurationExtraDataForOpenedDevice(
       default: {
         // The current item is an unknown descriptor, so add it as extraData to
         // the current (the most nested one) LibusbJs object.
-        const targetObject = currentEndpoint || currentInterface || libusbJsConfiguration;
+        const targetObject =
+            currentEndpoint || currentInterface || libusbJsConfiguration;
         if (targetObject) {
           appendExtraData(
               new Uint8Array(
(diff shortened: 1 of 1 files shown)


emaxx-google:
PTAL; thanks!
I'll also ask Reilly, as the USB expert, to take a look at this as well.

third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
```diff
@@ -292,6 +303,160 @@ function getLibusbJsEndpointType(webusbEndpointType) {
...
+  // interface/endpoint descriptor.
+  /** @type {!LibusbJsInterfaceDescriptor|undefined} */
+  let currentInterface = undefined;
+  /** @type {!LibusbJsEndpointDescriptor|undefined} */
+  let currentEndpoint = undefined;
+  for (let offset = 0; offset < totalLength;) {
+    const descriptorLength = descriptors.getUint8(offset);
+    if (descriptorLength < 2) {
+      // Ignore the invalid (too short) descriptor.
+      break;
```
emaxx-google, line 380: This should probably be `continue`.

TJ4Urservices: Guys The Amazing work  you have no idea how much you have helped

TJ4Urservices: how can I give you access my trace Logs

TJ4Urservices:
That has to be the back / Door Though Google calendar  it's late sometimes he's playing with me he'll give me access to my account but then take away certain parts of it and it just keeps going on and on so the only thing I've been able two Email accounts is By Creating Websites

third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
```diff
@@ -292,6 +306,169 @@ function getLibusbJsEndpointType(webusbEndpointType) {
...
+  // interface/endpoint descriptor.
+  /** @type {!LibusbJsInterfaceDescriptor|undefined} */
+  let currentInterface = undefined;
+  /** @type {!LibusbJsEndpointDescriptor|undefined} */
+  let currentEndpoint = undefined;
+  for (let offset = 0; offset < totalLength;) {
+    const descriptorLength = descriptors.getUint8(offset);
+    if (descriptorLength < 2) {
+      // Ignore the invalid (too short) descriptor.
+      continue;
```
Fabian-Sommer, line 383: I think we need to update offset here, otherwise we will get stuck in a loop

Fabian-Sommer (approved):
LGTM otherwise, but I can't really review details of the interface.

third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
emaxx-google: Great catch! Thanks, fixed.
And I'll add this case to the unit tests that I'm planning to add in a follow-up.

reillyeon: This will also get stuck in a loop if descriptorLength is 0. We can reject the configuration descriptor as invalid if the length is less than 2.

third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
```diff
@@ -292,6 +306,170 @@ function getLibusbJsEndpointType(webusbEndpointType) {
...
+          currentEndpoint = undefined;
+        }
+        break;
+      }
+      default: {
+        // The current item is an unknown descriptor, so add it as extraData to
+        // the current (the most nested one) LibusbJs object.
+        const targetObject = currentEndpoint ?
+            currentEndpoint :
+            (currentInterface ? currentInterface : libusbJsConfiguration);
```
reillyeon, line 440: I think JavaScript will let you write this as,

```suggestion
        const targetObject = currentEndpoint || currentInterface || libusbJsConfiguration;
```

third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
emaxx-google: Another great catch! Done.

third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
emaxx-google: Done, thank you.

reillyeon (approved):

emaxx-google:
Thanks!

PR #484, opened 2021-12-01
[usb] Add debug-only logging of some WebUSB errors

emaxx-google:
Log some WebUSB-related errors at the "fine" level, so that these errors
are surfaced in the Debug mode, but don't flood the application's logs
in production.


Fabian-Sommer (approved):
LGTM

PR #485, opened 2021-12-01
[usb] Implement {open,close}DeviceHandle in WebUSB

emaxx-google:
Implement openDeviceHandle()/closeDeviceHandle() requests in the
libusb-to-WebUSB adaptor. As WebUSB doesn't have a concept of device
handles, but Libusb does, our adaptor is generating them using a simple
counter. The main difficulty is to handle the cases of concurrent
open/close requests, as WebUSB doesn't like concurrent open/close calls,
and we want to correctly handle scenarios when an open request comes
before the close finishes.


emaxx-google:
Maybe I should ask Reilly to look at this as well...

third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
```diff
@@ -88,14 +111,50 @@ GSC.LibusbToWebusbAdaptor = class extends GSC.LibusbToJsApiAdaptor {
...
+      // Avoid concurrent WebUSB open()/close() calls, since WebUSB throws on
+      // these, hence track the ongoing request's promise in the device state
+      // and wait for it.
+      if (!deviceState.openOperationPromise) {
+        // Note: no "await" here, since we need to set `openOperationPromise`
+        // synchronously and therefore let all subsequent requests wait on it.
+        deviceState.openOperationPromise = openWebusbDevice(deviceState);
+      }
+      await deviceState.openOperationPromise;
+      deviceState.openOperationPromise = null;
```
emaxx-google, line 128: It's a question whether this should be wrapped inside "finally" section, so that one open() failure doesn't cause all subsequent ones to fail as well (especially those that are started much later).

emaxx-google: (did that)

third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
```diff
@@ -88,14 +111,50 @@ GSC.LibusbToWebusbAdaptor = class extends GSC.LibusbToJsApiAdaptor {
...
+    }
+    // WebUSB successfully opened the device. Generate a new handle and return.
+    // Note: It's important to update `handles` only after all asynchronous
+    // steps are completed, so that concurrent `openDeviceHandle()` calls wait
+    // for the promises as well.
+    const newHandle = this.nextFreeDeviceHandle_;
+    ++this.nextFreeDeviceHandle_;
```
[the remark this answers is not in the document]
emaxx-google: JS is single-threaded, so there should be no race conditions here. We should only take care about task-level granularity, like when one call is doing "await" there might be another call made.

emaxx-google:
> Maybe I should ask Reilly to look at this as well...

Actually I think there's no strictly USB-specific question in this pull request, it's more like solving a generic programming problem and expressing it in terms of JavaScript. So our internal review and manual testing should be sufficient.

Fabian-Sommer (approved):
LGTM

emaxx-google:
Thanks!

PR #486, opened 2021-12-01
[common] Fix boolean debug dumping in JS

emaxx-google:
Fix the regression introduced by #481: the dump() function was throwing
this error when called with a boolean:

  TypeError: Invalid value used in weak set

The reason is that WeakSet, used by us in order to detect circular
references, doesn't support non-objects. And also that we also never
handle the boolean value explicitly - it was dumped by the fallback JSON
serialization (which might've been a small performance issue itself).

This commit fixes both aspects: it dumps booleans explicitly, but also
adds a safety fallback for non-objects before triggering WeakSet
operations.

### DIFF
--- a/common/js/src/logging/debug-dump-unittest.js
+++ b/common/js/src/logging/debug-dump-unittest.js
@@ -32,6 +32,9 @@ goog.exportSymbol('testDebugDump', function() {
 
   assertEquals('null', dump(null));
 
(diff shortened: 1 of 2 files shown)


emaxx-google:
Apologies for that small thing now needing a follow-up...

ayaNader (approved):
LGTM

PR #488, opened 2021-12-01
[wasm] Support DataView-to-Value conversion

emaxx-google:
Add support for conversions from the JavaScript DataView type into the
C++ Value class. The DataView was one of types that we intentionally
skipped from the original emscripten-to-Value converter, because this
type seemed too exotic to be used in our project. However, it turned out
that WebUSB uses DataView a lot, so instead of doing artificial
conversions at the WebUSB adaptor layer we're filling the gap in the
emscripten-to-Value converter.


Fabian-Sommer (approved):
LGTM

PR #492, opened 2021-12-02
[common] Skip the "chrome" global from dumping

emaxx-google:
Bail out from JavaScript debugDump when the "chrome" global object is
passed.

If some JS code accidentally passes the "chrome" object to the debugDump
function, the function was attempting to recursively dump its properties.
The resulting dump isn't of any use (it'll contain a list of Extensions
API functions), so it'd be a waste of resources, and additionally it'll
trigger console warnings due to accessing deprecated functions. Hence
handle the case of the "chrome" global object explicitly, even if no
code should normally pass it.


emaxx-google:
@ayaNader this should improve the handling of the you observed, even though I couldn't repro it locally, and it's bizarre that we're passing `chrome` to debug dumping somewhere.

Fabian-Sommer (approved):
LGTM

PR #493, opened 2021-12-03
[usb] Log WebUSB calls

emaxx-google:
Log calls of WebUSB methods. Despite that we're already logging USB
commands at the libusb layer, this commit also adds logging at the place
we're calling into WebUSB, since there's some nontrivial functionality
between these two ends.

### DIFF
--- a/third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
+++ b/third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
@@ -44,8 +44,15 @@ const LibusbJsTransferRecipient =
 const LibusbJsTransferRequestType =
     GSC.LibusbProxyDataModel.LibusbJsTransferRequestType;
 const LibusbJsTransferResult = GSC.LibusbProxyDataModel.LibusbJsTransferResult;
+const debugDump = GSC.DebugDump.debugDump;
+const dump = GSC.DebugDump.dump;
 
 const logger = GSC.Logging.getScopedLogger('LibusbToWebusbAdaptor');
+/**
+ * The counter used for numbering transfer requests in logs.
+ * @type {number}
+ */
+let transferLogCounter = 1;
 
 class DeviceState {
   /** @param {!Object} webusbDevice The WebUSB USBDevice object. */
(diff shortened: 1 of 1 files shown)


emaxx-google:
The implementation is wordy, but I haven't found a way to make it more compact, given how we use different log levels and use different dump mechanisms for input/output arguments.

third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
```diff
@@ -363,16 +375,279 @@ GSC.LibusbToWebusbAdaptor = class extends GSC.LibusbToJsApiAdaptor {
...
+      debugDump(inTransferResult['data'])}`;
+}
+
+/**
+ * @param {!Object} outTransferResult The WebUSB USBOutTransferResult value.
+ * @return {string}
+ */
+function dumpWebusbOutTransferResultForLog(outTransferResult) {
+  return `status=${outTransferResult['status']} bytesWritten=${
+      outTransferResult['bytesWritten']}`;
```
Fabian-Sommer, line 529: This is fine to log at all times?

Fabian-Sommer (approved):
LGTM

third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
emaxx-google: Status is definitely safe - it's one of "ok"/"stall"/"babble". With bytesWritten, I was on the fence myself (is it security-sensitive?), but then noticed that we've been logging the transfer data length since ever:
https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/blob/0a88488ba506f32fd14ed1b38102411b4508f488/third_party/pcsc-lite/naclport/common/src/google_smart_card_pcsc_lite_common/scard_debug_dump.cc#L324

So from the security/privacy perspective we're retaining the status quo. Are there any considerations you had in mind?

Fabian-Sommer: Logging ok/stall/babble seems fine.
Logging the transfer length is also fine (but it is not entirely clear from the naming that bytesWritten is an int, and not the actual bytes that were written)

emaxx-google: Oh I see. Let me add a comment.

PR #494, opened 2021-12-03
[pcsc] Make exceptions in listener not break logic

emaxx-google:
Wrap calls of some of listeners into try-catch, so that even when some
of them throw error it doesn't prevent others from being notified and,
more importantly, it doesn't bubble the exception throw the whole stack.

A real example where this helps is when the PC/SC server notifies that
the list of readers changes. This notification triggers the UI update,
which can throw an exception when the UI is already closed (while this
can be addressed separately, it's still an example of something we want
to defend against). If we don't catch the exception, it bubbles back
into the PC/SC-Lite server code, causing the reader initialization to be
aborted.


Fabian-Sommer (approved):
LGTM

PR #495, opened 2021-12-03
[usb] Device selector dialog in WebUSB

emaxx-google:
Add implementation for prompting the user to grant a permission in the
configuration when WebUSB is used.

Due to the WebUSB limitations, we can't show the prompt directly after
the button click in the browser action, therefore we're opening a new
tab with the same window, letting the user to click the button again and
select the device. This is a bad UX, but it's a temporary measure until
Chrome is changed to allowlist the Smart Card Connector for accessing
all smart card readers.

### DIFF
--- a/smart_card_connector_app/src/window-devices-displaying.js
+++ b/smart_card_connector_app/src/window-devices-displaying.js
@@ -40,10 +40,10 @@ goog.require('goog.string');
 goog.scope(function() {
 
 /**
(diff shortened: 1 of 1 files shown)


emaxx-google:
This is, finally, the last commit that lets the MVP version to work with WebUSB. Yay :)

Fabian-Sommer (approved):
LGTM

PR #496, opened 2021-12-03
[usb] Fetch WebUSB extraData via handles

emaxx-google:
Switch the polyfill that fetches extraData from direct calls to WebUSB
open()/close() methods to using device handles. This allows to make sure
that we don't try to open a device twice (if a concurrent request is
started later), and that we don't close a device that's still in use
(again by some concurrent operation).

Also while doing this, apply a small optimization: don't open-close the
device for each configuration descriptor, but do this once for all
descriptors of a single device.

### DIFF
--- a/third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
+++ b/third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
@@ -331,7 +331,7 @@ GSC.LibusbToWebusbAdaptor = class extends GSC.LibusbToJsApiAdaptor {
   }
 
   /**
-   * Fills the extraData fields of the passed configurations, by fetching and
+   * Fills the extraData fields of all passed configurations by fetching and
    * parsing USB descriptors from the device.
    * @param {number} deviceId
    * @param {!Object} webusbDevice The WebUSB USBDevice object.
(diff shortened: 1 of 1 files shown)


third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
```diff
@@ -342,6 +330,40 @@ GSC.LibusbToWebusbAdaptor = class extends GSC.LibusbToJsApiAdaptor {
     return libusbJsDevice;
   }
 
+  /**
+   * Fills the extraData fields of the passed configurations, by fetching and
```
ayaNader, line 334: nit: "...of all the passed configurations..."

This is just a personal preference to avoid confusion with the comment for the function handling a single passed configuration

ayaNader (approved):
LGTM % nit

third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
Fabian-Sommer: "of all passed configurations" - also, don't follow that by a ","

third_party/libusb/webport/src/libusb-to-webusb-adaptor.js
```diff
@@ -780,33 +802,12 @@ async function openWebusbDevice(deviceState) {
...
-    // TODO(#429): Don't close if there are opened device handles.
-    await webusbDevice['close']();
-  }
+  await callWebusbDeviceOpen(deviceState.webusbDevice);
 }
 
 /**
- * Same as `fetchAndFillConfigurationExtraData()`, but assumes the device to be
- * opened.
+ * Fills the extraData fields of the passed configuration, by fetching and
```
Fabian-Sommer, line 809: nit: no ","

Fabian-Sommer (approved):
LGTM % nits

PR #497, opened 2021-12-04
[nacl] Fix illegal offsetTop modification

emaxx-google:
Stop trying to modify the Native Client embed's "offsetTop" property,
since it triggers an exception when the code uses "use strict":

  TypeError: Cannot set property offsetTop of #<HTMLElement> which has
  only a getter

This illegal assignment was silently ignored till #443 that added some
Closure Compiler parameters that specify the input/output language
version. Apparently, these new parameters also lead to the resulting
code containing "use strict", which in turn resulted in this exception
and the breakage when compiled in NaCl mode.

The fix is to write to value to a random other property, because we
don't actually care about the result: we only want an access to the
"offsetTop" property to be not optimized away by Closure Compiler, and
this access is itself a workaround for an old v8/NaCl bug with the
module not loading until this is done.


ayaNader (approved):
LGTM

Fabian-Sommer (approved):
LGTM

PR #499, opened 2021-12-07
[wasm] Keep background page forcibly

emaxx-google:
Make sure that the Smart Card Connector's background page is never
unloaded, even when the user and other applications don't communicate
with it. This allows to make sure that the Connector is responsive when
requests arrive to it.

This change is only needed for the Emscripten/WebAssembly mode, since
when using NaCl Chrome automatically keeps the page alive.


Fabian-Sommer (approved):
LGTM

PR #500, opened 2021-12-07
Fix permission prompt's width and height

ayaNader:
This PR adjusts the width and height for the permission prompt asking user to allow third party to access the Smart Card Connector.

Previously the width was large compared to that in the app mode and the height was too small that a scroll bar was displayed...which was unnecessary since the window height overall isn't big enough to require a scroll bar.

### DIFF
--- a/common/js/src/popup-window/in-popup-main-script.js
+++ b/common/js/src/popup-window/in-popup-main-script.js
@@ -43,6 +43,8 @@ const GSC = GoogleSmartCard;
 /** @type {!goog.log.Logger} */
 const logger = GSC.Logging.getScopedLogger('PopupWindow.InPopupMainScript');
 
+const titleBarHeight = 35;
+
 /**
  * Returns the additional data that was specified during the popup creation.
  * @return {!Object}
@@ -155,7 +157,7 @@ GSC.InPopupMainScript.setWindowHeightToFitContent = function() {
     chrome.app.window.current().innerBounds.height = wholeContentHeight;
   } else if (GSC.Packaging.MODE === GSC.Packaging.Mode.EXTENSION) {
     chrome.windows.update(
-        chrome.windows.WINDOW_ID_CURRENT, {'height': wholeContentHeight});
+        chrome.windows.WINDOW_ID_CURRENT, {'height': wholeContentHeight + titleBarHeight});
   }
 };
 
--- a/common/js/src/popup-window/in-popup-main-script.js
+++ b/common/js/src/popup-window/in-popup-main-script.js
@@ -147,7 +147,6 @@ GSC.InPopupMainScript.prepareAndShowAsModalDialog = function() {
  */
 GSC.InPopupMainScript.setWindowHeightToFitContent = function() {
   const wholeContentHeight = document.documentElement['offsetHeight'];
-  console.log('ayag ' + wholeContentHeight);
   GSC.Logging.checkWithLogger(
       logger,
       wholeContentHeight !== undefined &&
--- a/common/js/src/popup-window/in-popup-main-script.js
+++ b/common/js/src/popup-window/in-popup-main-script.js
@@ -157,7 +157,7 @@ GSC.InPopupMainScript.setWindowHeightToFitContent = function() {
     chrome.app.window.current().innerBounds.height = wholeContentHeight;
   } else if (GSC.Packaging.MODE === GSC.Packaging.Mode.EXTENSION) {
     // Adding TITLE_BAR_HEIGHT as it's not included in the calculation of the
-    // offsetHeight..this is not needed in APP mode as it doesn't have a
+    // offsetHeight. This is not needed in APP mode as it doesn't have a
     // titlebar
     chrome.windows.update(
         chrome.windows.WINDOW_ID_CURRENT,
--- a/common/js/src/popup-window/in-popup-main-script.js
+++ b/common/js/src/popup-window/in-popup-main-script.js
@@ -43,7 +43,7 @@ const GSC = GoogleSmartCard;
 /** @type {!goog.log.Logger} */
 const logger = GSC.Logging.getScopedLogger('PopupWindow.InPopupMainScript');
 
-const titleBarHeight = 35;
+const TITLE_BAR_HEIGHT = 35;
 
 /**
  * Returns the additional data that was specified during the popup creation.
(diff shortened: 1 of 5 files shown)


common/js/src/popup-window/in-popup-main-script.js
```diff
@@ -43,6 +43,8 @@ const GSC = GoogleSmartCard;
 /** @type {!goog.log.Logger} */
 const logger = GSC.Logging.getScopedLogger('PopupWindow.InPopupMainScript');
 
+const titleBarHeight = 35;
```
emaxx-google, line 46: nit: `TITLE_BAR_HEIGHT`
(See https://google.github.io/styleguide/jsguide.html#naming-constant-names)

common/js/src/popup-window/in-popup-main-script.js
```diff
@@ -155,7 +157,7 @@ GSC.InPopupMainScript.setWindowHeightToFitContent = function() {
     chrome.app.window.current().innerBounds.height = wholeContentHeight;
   } else if (GSC.Packaging.MODE === GSC.Packaging.Mode.EXTENSION) {
     chrome.windows.update(
-        chrome.windows.WINDOW_ID_CURRENT, {'height': wholeContentHeight});
+        chrome.windows.WINDOW_ID_CURRENT, {'height': wholeContentHeight + titleBarHeight});
```
emaxx-google, line 160: nit: Please add a comment explaining why we need to do that - so the code reader can understand why we do it here and why we don't do it in the "if (GSC.Packaging.MODE === GSC.Packaging.Mode.APP)" case above.

common/js/src/popup-window/popup-opener.js
```diff
@@ -98,9 +100,7 @@ GSC.PopupOpener.createWindow = function(url, windowOptions, opt_data) {
             'frame': windowOptions['frame'],
             'hidden': windowOptions['hidden'],
             'id': windowOptions['id'],
-            'innerBounds': {
```
emaxx-google, line 101: Is this change intentional? I suppose the old code was actually correct, since `chrome.app.window.create()` takes the "innerBounds" parameter.

common/js/src/popup-window/popup-opener.js
```diff
@@ -62,6 +62,8 @@ const DEFAULT_DIALOG_CREATE_WINDOW_OPTIONS = {
   'visibleOnAllWorkspaces': true
 };
 
+const dialogWindowWidth = 500;
```
emaxx-google, line 65: I don't think this constant is needed - first, because the callers provide the "width" parameter, and, second, it's conceptually wrong to hardcode the width in this common file.

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompt-dialog.html
```diff
@@ -34,8 +34,6 @@
 <body>
   <h1 id="header"></h1>
 
-  <button id="close-window">×</button>
```
emaxx-google, line 37: The changes related to the "close" button should better be extracted into a separate PR, I think.

Also we still need this button in the app PACKAGING mode, so we shouldn't just delete it. (A probable solution could be to check the PACKAGING mode in the JS code and make the button invisible when the mode is !=app.)

emaxx-google (commented):
Thanks.

common/js/src/popup-window/popup-opener.js
ayaNader: This was a workaround because the default size is bigger than that in the APP mode.

emaxx-google: The root cause is here I think:
https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/blob/master/third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker.js#L55

(We need to pass `'width'` without nesting inside `'innerBounds'`)

common/js/src/popup-window/in-popup-main-script.js
```diff
@@ -154,8 +156,12 @@ GSC.InPopupMainScript.setWindowHeightToFitContent = function() {
   if (GSC.Packaging.MODE === GSC.Packaging.Mode.APP) {
     chrome.app.window.current().innerBounds.height = wholeContentHeight;
   } else if (GSC.Packaging.MODE === GSC.Packaging.Mode.EXTENSION) {
+    // Adding TITLE_BAR_HEIGHT as it's not included in the calculation of the
+    // offsetHeight..this is not needed in APP mode as it doesn't have a
```
emaxx-google, line 160: nit: Typos: s/offsetHeight..this/offsetHeight. This/

common/js/src/popup-window/popup-opener.js
emaxx-google: After fixing that place, the `dialogWindowWidth` constant could be removed I believe.

emaxx-google (changes requested):

emaxx-google (approved):
LGTM

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker.js
```diff
@@ -52,7 +52,7 @@ const USER_PROMPT_DIALOG_URL =
     'pcsc_lite_server_clients_management/user-prompt-dialog.html';
 
 const USER_PROMPT_DIALOG_WINDOW_OPTIONS_OVERRIDES = {
-  'innerBounds': {'width': 300}
+  'width': 500
```
emaxx-google, line 55: Why was the constant changed from 300 to 500? Was it intentional?

ayaNader: I remember from our offline discussion that 500 was the correct width..and this is actually the width I always got when testing both app and extension mode...but looking back through previous commits "300" seems to be the original width indeed so will create another PR for that

emaxx-google: Hmm, I don't remember us discussing the need to increase the width... If you can help me recall the reason for increasing the width let's discuss it. Otherwise, yup, let's revert this to use 300 pixels as it was before.

PR #501, opened 2021-12-07
[connector] Require Chrome >=96 in extension

emaxx-google:
Bump the Smart Card Connector's requirements be "Chrome >=96".

This is the first version of Chrome that allowlists the Smart Card
Connector to access smart card readers via WebUSB: crbug.com/1233881.
Hence explicitly declare this is a hard dependency, and also adjust the
flags passed to Closure Compiler accordingly, so that it can produce a
more modern JavaScript code without too many polyfills - which is
slightly smaller (-3%), easier to debug and probably slightly faster.

### DIFF
--- a/common/make/js_building_common.mk
+++ b/common/make/js_building_common.mk
@@ -92,8 +92,6 @@ endif
 # Add common Closure Compiler flags.
 #
 # Explanation:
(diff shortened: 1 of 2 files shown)


Fabian-Sommer (approved):
LGTM

PR #502, opened 2021-12-08
Remove close button in extension mode

ayaNader:
This PR removes the close button "x" in the extension mode since the popup already has a title bar with a close button.

### DIFF
--- a/third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompt-dialog-main.js
+++ b/third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompt-dialog-main.js
@@ -110,7 +110,7 @@ function closeWindowClickListener() {
 }
 
 function removeCloseButton() {
-  var closeButton = document.getElementById('close-window');
+  const closeButton = document.getElementById('close-window');
   closeButton.remove();
 }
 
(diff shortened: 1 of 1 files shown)


third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompt-dialog-main.js
```diff
@@ -109,6 +109,14 @@ function closeWindowClickListener() {
   window.close();
 }
 
+function removeCloseButton() {
+  var closeButton = document.getElementById('close-window');
+  closeButton.remove();
```
emaxx-google, line 114: If we really remove the element here, wouldn't the `goog.events.listen()` call in line 125 crash?
Either we need to call `removeCloseButton()` after adding all listeners, or we should do `.style.display = 'none'` instead of `remove()`.

emaxx-google (changes requested):
Thanks for working on this.

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompt-dialog-main.js
```diff
@@ -109,6 +109,11 @@ function closeWindowClickListener() {
   window.close();
 }
 
+function removeCloseButton() {
+  var closeButton = document.getElementById('close-window');
```
emaxx-google, line 113: nit: "var" is deprecated, "let"/"const" should be used instead. ("const" in this case)

emaxx-google (approved):
LGTM % nit

PR #503, opened 2021-12-10
[common] Fix error handling in MessageChannelPair

emaxx-google:
Fix a corner case in the MessageChannelPair which could lead to the
"cannot read property of undefined" exception in some error handling
scenarios.

The problem was that if one channel of the pair gets destroyed quickly
after the message is sent to it, by the time it's delivered (which
happens asynchronously) we were trying to find the message handler using
already cleared data members. An example when this can happen is when
the Emscripten module fails to load (repro example: set a breakpoint in
DevTools shortly after we start loading the module, wait a minute after the
breakpoint is hit and then continue), and the exception would be triggered
while trying to deliver a PC/SC command message to the module.

The fix has two parts: first, check the disposal state right before
delivering the message, and, second, propagate the disposal of any of
the pair's items to the other item too.

### DIFF
--- a/common/js/src/messaging/message-channel-pair-unittest.js
+++ b/common/js/src/messaging/message-channel-pair-unittest.js
@@ -17,52 +17,81 @@
 
 goog.require('GoogleSmartCard.MessageChannelPair');
 
+goog.require('goog.messaging.AbstractChannel');
+
 goog.setTestOnly();
 
 goog.scope(function() {
 
 const GSC = GoogleSmartCard;
 
+/**
+ * Wraps the given message channel into a promise, for simplifying writing unit
+ * tests.
+ * @param {!goog.messaging.AbstractChannel} messageChannel
+ * @return {!Promise<!Object|string>} serviceName
+ */
+function getPromiseForReceivedMessage(messageChannel, serviceName) {
+  return new Promise((resolve, reject) => {
+    messageChannel.registerService(serviceName, (receivedMessagePayload) => {
+      resolve(receivedMessagePayload);
+    });
+  });
+}
+
+/**
+ * Adds a listener that aborts the test when an unexpected message is received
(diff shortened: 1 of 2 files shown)


common/js/src/messaging/message-channel-pair-unittest.js
```diff
@@ -0,0 +1,113 @@
...
+      return new Promise((resolve, reject) => {
+        const messageChannelPair = new GSC.MessageChannelPair();
+
+        // Abort the test if the message got delivered despite the disposal.
+        messageChannelPair.getSecond().registerDefaultService((payload) => {
+          reject();
+        })
+
+        messageChannelPair.getFirst().send(
+            /*serviceName=*/ 'some-service', /*payload=*/ 'data');
```
Fabian-Sommer, line 103: Is there any way that this may complete before the message channel is disposed?

Fabian-Sommer (approved):
LGTM % question

common/js/src/messaging/message-channel-pair-unittest.js
emaxx-google: Good question - It can't happen in the current code, but I've added comments to explain that.

Please reopen if you think the comments are unclear or the behavior needs to be changed.

emaxx-google:
@ayaNader As discussed offline, I've changed the tests to use promises - that way it should easier to follow them.

ayaNader (approved):
LGTM % nits

ayaNader:
> @ayaNader As discussed offline, I've changed the tests to use promises - that way it should easier to follow them.

Yeah thats great, thanks!

emaxx-google:
> > @ayaNader As discussed offline, I've changed the tests to use promises - that way it should easier to follow them.
> 
> Yeah thats great, thanks!

Thank you for bringing up this topic.

PR #504, opened 2021-12-11
[wasm] Fix background page keepalive

emaxx-google:
Fix the logic that forcibly keeps the Smart Card Connector'
background page alive. It was enabled in #499, however it turns out it
didn't work correctly for two reasons: first, we didn't copy the
needed .js+.html files into the Connector, and, second, the Connector's
background page gets confused by a messaging port opened by the iframe.

This commit addresses both of the problems, by invoking the necessary
build command in the makefile and by fixing the background page to
ignore the iframe's port. Additionally the commit improves some comments
and downgrades the failure to stay alive from a fatal exception to a
warning (we shouldn't do harakiri just because of the suspend event).


Fabian-Sommer (approved):
LGTM

ayaNader (approved):
LGTM

PR #505, opened 2021-12-12
Undo unwanted change to permission popup width

ayaNader:
This PR undoes an unwanted width change to the permission prompt popup window.


emaxx-google:
Thanks. Yeah, let's continue using the old 300-pixel width - I don't remember we discussed any reason for increasing it.

emaxx-google (approved):

PR #506, opened 2021-12-13
[logging] Lower severity of loginState message

emaxx-google:
Downgrade the logging about the lack of the chrome.loginState API from
warning to info - it's a non-fatal error, and it's expected when run on
desktop OSes.

This warning has confused a number of people because chrome://extensions
surfaces all warnings as if they were errors.


ayaNader (approved):
LGTM

Fabian-Sommer (approved):
LGTM

PR #507, opened 2021-12-15
Re-apply "Patch the Closure Library to fix the test runner in Debug m…

emaxx-google:
…ode"

This commit was lost when upreving Closure Library last time. Original
description:

> This fixes the recent regression in the Closure Library, where the test
> runner started to break when running in Debug mode from "file:///" URLs.
>
> Some details:
>
> https://github.com/google/closure-library/issues/872


Fabian-Sommer (approved):
LGTM

PR #508, opened 2021-12-15
[usb] Unit test listDevices WebUSB impl

emaxx-google:
Add a unit test for the listDevices method in LibusbToWebusbAdaptor.


Fabian-Sommer (approved):
LGTM

PR #509, opened 2021-12-15
[usb] More listDevices tests for WebUSB

emaxx-google:
Add unit tests for dynamic scenarios, when a device is added or removed.
Also refactor tests a bit to avoid repetitions of fake objects.


Fabian-Sommer (approved):
LGTM

PR #511, opened 2021-12-15
[wasm] Abort if Emscripten is too old

emaxx-google:
Explicitly abort the compilation if the used version of Emscripten is
too old. Our env/initialize.sh already installs a reasonably good
version, so this commit adds protection for the case the developer ran
this script a long time ago or is trying to use their own old Emscripten
checkout.


Fabian-Sommer (approved):
LGTM

PR #512, opened 2021-12-16
[wasm] Fix crash on zero-size read/write

emaxx-google:
Fix the ipc_emulation library to allow reads/writes to in-memory files
when both the buffer is null and the size is zero. Before this commit,
the library was crashing on these.

The example case in which this could happen is when the Smart Card
Connector is receiving a command, like SCardControl, which has a buffer
parameter equal to an empty array. In that case our C++ handler creates
a zero-sized std::vector, calls data() on it and passes the result to
the C PC/SC-Lite code, which in turn writes this data to an in-memory
file. The C++ Standard allows data() to return nullptr when the size is
zero, hence we were hitting the crash in ipc_emulation.

The fix is to simply quickly exit when the read/write size is zero. We
don't even check the file correctness before doing this -
POSIX-specified contract on read()/write() implementations doesn't
require to do so.

### DIFF
--- a/common/cpp/src/google_smart_card_common/ipc_emulation.cc
+++ b/common/cpp/src/google_smart_card_common/ipc_emulation.cc
@@ -80,7 +80,9 @@ class IpcEmulation::InMemoryFile final {
     GOOGLE_SMART_CARD_CHECK(size >= 0);
     if (!size)
       return true;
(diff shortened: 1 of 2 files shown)


Fabian-Sommer (approved):
LGTM

PR #513, opened 2021-12-16
[js] Compile using use_types_for_optimization

emaxx-google:
Specify the use_types_for_optimization flag as true to Closure Compiler.

This allows to optimize the resulting file significantly better - at
least the size is decreased by 13%. For the background, originally in
this project we only used "safe" options of Closure Compiler and
disabled all optimizations, but the paradigm shifted and we started to
pay more attention to minimizing and speeding up the resulting program.


Fabian-Sommer (approved):
LGTM

PR #515, opened 2021-12-16
Unsubscribe reader when extension popup is closed

ayaNader:
This PR fixes readers not being unsubscribed when SCC window is closed in the extension mode.

This is to prevent onReadersChanged() to be called when the UI is closed, since it'll try to update a not-existing-anymore UI controls, causing an exception.


emaxx-google (approved):
LGTM, thanks! I also suspect the same problem exists with the list of devices (controlled by window-apps-displaying.js); but this can be tackled in a separate commit later.

PR #517, opened 2022-01-25
[js] Fix unit tests for optimized mode

emaxx-google:
Fix a few tests to make them work in the advanced optimizations mode of
Closure Compiler. This paves the way for enabling this mode in unit
tests in Release builds - currently tests are built without
optimizations, which puts us under risk of shipping broken code.

The challenge with unit tests is that many mocking tricks are broken
when the compiler renames fields and methods, inlines function calls
and substitutes constants. Hence in this commit we go back to classic
methods like "overrideXForTesting" and class inheritance in a few
places - these approaches are guaranteed to work with any kind of
optimizations.

Other issues fixed in this commit are mostly related to quoting object
properties where needed, in case the test injects or expects data in
particular format.


third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/user-prompting-checker-unittest.js
```diff
@@ -62,24 +63,26 @@ const MockedDialogBehavior = {
 };
 
 /**
- * Sets up a mock instance of KnownAppsRegistry.
+ * A fake implementtion of `TrustedClientsRegistry`.
```
Fabian-Sommer, line 66: nit: missing "a" in "implementation"

Fabian-Sommer (approved):
LGTM % nit

PR #519, opened 2022-01-25
[js] Fix Closure Library jsunit in optimized mode

emaxx-google:
Patch the Closure Library locally in order to fix an error in the unit
test framework when optimizations are on: the framework fails to call
the setUp() and tearDown() methods due to property renaming.

We'll try to upstream the fix separately from this local patch.

### DIFF
--- a/third_party/closure-library/src/closure/goog/testing/testcase.js
+++ b/third_party/closure-library/src/closure/goog/testing/testcase.js
@@ -2013,8 +2013,8 @@ goog.testing.TestCase.Test = function(name, ref, scope, objChain) {
 
   if (objChain) {
     for (var i = 0; i < objChain.length; i++) {
-      if (typeof objChain[i].setUp === 'function') {
-        this.setUps.push(goog.bind(objChain[i].setUp, objChain[i]));
+      if (typeof objChain[i]['setUp'] === 'function') {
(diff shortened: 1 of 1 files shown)


third_party/closure-library/src/closure/goog/testing/testcase.js
```diff
@@ -2013,8 +2013,8 @@ goog.testing.TestCase.Test = function(name, ref, scope, objChain) {
 
   if (objChain) {
     for (var i = 0; i < objChain.length; i++) {
-      if (typeof objChain[i].setUp === 'function') {
-        this.setUps.push(goog.bind(objChain[i].setUp, objChain[i]));
+      if (typeof objChain[i]['setUp'] === 'function') {
+        this.setUps.push(goog.bind(objChain[i]['setUp'], objChain[i]));
       }
       if (typeof objChain[i].tearDown === 'function') {
```
Fabian-Sommer, line 2019: tearDown is called properly?

Fabian-Sommer (approved):
LGTM % question

third_party/closure-library/src/closure/goog/testing/testcase.js
emaxx-google: Good catch!

PR #520, opened 2022-01-25
[js] Run release tests with optimizations

emaxx-google:
Enable Closure Compiler's aggressive optimizations for unit tests when
building in the Release mode.

This makes sure that the unit tests are running in the same
configuration as the code we ship, making it able for us to catch errors
caused by optimizations (which can happen as the compiler makes some
implicit assumptions about the code) in tests.


Fabian-Sommer (approved):
LGTM

PR #521, opened 2022-01-25
[uprev] Roll Closure Compiler to 20220104

emaxx-google:
Update the pinned third-party dependency on Closure Compiler to the
version 20220104. No local patches were necessary.


emaxx-google:
This is a mechanical change - only README.google files were edited, and the rest is just overwritten as-is.

Fabian-Sommer (approved):
LGTM

PR #522, opened 2022-01-25
[js] Use JavaScript 2020 in more configs

emaxx-google:
Stop specifying the "--language_out" flag to Closure Compiler, since
its recent versions complain about using it together with the
"--browser_featureset_year" flag. This commit leaves only
"--browser_featureset_year" used in our code.

Also start using the 2020 version of JavaScript (ECMAScript) in
more build configurations, and only stick to the old 2012 version in the
"App+PNaCl" build. Both of Emscripten and WebUSB require a fresh (at
least from 2020) version of Chrome.

### DIFF
--- a/common/make/js_building_common.mk
+++ b/common/make/js_building_common.mk
@@ -129,42 +129,33 @@ JS_BUILD_COMPILATION_FLAGS += \
 	--use_types_for_optimization=false \
 	--warning_level=VERBOSE \
 
-ifeq ($(PACKAGING),app)
+ifeq ($(PACKAGING) $(TOOLCHAIN),app pnacl)
 
 # Add Closure Compiler flags specific to App builds.
 #
 # Note that PACKAGING=app TOOLCHAIN=pnacl builds keep backwards compatibility
 # with all versions of Chrome since the start of this project, which is 2015.
-# For simplicity of the conditions here, we're only checking PACKAGING=app here,
-# since practically we don't want to deploy this packaging mode with other
-# toolchains.
 #
 # browser_featureset_year: Specify the year that's appropriate for the oldest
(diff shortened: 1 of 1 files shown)


common/make/js_building_common.mk
```diff
@@ -129,42 +129,33 @@ JS_BUILD_COMPILATION_FLAGS += \
...
+# versions had a few critical bugs, so for simplicity we assume >=96 as well.
 #
 # browser_featureset_year: Specify the year that's appropriate for the oldest
-#   Chrome release still supported by Emscripten builds, which is 2020.
-# language_out: Specify the JS language revision that corresponds to the oldest
-#   Chrome release still supported by Emscripten builds: ECMASCRIPT_2020.
+#   still supported Chrome release, which is 2020.
 JS_BUILD_COMPILATION_FLAGS += \
 	--browser_featureset_year=2020 \
-	--language_out=ECMASCRIPT_2020 \
```
Fabian-Sommer, line 166: I am a bit confused about the commit message vs. the actual change - The commit message says that ECMASCRIPT_2020 is used in more configs, but actually in both configurations the language_out flag is deleted. I guess that --browser_featureset_year=2020 implies --language_out=ECMASCRIPT_2020, and it is no longer possible to specify both flags because the closure compiler complains otherwise?

LGTM to the actual change, but I would restructure the commit message like

(1) delete language_out flag because of new closure compiler
(2) use 2020 version in more builds because ...

Fabian-Sommer (approved):
LGTM % comment

common/make/js_building_common.mk
emaxx-google: Thanks, done.

PR #524, opened 2022-01-26
[webusb] Unit-test config parsing for SCM SCR 3310

emaxx-google:
Add a unit test for the libusb-to-WebUSB bridge that exercises the
getConfigurations operation against the data captured from a real SCM
SCR 3310 device. Most importantly, this adds basic test coverage for the
code that parses extraData from the USB device descriptors.


third_party/libusb/webport/src/libusb-to-webusb-adaptor-unittest.js
```diff
@@ -315,5 +375,95 @@ goog.exportSymbol('testLibusbToWebusbAdaptor', {
...
+      'recipient': 'device',
+      'request': 0x06,
+      'value': 0x200,
+      'index': 0
+    };
+
+    // Arrange:
+    const device = deepClone(SCM_SCR_3310_DEVICE);
+    device['open'] = goog.testing.createFunctionMock();
+    device['open']().$do(async () => {});
```
Fabian-Sommer, line 392: Should this be does instead of do (also in line 395)? If not, should it be do instead of does in line 400/408? If both are accurate, could you explain the difference to me, please?

emaxx-google: Good catch, it should be "does" indeed. Apparently, the framework has both, and they have some subtle difference. I suspect the wrong one was working accidentally, because we pass an empty function.

Fabian-Sommer (approved):
LGTM

PR #526, opened 2022-02-18
[allowlist] Autoapprove permission for kiosk apps

emaxx-google:
Add a few kiosk apps that use smart card communication. Autoapprove
their permissions (without asking the user), in order to mitigate the
issue with invisible prompts in the kiosk mode when the Smart Card
Connector runs as a secondary app (normally the solution to this
problem is policy-for-extensions, however it doesn't exist for
secondary kiosk apps).

### DIFF
--- a/third_party/closure-library/src/closure/goog/testing/testcase.js
+++ b/third_party/closure-library/src/closure/goog/testing/testcase.js
@@ -2013,11 +2013,11 @@ goog.testing.TestCase.Test = function(name, ref, scope, objChain) {
 
   if (objChain) {
     for (var i = 0; i < objChain.length; i++) {
-      if (typeof objChain[i].setUp === 'function') {
-        this.setUps.push(goog.bind(objChain[i].setUp, objChain[i]));
+      if (typeof objChain[i]['setUp'] === 'function') {
+        this.setUps.push(goog.bind(objChain[i]['setUp'], objChain[i]));
       }
-      if (typeof objChain[i].tearDown === 'function') {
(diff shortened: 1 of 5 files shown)


third_party/closure-library/src/closure/goog/testing/testcase.js
```diff
@@ -2013,11 +2013,11 @@ goog.testing.TestCase.Test = function(name, ref, scope, objChain) {
...
     for (var i = 0; i < objChain.length; i++) {
-      if (typeof objChain[i]['setUp'] === 'function') {
-        this.setUps.push(goog.bind(objChain[i]['setUp'], objChain[i]));
+      if (typeof objChain[i].setUp === 'function') {
+        this.setUps.push(goog.bind(objChain[i].setUp, objChain[i]));
       }
-      if (typeof objChain[i]['tearDown'] === 'function') {
-        this.tearDowns.push(goog.bind(objChain[i]['tearDown'], objChain[i]));
+      if (typeof objChain[i].tearDown === 'function') {
+        this.tearDowns.push(goog.bind(objChain[i].tearDown, objChain[i]));
```
Fabian-Sommer: This reverts #519 - is this intentional?

Fabian-Sommer (approved):
LGTM % one question

third_party/closure-library/src/closure/goog/testing/testcase.js
Fabian-Sommer: I see you already fixed this - nvm then :)

PR #529, opened 2022-04-29
Automatically rebuild supported readers TXT file

emaxx-google:
Build the human_readable_supported_readers.txt file by default, instead
of requiring the developer to run "make package".

This makes it a bit more convenient for developers, and also it doesn't
slow anything down as the file is cheap to build.


Fabian-Sommer (approved):
LGTM

PR #531, opened 2022-04-30
Fix UI buttons not reacting

emaxx-google:
Fix the UI button listeners (like "About", "Help") to correctly work and
not throw exceptions.

The problem was that the "e.preventDefault()" calls were throwing
errors: apparently Closure Library passes not Event but
goog.events.BrowserEvent objects, and in the Release build the minifier
renames properties and methods of the goog.events.BrowserEvent class.
Hence the incorrect type signature we gave in the code confused the
compiler and didn't let it rename the "preventDefault()" calls into the
minified names.


Fabian-Sommer (approved):
LGTM

PR #532, opened 2022-05-03
Use Python 3 in build scripts

emaxx-google:
Switch most of the build scripts to using Python3 instead of simply
calling "python" (and assuming it's Python ±2.7).

What this commit doesn't address is the NaCl builds, since they use the
unmaintained third-party build scripts that still rely on Python 2.
Fixing those will be a separate effort.


Fabian-Sommer (approved):
LGTM

PR #534, opened 2022-05-03
Pin more Pip2 modules for NaCl webports

emaxx-google:
Workaround Python 2 errors occurring when initializing NaCl webports, by
requiring two modules ("logilab-common" and "lazy-object-proxy") not go
beyond their last Python2-supporting versions.

This commit is similar to #360. New errors will probably continue to
appear over time, as more and more pip modules drop Python 2
compatibility. The only solution is stopping the usage of NaCl, which
will happen in some mid-term future.


Fabian-Sommer (approved):
LGTM

PR #539, opened 2022-05-06
Install Python 2 locally under /env/

emaxx-google:
Stop relying on the system-wide Python 2 installation. Due to the sunset
of Python 2, it's becoming increasingly cumbersome to have it available
and not conflict with Python 3: both on developer machines and on
Continuous Integration bots.

Specific things done in this commit:
* Extend the env/initialize.sh script to bootstrap a completely
  standalone Python 2 installation (compiled from sources) under
  env/python2/.
* Create an associated virtual environment under
  env/python2_venv/, so that it's easy to temporarily switch into
  the "python is Python 2" mode for developer or in scripts.
* Install the "six" Pip 2 module into the venv, so that we don't
  need to ask the developer to install it manually.
* Patch the NaCl webports SDK to use that same virtualenv
  instead of creating its own, since nested virtualenvs don't
  work properly.
* Change the CI config to drop the system-wide Python 2
  installation.

A final note: all this will become unnecessary once we drop NaCl
support (in some mid-term future). However, we can't wait until this
happens and suffer from lacking development processes and CI.


emaxx-google:
Oops forgot to update make-all.sh et al.

emaxx-google:
> Oops forgot to update make-all.sh et al.

Done.

docs/building.md
```diff
@@ -113,6 +112,30 @@ You should only make sure, however, that the environment definitions are always
...
+project much faster, simply by running `make` in that directory.
+
+A sample command for an incremental build in the WebAssembly mode:
+
+```shell
+TOOLCHAIN=emscripten make -j30
+````
+
+A command for an incremental build in the NaCl mode includes activating a
+Python 2 virtual environment (which we don't enable by default since due to the
```
Fabian-Sommer, line 127: nit: cut "since"

Fabian-Sommer (approved):
LGTM

PR #541, opened 2022-05-06
Stop using Make >=4.3 functionality

emaxx-google:
Update our GoogleTest makefile to stop using the "&:"
syntax: it's only added in Make 4.3 which is apparently
too new and not available on many systems.

While the older Make can also parse and understand
the rule, it handled it incorrectly in the multi-threading
mode ("-j30" et al.), triggering multiple invocations of
the same rule, causing corruptions of build artifacts.

The older equivalent of these "&:" rules is weird (a pattern
rules), but should work equivalently good.


docs/building.md
```diff
@@ -11,8 +11,9 @@ Connector App and examples of how programs can communicate with the app.
   Building under a different \*nix system, Mac OS or Windows should be possible
   too, though most probably will require more efforts.
 
-* The following tools should be present in the system: **bash**, **make**,
-  **curl**, **sed**, **mktemp**, **realpath**, **xxd**.
+* The following tools should be present in the system: **bash**, **make**
+  (version 4.3+ is recommended), **curl**, **sed**, **mktemp**, **realpath**,
```
Fabian-Sommer, line 15: "recommended", or "required"?

Fabian-Sommer (approved):
LGTM

docs/building.md
emaxx-google: It's recommended, because the build will still work fine in a single-thread mode.

However, thinking more on this, perhaps we shouldn't depend on something that's so new that even the latest Ubuntu doesn't ship it by default... Also it might be even more problematic to set up for other developers (we know that some of them use Redhat).

So I'm thinking now if I should abandon this "compile stuff from sources" exercise and rather update our makefile to stop using this shiny new feature...

Fabian-Sommer (approved):
LGTM

PR #542, opened 2022-05-07
Use Python 3 venv

emaxx-google:
Create a Python 3 virtual environment under env/, and install the
Pip packages we need into it. This allows to avoid requiring the
developer of installing the packages themselves.

Note that we still rely on the developer to have a system-wide Python 3
installed. This is a valid expectation (unlike the sunset Python 2).


Fabian-Sommer (approved):
LGTM

emaxx-google:
(sorry, forgot to commit one file)

Fabian-Sommer (approved):
LGTM

PR #544, opened 2022-05-07
Run JS tests via Selenium

emaxx-google:
This commit is the final step in making the JavaScript tests run
via Selenium. In this change we integrate the previously prepared tools
and change the Makefile to run them: the helper Python script that runs
Selenium (that's installed as a Pip module into env/python3_venv),
executes Chromedriver (that's downloaded into env/).


Fabian-Sommer (approved):
LGTM

PR #545, opened 2022-05-07
Refactor SingleMessageBasedChannel JS tests

emaxx-google:
Refactor the unit tests of the SingleMessageBasedChannel JavaScript
class to be more readable and compact, by making a heavier use of
async/await and by extracting repeated blocks into SetUp/TearDown
methods.

This also fixes the flakiness that was caused by the tests not being
switched to the new override{Timeout|Interval}ForTesting() methods.


Fabian-Sommer (approved):
LGTM

PR #547, opened 2022-05-09
Bring back support of extension IDs in policies

### LINKED ISSUE #540: Policy handling is not backwards compatible
We used to have the following policy value for the smart card app:
```
{
   "force_allowed_client_app_ids":{
      "Value":[
         "haiffjcadagjlijoggckpgfnoeiflnem",
         "lbfgjakkeeccemhonnolnmglmfmccaag"
      ]
   }
}
```
i.e. a list of extension IDs.

With the latest dogfood release (version 1.3.7.1), these values no longer work. We are unexcpectedly seeing the "Don't grant access to unknown applications unless you know what you are doing!" dialog.

The logs indicate that it is looking for "chrome-extension://<<<EXTENSION_ID>>>" instead of "<<<EXTENSION_ID>>>" in the policy values.
Screenshot 2022-05-07 at 00 58 51

Updating the extension policy to the following:
```
{
   "force_allowed_client_app_ids":{
      "Value":[
         "chrome-extension://haiffjcadagjlijoggckpgfnoeiflnem",
         "chrome-extension://lbfgjakkeeccemhonnolnmglmfmccaag"
      ]
   }
}
```
solves the issue.

The old values should likely still work in order to be backwards-compatible.

emaxx-google:
Fix the regression introduced in the effort to switch to generic
messaging origins (#377). The regression is that we accidentally
switched the policy syntax from the list of extension IDs to the list of
origins, like "chrome-extension://something". This is a breaking
change for existing customer deployments who use the policy,
because the policy effectively becomes ignored.

This commit brings the backwards compatibility back, so that we still
recognize the policies that mention lists of extension IDs. Using the
new, origin-based, syntax in the policies is allowed too, although
not enforced.

This fixes #540.

### DIFF
--- a/third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/checker.js
+++ b/third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/checker.js
@@ -116,7 +116,7 @@ Checker.prototype.checkByManagedRegistry_ = function(
       'Checking permissions for the client ' + clientOrigin +
           ' through the managed registry...');
 
-  this.managedRegistry_.getById(clientOrigin)
+  this.managedRegistry_.getByOrigin(clientOrigin)
       .then(
           function() {
             goog.log.log(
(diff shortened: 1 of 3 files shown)


alex292 (approved):
LGTM, thanks for the quick fix!

Fabian-Sommer (approved):
LGTM

PR #550, opened 2022-05-10
Drop one more Make 4.3 usage in Emscripten builds

emaxx-google:
Stop using the Make >=4.3 only functionality in Emscripten build
scripts.

This commit is analogous to #541 - it was an oversight that there's one
more place left that's using the "&:" syntax.


Fabian-Sommer (approved):
LGTM

PR #551, opened 2022-05-10
Parallelize WebAssembly builds in CI

emaxx-google:
Pass "-j30" to Make in the Continuous Integration script when building
in the WebAssembly mode, so that the builds finish faster.

This is possible now that we fixed the issue with Make version
incompatibility (where parallel builds could corrupt files due to us
using the "&:" directive on Make <4.3).

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -31,4 +31,4 @@ jobs:
       run: >
         source env/activate &&
           source env/python2_venv/bin/activate &&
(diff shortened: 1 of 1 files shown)


Fabian-Sommer (approved):
LGTM

emaxx-google:
Going to land one half of this commit: looks like there's still some flakiness in the parallel NaCl builds, and I'm not sure it's worth it spending much time on that.

PR #555, opened 2022-05-11
Explicitly specify the "C" compilation language

emaxx-google:
Use the "-xc" compiler flag when compiling C (not C++) source files.

This is needed when compiling with recent compiler versions (e.g., Clang
13); previously the compiler was automatically selecting the language
based on the input file's extension.

### DIFF
--- a/third_party/ccid/webport/build/Makefile
+++ b/third_party/ccid/webport/build/Makefile
@@ -76,6 +76,7 @@ CCID_SOURCES := \
 #   under Linux (but are actually provided by the NaCl SDK too);
 # * The "macro-redefined" warning diagnostic is disabled because of some
 #   non-clean code;
+# * "xc": Use the "C" language.
 CCID_CPPFLAGS := \
 	$(COMMON_CPPFLAGS) \
 	-DBUNDLE='"ifd-ccid.bundle"' \
@@ -87,6 +88,7 @@ CCID_CPPFLAGS := \
 	-I$(ROOT_PATH)/common/cpp/src/google_smart_card_common/logging/syslog \
 	-I$(ROOT_PATH)/third_party/libusb/src/libusb \
 	-Wno-macro-redefined \
+	-xc \
 
 $(foreach src,$(CCID_SOURCES),$(eval $(call COMPILE_RULE,$(src),$(CCID_CPPFLAGS))))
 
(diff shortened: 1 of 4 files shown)


Fabian-Sommer (approved):
LGTM - But it is surprising to me that this is necessary, we don't need this flag for NaCl (perhaps because there, we specify something like -std=gnu99, which implies C?)

emaxx-google:
> LGTM - But it is surprising to me that this is necessary, we don't need this flag for NaCl (perhaps because there, we specify something like -std=gnu99, which implies C?)

Yeah I suspect it's something that LLVM changed recently, and both NaCl and Emscripten still use the older version that guesses the language itself.

Fabian-Sommer:
Actually, both NaCl and Emscripten are (also) using an even newer version of LLVM (near ToT) which (at least for NaCl) does not require this

emaxx-google:
> Actually, both NaCl and Emscripten are (also) using an even newer version of LLVM (near ToT) which (at least for NaCl) does not require this

Interesting! So if I revert this PR locally and use the stock LLVM/Clang (13.0.1-3+build2), I'm getting these warnings:

```
clang: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated [-Wdeprecated]
```

and then a bunch of failures because it tries to compile C code as C++.

Fabian-Sommer:
Hm, which binary is used to compile the C files? In the NaCl builds, we use `clang` for C and `clang++` for C++.

emaxx-google:
> Hm, which binary is used to compile the C files? In the NaCl builds, we use `clang` for C and `clang++` for C++.

We use "emcc" in Emscripten; somehow it's guessing the language automatically, without any warnings.
For NaCl, it turns out the makefiles provided by the NaCl SDK check the source file's extension and choose between "pnacl-clang++" and "nacl-clang".

Maybe I was wrong when testing the stock Clang because I used "clang++" for compiling everything. Probably I should've used "clang"? Let me try that out.

emaxx-google:
> Maybe I was wrong when testing the stock Clang because I used "clang++" for compiling everything. Probably I should've used "clang"? Let me try that out.

Yeah, looks like "clang" autodetects the input language correctly, without any warnings. I suppose I still have to use "clang++" for the linking stage, in order to link against the C++ standard library, but that's fine.

Alright, let me revert the current PR. Thanks for bringing this up!

PR #556, opened 2022-05-11
Run tests in CI

emaxx-google:
Append the execution of tests (both C++ and JavaScript) into the
Continuous Integration script.

This is possible now that the test execution is fully automated.

Note that we don't run tests in the NaCl mode yet: these still
require manual intervention at the moment, and automating these
will be done as a separate effort.


Fabian-Sommer (approved):

kaszarek (approved):
LGTM.

PR #557, opened 2022-05-11
Fix preprocessor issue with PC/SC and std headers

emaxx-google:
Fix weird compilation errors appearing when using a new Clang and
libstdc++. They are caused by some of PC/SC-Lite headers making
preprocessor definitions that collide with the Standard C++ Library's
symbols.


Fabian-Sommer (approved):
LGTM

PR #559, opened 2022-05-11
Add TOOLCHAIN=coverage

emaxx-google:
Add initial boilerplate for the "coverage" build mode.

Not all code actually builds successfully, and neither is actually any
instrumenting (like coverage collection) enabled. These will be tackled
in follow-up commits separately.


Fabian-Sommer (approved):
LGTM

PR #560, opened 2022-05-11
Support building Googletest for coverage

emaxx-google:
Update Googletest's makefile to support the TOOLCHAIN=coverage mode.

### DIFF
--- a/third_party/googletest/webport/build/Makefile
+++ b/third_party/googletest/webport/build/Makefile
@@ -23,10 +23,6 @@ TARGET := googletest
 include ../../../../common/make/common.mk
 include $(ROOT_PATH)/common/make/executable_building.mk
 
-ifneq ($(TOOLCHAIN),emscripten)
-$(error This Makefile is only supported in Emscripten builds.)
-endif
-
 # Library files that are created by this Makefile.
 #
 # Comments for the libraries:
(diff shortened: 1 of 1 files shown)


third_party/googletest/webport/build/Makefile
```diff
@@ -90,44 +86,66 @@ ARTIFACTS_LIBS_PATTERN := \
...
 $(ARTIFACTS_LIBS_PATTERN):
 	rm -rf out-artifacts
 	mkdir out-artifacts
-	emcmake \
-		cmake \
-		-E env \
-		CXXFLAGS="-pthread" \
-		LDFLAGS="-pthread" \
-		cmake \
+	$(CMAKE_ENV) $(CMAKE_TOOL) \
```
Fabian-Sommer, line 145: Is this the correct order?
Also, this now has one less `-E env` and one less `cmake` - but perhaps this is ok

Fabian-Sommer: I guess this works

Fabian-Sommer (approved):

third_party/googletest/webport/build/Makefile
emaxx-google: Yes, the order is first setting the environment vars and then running the tool.
As for `cmake -E env` - it was unnecessary indeed; turns out you can just set the environment variables (I'm not sure why I added this last time).

PR #563, opened 2022-05-12
Stub out main() in Coverage builds

emaxx-google:
Normal builds (WebAssembly/Emscripten) don't need a main() function,
because the framework takes care of the startup itself. In the coverage
builds, however, we're building using stock Clang, and it fails the
linking stage if there's no main() provided.

Note that the empty main() is just fine for coverage builds, since all
we care about is building an executable that the LLVM instrumentation
can see. All we want from it in these binaries is to account for all
functions that were present in corresponding object files (in order
to get a realistic coverage of the whole codebase, not only of the
files with unit tests).

### DIFF
--- a/common/cpp/src/empty_main.cc
+++ b/common/cpp/src/empty_main.cc
@@ -1,0 +1,17 @@
+// Copyright 2022 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


emaxx-google:
I hope I'm not missing something and that's how you do that: try to compile everything, with the intention to only run the unit test binaries.

Other binaries, with the empty main(), are only here for the purpose of making llvm-cov aware of all that code that is NOT exercised in unit tests, so that we get a realistic (low) coverage number in the end.

Fabian-Sommer (approved):

kaszarek (approved):
LGTM

PR #565, opened 2022-05-12
Store Debug/Release coverage profiles separately

emaxx-google:
Set up an environment variable to tell Clang instrumentation store the
coverage raw profiles in files called "Debug.profraw"/"Release.profraw"
instead of just "default.profraw".

In the follow-up change we'll implement merging all these raw profiles
into a single profile and building a summarized coverage report from it.


Fabian-Sommer (approved):

kaszarek (approved):
LGTM

PR #567, opened 2022-05-13
Script for creating a coverage report

emaxx-google:
Add a script "coverage-report.sh" that builds the code in the
TOOLCHAIN=coverage mode, merges the collected raw profiles (*.profraw),
and creates the resulting coverage report.

### DIFF
--- a/coverage-report.sh
+++ b/coverage-report.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+
+# Copyright 2022 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+CONFIGS="Release Debug"
+COVERAGE_PROFILE_DIRS="
+  common/cpp/build/tests
+  third_party/libusb/webport/build/tests"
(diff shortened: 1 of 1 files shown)


coverage-report.sh
```diff
@@ -0,0 +1,75 @@
...
+UNIT_TEST_EXECUTABLE_DIRS="
+  common/cpp/build/tests
+  third_party/libusb/webport/build/tests"
+IGNORE_FILENAME_REGEX="
+  third_party/googletest"
+
+log_message() {
+  local message=${1}
+
+  echo -e "\033[33;32m******* ${message} *******\033[0m"
```
Fabian-Sommer, line 34: What does this line do (apart from printing the message)? I would guess it is some coloring for the shell, but it is very cryptic.

Fabian-Sommer (approved):

coverage-report.sh
emaxx-google: Done - Added a comment. The syntax is cryptic indeed, but I don't know an easier way (and coloring helps read the logs significantly).

Fabian-Sommer (approved):

coverage-report.sh
```diff
@@ -0,0 +1,77 @@
...
+COVERAGE_PROFILE_DIRS="
+  common/cpp/build/tests
+  third_party/libusb/webport/build/tests"
+EXECUTABLE_MODULE_DIRS="
+  smart_card_connector_app/build/executable_module"
+UNIT_TEST_EXECUTABLE_DIRS="
+  common/cpp/build/tests
+  third_party/libusb/webport/build/tests"
+IGNORE_FILENAME_REGEX="
+  third_party/googletest"
```
emaxx-google, line 29: BTW, should we also exclude the unit test files? ("*test.cc")
Not sure what's the common practice for calculating the overall coverage percentage.

kaszarek: Unit tests provide coverage as well. 
IMHO we could start with having them and we will see if some lines are too exhaustively covered. Then, if needed, we will take it from there.

kaszarek (approved):
LGTM

PR #568, opened 2022-05-13
Install coverage deps in CI

emaxx-google:
Extend the Continuous Integration script to install the dependencies
needed for building in the TOOLCHAIN=coverage mode.

We also need to do "apt-get update" in the beginning, since by default
Github Actions VMs come with outdated Apt state (and installing some
packages will result in 404 errors).


Fabian-Sommer (approved):

kaszarek (approved):
LGTM

PR #575, opened 2022-05-16
Pass coverage flags to Clang linker too

emaxx-google:
The "fcoverage-mapping", "fprofile-instr-generate" flags should be
passed not only to the Clang compiler, but also to the linker too.
Otherwise, no coverage reports are collected, even though the
code is compiled with an instrumentation.

### DIFF
--- a/common/make/internal/executable_building_coverage.mk
+++ b/common/make/internal/executable_building_coverage.mk
@@ -59,20 +59,19 @@ endef
 
 # Flags passed to both compiler and linker.
 #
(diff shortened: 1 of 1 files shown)


emaxx-google:
PTAL - at some point I went too aggressive with cleaning up the code. To be honest, Clang's docs are also not very clear on that.

Fabian-Sommer (approved):

kaszarek (approved):

emaxx-google:
Thanks for the quick reviews folks!

PR #578, opened 2022-05-17
CI posts coverage percentage on PRs

emaxx-google:
Add a script that parses llvm-cov's report table and prints the
resulting line coverage percentage for C/C++ code. Run it in the
Continuous Integration script and post the result as an autogenerated
comment on the Pull Request.


emaxx-google:
(Still struggling to make the comment posting only work when the run is associated with a PR. If we don't skip it, I believe the whole action will fail when run on commits from the main branch.)

emaxx-google:
> (Still struggling to make the comment posting only work when the run is associated with a PR. If we don't skip it, I believe the whole action will fail when run on commits from the main branch.)

Done. Apparently samples in Github's own docs used a wrong syntax.

.github/workflows/makefile.yml
```diff
@@ -63,7 +63,22 @@ jobs:
...
     - name: Build coverage report
+      id: build-coverage
       env:
         CPLUS_INCLUDE_PATH: /usr/i686-linux-gnu/include/c++/${{ env.COVERAGE_LIBSTDCPP_VERSION }}/i686-linux-gnu/
-      run:
-        ./coverage-report.sh
+      run: |
+        ./coverage-report.sh | tee coverage-output.txt
+        echo "::set-output name=LINE_COVERAGE_PERCENT::$(
+          scripts/parse-coverage-output.py <coverage-output.txt
```
Fabian-Sommer, line 72: I like having a space after < and > (as we do for | )

coverage-report.sh
```diff
@@ -31,16 +31,18 @@ IGNORE_FILENAME_REGEX="
 log_message() {
   local message=${1}
 
-  # Create a colored log ("\033[33;32m" switches to the green color, and
+  # Create a colored stderr log ("\033[33;32m" switches to the green color, and
   # "\033[0m" switches back).
-  echo -e "\033[33;32m******* ${message} *******\033[0m"
+  echo -e "\033[33;32m******* ${message} *******\033[0m" >&2
```
Fabian-Sommer, line 36: ditto

coverage-report.sh
```diff
@@ -31,16 +31,18 @@ IGNORE_FILENAME_REGEX="
...
 }
 
+# Build all code and run tests. Make sure the output is redirected to stderr, so
+# that stdout stays non-littered.
 for config in ${CONFIGS}; do
   log_message "Building in mode \"${config}\"..."
-  TOOLCHAIN=coverage CONFIG=${config} make -j30
+  TOOLCHAIN=coverage CONFIG=${config} make -j30 >&2
```
Fabian-Sommer, line 43: ditto

coverage-report.sh
```diff
@@ -31,16 +31,18 @@ IGNORE_FILENAME_REGEX="
...
   log_message "Running tests in mode \"${config}\"..."
-  TOOLCHAIN=coverage CONFIG=${config} make -j30 test
+  TOOLCHAIN=coverage CONFIG=${config} make -j30 test >&2
```
Fabian-Sommer, line 45: ditto

emaxx-google:
> **Continuous Integration**: All tests passed. C/C++ test coverage: 5.35% lines.

That's a proud number!

scripts/parse-coverage-output.py
```diff
@@ -0,0 +1,67 @@
...
+  rows = []
+  for line in sys.stdin:
+    cells = split_columns(line)
+    if len(cells) <= 1:
+      # Empty line, or line with separators ("---..."), or line with a subtitle
+      # (e.g., "Files which contain no functions:").
+      continue
+    if len(cells) != len(column_names):
+      raise RuntimeError(f'Unexpected row line: ${line}')
+    rows.append(cells)
```
Fabian-Sommer, line 50: Looks good, but do we actually need to parse all of this if we only look at the last (non-separator etc.) line anyway?

Fabian-Sommer (approved):

scripts/parse-coverage-output.py
emaxx-google: I was planning to print the whole table in a follow-up commit.

The reason I parse it (and not just post the whole llvm-cov's output) is that this textual format is really hard to read because of long lines. Here's how it looks like: https://pastebin.com/yFajjvni
If I reformat the same data into a Markdown table it should be more readable (we can omit some columns if we want).

coverage-report.sh
emaxx-google: Unfortunately this doens't work:
```
syntax error near unexpected token `&'
```

P.S. I'm wondering whether I should've used Python for this script instead?..

Fabian-Sommer (approved):

.github/workflows/makefile.yml
```diff
@@ -63,7 +63,22 @@ jobs:
...
+
+    # Post an autogenerated comment on the Pull Request (in case the action is
+    # run because of one).
+    - name: Comment on pull request
+      if: github.event.pull_request
+      uses: peter-evans/create-or-update-comment@v2
+      with:
+        issue-number: ${{ github.event.pull_request.number }}
+        body: |
+          __Continuous Integration__: All tests passed.
```
kaszarek, line 83: Just curious - what will happen if coverage will fail or build fails? This message will not pop, right?

kaszarek (approved):
LGTM % nit.

.github/workflows/makefile.yml
emaxx-google: Yeah indeed, but when the action fails GitHub sends an email about the failed workflow.
When everything passes, there's no email, hence a separate comment seems useful; also we're posting the most important excerpt of the otherwise huge logs.

PR #581, opened 2022-05-17
Drop doc on manual generation of supported readers

emaxx-google:
Remove the mention of special build steps in order to generate
human_readable_supported_readers.txt from the docs.

Since #529, this file is always generated automatically.


Fabian-Sommer (approved):

emaxx-google:
Thanks!

PR #589, opened 2022-05-19
Rename global to pcsc_lite_server_web_port_service

emaxx-google:
This is a pure refactoring commit. The "global.{h,cc}" was a poor naming
scheme for the file and the corresponding class, so rename them to
"pcsc_lite_server_web_port_service.{h,cc}". Also rename their folder
from "google_smart_card_pcsc_lite_server" to just "public", which
reflects that these .h files can be included by code from other
subtrees.

Also while editing #include's pointing to it switch to using full
include paths like "third_party/pcsc-lite/etc.". This is a better and a
more maintainable naming scheme.

The new naming/inclusion scheme is inspired by Chromium's code
base.

### DIFF
--- a/third_party/pcsc-lite/naclport/server/build/Makefile
+++ b/third_party/pcsc-lite/naclport/server/build/Makefile
@@ -61,6 +61,7 @@ COMMON_CPPFLAGS := \
 	-I$(SOURCES_PATH) \
 	-I$(PCSC_LITE_ORIGINAL_HEADERS_DIR_PATH) \
 	-I$(PCSC_LITE_SOURCES_PATH) \
+	-I$(ROOT_PATH) \
 	-I$(ROOT_PATH)/common/cpp/src \
 
 # * DEBUG_HOTPLUG definition enables debug logging for the reader drivers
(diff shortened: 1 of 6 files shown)


third_party/pcsc-lite/naclport/server/src/public/pcsc_lite_server_web_port_service.h
```diff
@@ -42,14 +42,15 @@ namespace google_smart_card {
 //
 // Note: All methods except GetInstance are thread safe. Calls to GetInstance
 //       concurrent to class construction or destruction are not thread safe.
-class PcscLiteServerGlobal final {
+class PcscLiteServerWebPortService final {
```
emaxx-google, line 45: I'm only unsure about having both of words "server" and "service" within a single name.

third_party/pcsc-lite/naclport/server/build/Makefile
```diff
@@ -61,6 +61,7 @@ COMMON_CPPFLAGS := \
 	-I$(SOURCES_PATH) \
 	-I$(PCSC_LITE_ORIGINAL_HEADERS_DIR_PATH) \
 	-I$(PCSC_LITE_SOURCES_PATH) \
+	-I$(ROOT_PATH) \
 	-I$(ROOT_PATH)/common/cpp/src \
```
Fabian-Sommer, line 65: Are both of these lines required?

third_party/pcsc-lite/naclport/server/src/public/pcsc_lite_server_web_port_service.h
Fabian-Sommer: I think it is fine

Fabian-Sommer (approved):

third_party/pcsc-lite/naclport/server/build/Makefile
emaxx-google: For now, yes. Eventually my plan is to only leave `$(ROOT_PATH)` and get rid of most of other paths, but will tackle this step-by-step.

PR #592, opened 2022-05-21
Add Value::StrictlyEquals()

emaxx-google:
Implement comparison operation for the C++ Value class. To avoid any
confusion (e.g., "is floating-point 123.0 same as integer 123?") the
operation is defined as a regular method, rather than an overloaded
operator.

This new operation will be used by follow-up changes for implementing
new unit tests.

### DIFF
--- a/common/cpp/src/google_smart_card_common/value.cc
+++ b/common/cpp/src/google_smart_card_common/value.cc
@@ -108,6 +108,46 @@ Value::~Value() {
   Destroy();
 }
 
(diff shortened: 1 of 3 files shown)


Fabian-Sommer (commented):
I think all the test cases that are "Values of one type are different from values of another type" can be put into one test instead of getting spread out across multiple ones.
Also, we can probably test that `123`, `123.0` and `"123"` are actually different.

PR #594, opened 2022-05-21
Add TestingGlobalContext

emaxx-google:
Add a test implementation of the GlobalContext class that allows to
mock out the JavaScript counterpart in tests of C++ components. The
class simplifies setting up test expectations on the C++-to-JS messages
and simulating JS-to-C++ replies.

This class will be used by follow-up changes for implementing new unit
tests.


common/cpp/src/public/testing_global_context.cc
```diff
@@ -0,0 +1,165 @@
...
+  expectations_.push_back(std::move(expectation));
+}
+
+optional<TestingGlobalContext::Expectation>
+TestingGlobalContext::PopMatchingExpectation(const std::string& message_type,
+                                             const Value& request_payload) {
+  for (auto iter = expectations_.begin(); iter != expectations_.end(); ++iter) {
+    if (message_type == GetRequestMessageType(iter->requester_name) &&
+        request_payload.StrictlyEquals(iter->awaited_request_payload)) {
+      // A match found. The expectation is one-off, so remove it.
```
Fabian-Sommer, line 114: nit: `The expectation is a one of`

Fabian-Sommer (approved):

PR #598, opened 2022-05-22
Update PC/SC-Lite to 1.9.7

emaxx-google:
Uprev the PC/SC-Lite sources to their latest upstream release,
1.9.7.


emaxx-google:
PTAL; thanks! The review is only required for the second commit ("Fixes"), whereas the first commit contains the downloaded PC/SC-Lite 1.9.7 sources without any local modifications.

Fabian-Sommer (approved):

PR #603, opened 2022-05-23
Run CI jobs in parallel

emaxx-google:


kaszarek (approved):
LGTM and good thinking about reducing build time.

emaxx-google:
> Unfortunately, this causes some code duplication in the workflow .yml file; I looked at possible workarounds (composite actions or matrices), but due to subtle differences between commands we use for different modes neither will work smoothly. So I decided to keep the duplication for now.

Actually, after rethinking this, a composite action works fine for the steps related to build dependency installation. Updated the PR - we get rid of most of code duplication and still have the same speedup.

.github/workflows/makefile.yml
```diff
@@ -10,50 +10,44 @@ env:
...
-    - name: Apt update
-      run: sudo apt-get update
-
-    - name: Set up Python 3
-      uses: actions/setup-python@v3.1.2
+    - uses: ./.github/actions/build-dependencies-installation
       with:
-        python-version: 3.9
-        cache: pip
+        toolchain: pnacl
```
Fabian-Sommer, line 22: `pnacl` -> `emscripten`

Fabian-Sommer (approved):

.github/workflows/makefile.yml
emaxx-google: Good catch!

PR #611, opened 2022-05-24
Build in both Debug and Release modes in CI

emaxx-google:
Build and test code in Debug and Release modes in Continuous
Integration, as opposed to only using the Release mode (the
default).

That way, we make sure all unit tests pass with tighter
debug-only assertions, and also that no PR breaks the
Debug-only build mode.

Note that we don't cause a slowdown of CI, because
GitHub will spawn separate workers (VMs) for these
new jobs and run them concurrently.


kaszarek (approved):
LGTM.

Fabian-Sommer (approved):

PR #612, opened 2022-05-24
Run PC/SC-Lite's autoconf scripts in our builds

emaxx-google:
Change our makefiles to also run PC/SC-Lite's original "bootstrap" and
"configure" scripts (which are based on autoconf themselves).

This will be needed as a follow-up change is about to uprev PC/SC-Lite
to 1.9.7, and that version stopped distributing some autogenerated
headers (pcsclite.h and pcscd.h). We could try replicating the
autogeneration rules in our own makefiles, but a more future-proof
solution seems to just let the original PC/SC-Lite's build scripts
prepare them - which is what this commit implements.

Note that, as a side effect, we start requiring new dependencies that
a developer has to install on their machines: "autoconf",
"autoconf-archive" and "pkg-config".


third_party/pcsc-lite/naclport/build_configuration/Makefile
```diff
@@ -0,0 +1,56 @@
...
+# changed.
+#
+# Before running the build, we clean old build artifacts. On completion, the
+# target creates build.stamp that allows Make to track file timestamps and
+# detect changes.
+build.stamp: $(shell find ../../src)
+	@rm -f build.stamp
+	@git status --ignored ../../src | xargs rm -rf
+	cd ../../src && ./bootstrap && ./configure
+	@touch build.stamp
```
Fabian-Sommer, line 50: I assume this instruction is unaffected by the `cd` in the line above?

Fabian-Sommer (approved):

third_party/pcsc-lite/naclport/build_configuration/Makefile
emaxx-google: Correct, the current directory is reset to the original one in every line in the recipe (IIUC, because technically each line runs in a subshell).

PR #638, opened 2022-06-01
Tests  connector adding/removing reader

emaxx-google:
Add unit tests for the Connector app that verify scenarios
of plugging/unplugging a smart card reader.

### DIFF
--- a/smart_card_connector_app/src/application_unittest.cc
+++ b/smart_card_connector_app/src/application_unittest.cc
@@ -39,6 +39,7 @@
 #endif  // __native_client__
 
 using testing::ElementsAre;
+using testing::IsEmpty;
 
 namespace google_smart_card {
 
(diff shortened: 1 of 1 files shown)


smart_card_connector_app/src/application_unittest.cc
```diff
@@ -183,33 +223,114 @@ TEST_F(SmartCardConnectorApplicationTest, InternalApiSingleDeviceListing) {
...
+  // Arrange:
+
+  // Start with a single reader.
+  TestingSmartCardSimulation::Device device;
+  device.id = 123;
+  device.type = TestingSmartCardSimulation::DeviceType::kGemaltoPcTwinReader;
+  SetUsbDevices({device});
+
+  StartApplication();
+  EXPECT_EQ(reader_notification_observer().Pop(),
```
Fabian-Sommer, line 297: I guess we have some guarantee that WaitAndPop is not required here

Fabian-Sommer (approved):
LGTM (apart from missing CL description)

smart_card_connector_app/src/application_unittest.cc
emaxx-google: Added a clarification comment.

(BTW - between us - I myself don't fully understand why we need to wait in the scenarios added in this PR ::confused: . When I didn't wait, it failed on Github's VMs, but only in Release mode and not reproducible for me locally. But I suppose it might be legitimate, because PC/SC's threading model is not simple.)

PR #645, opened 2022-06-06
wip

emaxx-google:


smart_card_connector_app/src/application_unittest.cc
```diff
@@ -165,11 +180,117 @@ class SmartCardConnectorApplicationTest : public ::testing::Test {
...
+                       ArrayValueBuilder()
+                           .Add(scope)
+                           .Add(std::move(reserved1))
+                           .Add(std::move(reserved2))
+                           .Get());
+    waiter->Wait();
+    Value reply = std::move(*waiter).take_value();
+
+    GOOGLE_SMART_CARD_CHECK(reply.is_array());
+    GOOGLE_SMART_CARD_CHECK(!reply.GetArray().empty());
```
Fabian-Sommer, line 259: nit: Perhaps here we can put reply.GetArray() into some variable to save some GetArray() calls

Fabian-Sommer (approved):

PR #651, opened 2022-06-07
Unit-test device listing from JS

emaxx-google:
Add unit tests for SCardListReaders calls arriving from JavaScript.


Fabian-Sommer (approved):

emaxx-google:
Re-requesting the review - I've made the change you requested to use a helper to convert the call results, to save on code duplication.

smart_card_connector_app/src/application_unittest.cc
```diff
@@ -282,14 +290,24 @@ class SmartCardConnectorApplicationTest : public ::testing::Test {
...
+                                 std::vector<std::string>& out_readers) {
+    const RequestId request_id = ++request_id_counter_;
+    const std::string requester_name = GetJsClientRequesterName(handler_id);
+    auto waiter =
+        global_context_.CreateResponseWaiter(requester_name, request_id);
+    SimulateCallFromJs(
+        requester_name, request_id,
+        /*function_name=*/"SCardListReaders",
+        ArrayValueBuilder().Add(scard_context).Add(std::move(groups)).Get());
+    waiter->Wait();
```
Fabian-Sommer, line 308: Can this also be refactored into a method with inputs handler_id, function_name and an ArrayValue for the args?

emaxx-google: Good idea, done.

Fabian-Sommer (approved):
Thanks, this looks much better

PR #655, opened 2022-06-07
Unit-test hex dumping C++ helpers

emaxx-google:
Add unit tests for untested helpers in hex_dumping.h in //common/.

### DIFF
--- a/common/cpp/src/google_smart_card_common/logging/hex_dumping_unittest.cc
+++ b/common/cpp/src/google_smart_card_common/logging/hex_dumping_unittest.cc
@@ -36,9 +36,23 @@ TEST(LoggingHexDumpingTest, HexDumpByte) {
   EXPECT_EQ("0xFF", HexDumpByte(static_cast<uint8_t>(255)));
 }
 
+TEST(LoggingHexDumpingTest, HexDumpDoublet) {
+  EXPECT_EQ("0x0000", HexDumpDoublet(static_cast<int16_t>(0)));
+  EXPECT_EQ("0x0000", HexDumpDoublet(static_cast<uint16_t>(0)));
+
+  EXPECT_EQ("0x0001", HexDumpDoublet(static_cast<int16_t>(1)));
+  EXPECT_EQ("0x0001", HexDumpDoublet(static_cast<uint16_t>(1)));
+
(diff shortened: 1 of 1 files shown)


common/cpp/src/google_smart_card_common/logging/hex_dumping_unittest.cc
```diff
@@ -96,4 +119,16 @@ TEST(LoggingHexDumpingTest, HexDumpUnknownSizeInteger) {
...
+TEST(LoggingHexDumpingTest, HexDumpBytes) {
+  EXPECT_EQ(HexDumpBytes(nullptr, 0), "");
+
+  const uint8_t kArray[] = {1, 2, 123};
+  EXPECT_EQ(HexDumpBytes(kArray, 3), "0x01 0x02 0x7B");
+}
+
+TEST(LoggingHexDumpingTest, HexDumpBytesVector) {
+  EXPECT_EQ(HexDumpBytes({}), "");
+  EXPECT_EQ(HexDumpBytes({1, 2, 123}), "0x01 0x02 0x7B");
```
Fabian-Sommer, line 131: Parameters to EXPECT_EQ seem to be in the wrong order here

Fabian-Sommer (approved):

common/cpp/src/google_smart_card_common/logging/hex_dumping_unittest.cc
emaxx-google: Do you think the inverse order would be more helpful?
There was a discussion about the inverse ("Yoda-style") order in one of groups, and seems that people aren't supportive of it: https://groups.google.com/a/google.com/g/chromium-dev-internal/c/W7nPEr5Y0R8/m/i8qsXPg8DAAJ

Fabian-Sommer: Correction: This seems right, but the rest of the file has it reversed :)

emaxx-google: Agreed, let me fix it in the same CL.

emaxx-google: Done.

PR #666, opened 2022-06-10
Simulate Cosmo ID 7.0 card presence in unit tests

emaxx-google:
Add the most basic simulation of an arbitrary smart card (Cosmo ID 7.0
currently) into the Smart Card Connector unit tests. All we simulate at
the moment is the presence of the card and its ATR (answer-to-reset).

### DIFF
--- a/smart_card_connector_app/src/application_unittest.cc
+++ b/smart_card_connector_app/src/application_unittest.cc
@@ -836,6 +836,54 @@ TEST_F(SmartCardConnectorApplicationSingleClientTest,
             SCARD_STATE_CHANGED | SCARD_STATE_UNAVAILABLE));
 }
 
+// Test `SCardGetStatusChange()` call from JS returns the reader and card
+// information.
+TEST_F(SmartCardConnectorApplicationSingleClientTest,
+       SCardGetStatusChangeWithCard) {
+  // Arrange:
+  TestingSmartCardSimulation::Device device;
+  device.id = 123;
+  device.type = TestingSmartCardSimulation::DeviceType::kGemaltoPcTwinReader;
+  device.card_type = TestingSmartCardSimulation::CardType::kCosmoId70;
+  SetUsbDevices({device});
+  StartApplication();
+  SetUpJsClient();
+  SetUpSCardContext();
+
+  // Act:
+  std::vector<Value> reader_states;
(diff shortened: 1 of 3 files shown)


smart_card_connector_app/src/application_unittest.cc
```diff
@@ -840,6 +840,54 @@ TEST_F(SmartCardConnectorApplicationSingleClientTest,
...
+  // Act:
+  std::vector<Value> reader_states;
+  // This call is expected to return immediately, since we pass
+  // `SCARD_STATE_UNKNOWN`.
+  EXPECT_EQ(SimulateGetStatusChangeCallFromJsClient(
+                kFakeHandlerId, scard_context(),
+                /*timeout=*/INFINITE,
+                ArrayValueBuilder()
+                    .Add(DictValueBuilder()
+                             .Add("reader_name", "Gemalto PC Twin Reader 00 00")
```
Fabian-Sommer, line 866: nit: We are using this Gemalto string a lot, we can probably use a constant for it

smart_card_connector_app/src/application_unittest.cc
```diff
@@ -840,6 +840,54 @@ TEST_F(SmartCardConnectorApplicationSingleClientTest,
...
+              DictContains("current_state", SCARD_STATE_UNKNOWN));
+  EXPECT_THAT(
+      reader_states[0],
+      DictContains("event_state", SCARD_STATE_CHANGED | SCARD_STATE_PRESENT));
+  EXPECT_THAT(
+      reader_states[0],
+      DictContains("atr", std::vector<uint8_t>{
+                              0x3B, 0xDB, 0x96, 0x00, 0x80, 0xB1, 0xFE, 0x45,
+                              0x1F, 0x83, 0x00, 0x31, 0xC0, 0x64, 0xC7, 0xFC,
+                              0x10, 0x00, 0x01, 0x90, 0x00, 0x74}));
```
Fabian-Sommer, line 888: Where does this come from?

Fabian-Sommer (approved):

smart_card_connector_app/src/application_unittest.cc
emaxx-google: Replaced with a method call.

PR #668, opened 2022-06-10
Unit-test GetStatusChange with card insertion

emaxx-google:
Add unit tests for the SCardGetStatusChange calls arriving from
JavaScript in case a card gets dynamically inserted into a reader.

### DIFF
--- a/smart_card_connector_app/src/testing_smart_card_simulation.cc
+++ b/smart_card_connector_app/src/testing_smart_card_simulation.cc
@@ -312,6 +312,10 @@ std::vector<uint8_t> MakeParametersTransferReply(
 // Builds a RDR_to_PC_NotifySlotChange message.
 std::vector<uint8_t> MakeNotifySlotChangeTransferReply(CcidIccStatus icc_status,
                                                        bool slot0_changed) {
+  // The message format is per CCID specs. The status byte contains two bits per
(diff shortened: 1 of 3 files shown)


smart_card_connector_app/src/testing_smart_card_simulation.cc
```diff
@@ -309,11 +309,47 @@ std::vector<uint8_t> MakeParametersTransferReply(
   return transfer_reply;
 }
 
+// Builds a RDR_to_PC_NotifySlotChange message.
+std::vector<uint8_t> MakeNotifySlotChangeTransferReply(CcidIccStatus icc_status,
+                                                       bool slot0_changed) {
+  const uint8_t slot0_current_bit = icc_status != CcidIccStatus::kNotPresent;
```
Fabian-Sommer, line 315: For my own understanding: This is 0x00 if `icc_status` is `CcidIccStatus::kNotPresent`, and 0x01 if it is anything else, right?

Fabian-Sommer (approved):

smart_card_connector_app/src/testing_smart_card_simulation.cc
emaxx-google: Correct. Let me add a comment to further clarify that.

PR #671, opened 2022-06-13
Update PC/SC-Lite to 1.9.8

emaxx-google:
Uprev the PC/SC-Lite sources to their latest upstream release,
1.9.8. We also drop all our patches that we've submitted upstream.


emaxx-google:
Please review all commits except for the first one ("upstream 1.9.8" contains only verbatim upstream release).

Fabian-Sommer (approved):

PR #672, opened 2022-06-13
Update CCID to 1.5.0

emaxx-google:
Uprev the CCID sources to their latest upstream release, 1.5.0.
We also drop the local patch that we've submitted upstream.


emaxx-google:
Please review all patches except for the first one ("upstream 1.5.0" is a verbatim copy of the upstream release).

PR #675, opened 2022-06-13
Unit-test Connector unknown PC/SC function call

emaxx-google:
Add unit test for the Smart Card Connector for the scenario in which it
receives a PC/SC request with an unknown function name.

This change includes the refactoring that was needed in order to
properly parse failure response messages (previously the tests were
parsing both requests and response messages in the same way, assuming
that the "payload" field is always present).

### DIFF
--- a/smart_card_connector_app/src/application_unittest.cc
+++ b/smart_card_connector_app/src/application_unittest.cc
@@ -109,6 +109,7 @@ class ReaderNotificationObserver final {
   void OnMessageToJs(const std::string& event_name,
                      optional<Value> message_data,
                      optional<RequestId> /*request_id*/) {
+    ASSERT_TRUE(message_data);
     std::string notification =
         event_name + ":" +
         message_data->GetDictionaryItem("readerName")->GetString();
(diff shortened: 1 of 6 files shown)


Fabian-Sommer:
This seems to have a merge conflict

emaxx-google:
> This seems to have a merge conflict

Rebased.

smart_card_connector_app/src/application_unittest.cc
```diff
@@ -106,11 +107,11 @@ class ReaderNotificationObserver final {
 
  private:
   void OnMessageToJs(const std::string& event_name,
-                     Value message_data,
+                     optional<Value> message_data,
                      optional<RequestId> /*request_id*/) {
```
Fabian-Sommer, line 111: should we `GOOGLE_SMART_CARD_CHECK(message_data);` here?

Fabian-Sommer (approved):

smart_card_connector_app/src/application_unittest.cc
emaxx-google: Good idea - added an ASSERT here and the CHECK in a couple of other places.

PR #683, opened 2022-06-19
Reapply CCID patch for ifdLogLevel

emaxx-google:
Apply the 0001-Patch-ccid-tweak-ifdDriverOptions-option.patch patch to
our CCID fork. It was forgotten during the last CCID uprev.

### DIFF
--- a/third_party/ccid/src/src/Info.plist.src
+++ b/third_party/ccid/src/src/Info.plist.src
@@ -51,7 +51,7 @@
 	-->
 
 	<key>ifdDriverOptions</key>
-	<string>0x0000</string>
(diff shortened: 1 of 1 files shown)


third_party/ccid/src/src/Info.plist.src
```diff
@@ -51,7 +51,7 @@
 	-->
 
 	<key>ifdDriverOptions</key>
-	<string>0x0000</string>
```
emaxx-google, line 54: BTW it's not the first time I forget to reapply it after the uprev.
I'm starting to believe I should just write a script to do this change automatically during the build...

PR #684, opened 2022-06-20
Unit-test LibusbTracingWrapper

emaxx-google:
Extend the libusb_js_proxy_unittest.cc file to additionally test
LibusbTracingWrapper.

We don't test the latter in isolation in its own test file, since we'd
essentially want to repeat all test cases we wrote for LibusbJsProxy.
Moreover, testing it in isolation would be weird as the test mocks would
need repeating pieces of LibusbJsProxy (e.g., "automatically delete the
transfer if LIBUSB_TRANSFER_FREE_TRANSFER is specified").


emaxx-google:
Note: This PR will fail on bots, because apparently there are a few issues in `LibusbTracingWrapper` that we need to fix. But I'll upload the fixes separately - the current PR still holds.

Fabian-Sommer (approved):

Issue #685, opened 2022-06-20
Fix LibusbTracingWrapper bugs


emaxx-google:
This class (which is used as a thin wrapper around the real `LibusbJsProxy` implementation) apparently has a few issues:

- It effectively discards the `LIBUSB_TRANSFER_FREE_TRANSFER` flag, leading to memory leaks when this flag is passed by callsites.
- It hits a debug CHECK assertion when a callsite attempts to cancel a not-submitted-yet transfer.
- It triggers use-after-free when a callsite attempts to cancel an already-completed transfer.
- It triggers reads past end-of-array when a control transfer is passed.

PR #687, opened 2022-06-20
Fix USB tracing wrapper on unsent transfer cancel

emaxx-google:
Fix the LibusbTracingWrapper class to not crash on a debug assertion
when a cancellation of a not-yet-submitted transfer is requested.

The callsites shouldn't normally trigger such calls, but apparently the
original Libusb interface does support them, and we also support that in
the real LibusbJsProxy implementation.

This bugfix contributes to the bugfixes tracked by #685. The
regression unit test is part of #684.

### DIFF
--- a/third_party/libusb/webport/src/libusb_tracing_wrapper.cc
+++ b/third_party/libusb/webport/src/libusb_tracing_wrapper.cc
@@ -850,7 +850,7 @@ int LibusbTracingWrapper::LibusbCancelTransfer(libusb_transfer* transfer) {
   // here the actual cancellation should be called with the wrapper transfer.
   libusb_transfer* const wrapped_transfer = GetWrappedTransfer(transfer);
   if (!wrapped_transfer) {
-    // The original transfer not submitted yet or already completed.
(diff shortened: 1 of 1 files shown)


third_party/libusb/webport/src/libusb_tracing_wrapper.cc
```diff
@@ -849,7 +849,10 @@ int LibusbTracingWrapper::LibusbCancelTransfer(libusb_transfer* transfer) {
   // wrapped transfer (see the LibusbSubmitTransfer method implementation). So
   // here the actual cancellation should be called with the wrapper transfer.
   libusb_transfer* const wrapped_transfer = GetWrappedTransfer(transfer);
-  GOOGLE_SMART_CARD_CHECK(wrapped_transfer);
+  if (!wrapped_transfer) {
+    // The original transfer not submitted yet or already completed.
```
Fabian-Sommer, line 853: nit: "The original transfer is either not yet submitted or already completed."

Fabian-Sommer (approved):

Issue #681, opened 2022-06-18
Fix handle leaks on reader disconnection


emaxx-google:
Apparently we have a problem that after disconnecting a smart card reader various in-memory structures might get out of sync. As a result, we can have these issues:

- `SCardReleaseContext()` might return an error for a correct context (it can happen if there was an orphan `SCARDHANLE` pointing to an already-disappeared reader: ReleaseContext attempts to disconnect all active handles, which ends with an error for orphan handles);
- `PcscLiteClientRequestProcessor::s_card_handles_registry_` might contain contexts&handles that don't exist anymore (because this C++ layer doesn't know that the underlying PC/SC daemon decided to close some handles internally).

Note also that `SCardDisconnect()` starts returning an error for a correct handle that points to an already-disappeared reader. This might be intended from PC/SC-Lite's implementation perspective, but is not obvious from the API documentation.

emaxx-google:
It seems that a full solution would require significant rearchitecture of our webport or significant changes in the upstream PC/SC-Lite, so I'm looking at more narrow-scoped mitigations: fixing unexpected "invalid context" errors in PC/SC-Lite and cleaning up known-to-be-invalid handles in `PcscLiteClientRequestProcessor`.

PR #695, opened 2022-06-20
Unit-test Disconnect&ReleaseContext after unplug

emaxx-google:
Add a regression unit test for the PC/SC-Lite bugs when it failed to
close the handle and release the context after the reader got unplugged
(see #681).

### DIFF
--- a/smart_card_connector_app/src/application_unittest.cc
+++ b/smart_card_connector_app/src/application_unittest.cc
@@ -1132,7 +1132,7 @@ TEST_F(SmartCardConnectorApplicationSingleClientTest, NonExistingFunctionCall) {
   EXPECT_THAT(response, IsNullOptional());
 }
 
-// `SCardDisconnect()` and `SCardReleaseContext` calls from JS should succeed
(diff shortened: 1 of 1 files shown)


smart_card_connector_app/src/application_unittest.cc
```diff
@@ -1131,4 +1132,45 @@ TEST_F(SmartCardConnectorApplicationSingleClientTest, NonExistingFunctionCall) {
   EXPECT_THAT(response, IsNullOptional());
 }
 
+// `SCardDisconnect()` and `SCardReleaseContext` calls from JS should succeed
```
Fabian-Sommer, line 1135: nit: Does this actually test SCardReleaseContext?

Fabian-Sommer (approved):

smart_card_connector_app/src/application_unittest.cc
emaxx-google: Yes. Clarified that in a comment.

PR #702, opened 2022-06-21
Fix goog exports from standalone JS library

emaxx-google:
Make a few symbols (mostly related to logging) from Closure Library be
exposed as global unminified names from the standalone JavaScript smart
card client library.

We don't know if anyone uses these, but these might be useful if the
application wants to do something with the logs produced by our library.
The exposure of these symbols worked in the past, but probably broke
when we updated to a newer Closure Compiler that became more aggressive
in skipping input files; also our exports became outdated as Closure
Library reworked the logging library significantly.

This change fixes the exports to reflect the up-to-date Closure Library,
and also to make them applied indeed.


Fabian-Sommer (approved):

frogmam:
can someone tell me the javascript for background color change

PR #704, opened 2022-06-24
Add policy to allow SCardDisconnect workaround

annklr:
Add a new policy that allows configuring client app IDs using a workaround if a protocol mismatch error occurs during SCardConnect.

- `admin-policy-service.js` reads the policy from the managed storage and sends policy update messages to the server
- `admin_policy_getter.cc` stores the current policy
- `client_request_processor.cc` reads the policy and applies the card reset, if required


common/cpp/src/google_smart_card_common/admin_policy_getter.h
```diff
@@ -0,0 +1,53 @@
...
+
+#ifndef GOOGLE_SMART_CARD_COMMON_ADMIN_POLICY_GETTER_H_
+#define GOOGLE_SMART_CARD_COMMON_ADMIN_POLICY_GETTER_H_
+
+#include <string>
+#include <vector>
+
+namespace google_smart_card {
+
+struct AdminPolicy {
```
emaxx-google, line 23: There's some code layering issue - the problem is that //common/ is reused between Smart Card Connector and third-party client applications, and we don't want to put Connector-specific details into common classes.

Just thinking out loud:
1. We have generic piece of code that sends the policy blob from the JS side to the C++ side - it's definitely abstract enough to be fine in //common/.
2. We also have the knowledge about which policies exist in Smart Card Connector - this should ideally NOT sit in //common/.
3. And we have a callsite in //third_party/pcsc-lite/naclport/server_clients_management/ that wants to check the policy.

I think the fastest option would be to move the whole admin_policy_getter.h to //third_party/pcsc-lite/naclport/server_clients_management/. That directory is Connector-specific, so it's fine to put all that code there.

Or we can think about some more future-proof solution, that'll work for policy consumers outside //third_party/pcsc-lite/naclport/server_clients_management/. One immediate option would be to change AdminPolicyGetter to only expose the policy blob as a `Value` object; but that has the downside that each callsite needs to parse the policy themselves.

third_party/pcsc-lite/naclport/server_clients_management/src/clients_manager.cc
```diff
@@ -146,15 +169,35 @@ bool PcscLiteServerClientsManager::DeleteHandlerMessageListener ::
...
+
+std::string PcscLiteServerClientsManager::UpdateAdminPolicyMessageListener::
+    GetListenedMessageType() const {
+  return kUpdateAdminPolicyMessageType;
+}
+
+bool PcscLiteServerClientsManager::UpdateAdminPolicyMessageListener::
+    OnTypedMessageReceived(Value data) {
+  const UpdateAdminPolicyMessageData message_data =
+      ConvertFromValueOrDie<UpdateAdminPolicyMessageData>(std::move(data));
```
emaxx-google, line 184: Here we shouldn't use the "OrDie" variant because it'll crash the program if someone sets a misconfigured policy. Let's use `ConvertFromValue()` and print a warning log when there's a conversion issue.

third_party/pcsc-lite/naclport/server_clients_management/src/clients_manager.h
```diff
@@ -95,6 +97,21 @@ class PcscLiteServerClientsManager final {
   void ShutDown();
 
  private:
+  class UpdateAdminPolicyMessageListener final : public TypedMessageListener {
```
emaxx-google, line 100: Is this separate class needed? Can we use `AdminPolicyGetter` directly?

emaxx-google (changes requested):

common/cpp/src/google_smart_card_common/admin_policy_getter.cc
```diff
@@ -0,0 +1,69 @@
...
+  return "{[" + force_allowed_client_app_ids + "], [" +
+         scard_disconnect_fallback_client_app_ids + "]}";
+}
+
+AdminPolicyGetter::AdminPolicyGetter() = default;
+
+AdminPolicyGetter::~AdminPolicyGetter() = default;
+
+AdminPolicy AdminPolicyGetter::Get() {
+  return admin_policy_;
```
emaxx-google, line 58: Let's also make it a blocking call until the policies are received for the first time. You'd need an `std::condition_variable` for that (you'd call `wait()` here and `notify_all()` in `UpdateAdminPolicy()`).

common/cpp/src/google_smart_card_common/admin_policy_getter.cc
```diff
@@ -0,0 +1,69 @@
...
+  // Note: Strings passed to WithField() below must match the keys in
+  // //third_party/pcsc-lite/naclport/server_clients_management/src/client-handler.js.
+  return Describe("AdminPolicy")
+      .WithField(&AdminPolicy::force_allowed_client_app_ids,
+                 "force_allowed_client_app_ids")
+      .WithField(&AdminPolicy::scard_disconnect_fallback_client_app_ids,
+                 "scard_disconnect_fallback_client_app_ids");
+}
+
+std::string DebugDumpAdminPolicyString(const AdminPolicy admin_policy) {
```
emaxx-google, line 35: This function can probably be replaced by dumping the received `Value` object (see value_debug_dumping.h).

third_party/pcsc-lite/naclport/server_clients_management/src/clients_manager.cc
emaxx-google, line 184: Another interesting aspect here is that when one adds a new policy in the future, they might forget about registering it in `StructValueDescriptor<AdminPolicy>::GetDescription()`, which will lead to breaking old policies too.

I'll say it's not a blocker for your PR, just something that came to my mind and that we might want to fix eventually. (I suppose it can be something like adding something like `StructValueDescriptor::PermitUnknownKeys()`.)

third_party/pcsc-lite/naclport/server_clients_management/src/admin-policy-service.js
```diff
@@ -0,0 +1,217 @@
...
+  this.logger = GSC.Logging.getScopedLogger(
+      'PcscLiteServerClientsManagement.AdminPolicyService');
+
+  /** @private */
+  this.serverMessageChannel_ = serverMessageChannel;
+
+  /** @private */
+  this.serverReadinessTracker_ = serverReadinessTracker;
+
+  // The policy update messages are deferred until the server readiness is
```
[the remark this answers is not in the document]
annklr: Unfortunately it fails without it in PcscLiteServerSocketsManager::GetInstance() (https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/blob/135d5564ab6cc2b4f6486d2446aa0bdc58ed59eb/third_party/pcsc-lite/naclport/server/src/server_sockets_manager.cc#L54)

common/cpp/src/google_smart_card_common/admin_policy_getter.h
annklr: Thanks, I've moved the file. But please let me know if you would prefer the other option.

third_party/pcsc-lite/naclport/server_clients_management/src/admin-policy-service.js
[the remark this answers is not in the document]
emaxx-google: Hmm but why sending policies touches `g_pcsc_lite_server_sockets_manager`? In my mental model sending policies from JS to C++ should just store them in some in-memory variable (inside the `AdminPolicy` class) but not touch anything else.

common/cpp/src/google_smart_card_common/value_conversion.h
```diff
@@ -381,6 +384,12 @@ class StructValueDescriptor {
       return std::move(*this);
     }
 
+    Description&& PermitUnknownFields() && {
```
emaxx-google, line 387: nit: Let's add some short comment:

```
  // If called, the value-to-struct conversion will ignore unexpected dictionary keys.
```

P.S. Thanks for adding this helper!

smart_card_connector_app/src/application_unittest.cc
```diff
@@ -1066,6 +1066,56 @@ TEST_F(SmartCardConnectorApplicationSingleClientTest,
...
+  EXPECT_EQ(SimulateDisconnectCallFromJsClient(kFakeHandlerId, scard_handle,
+                                               SCARD_LEAVE_CARD),
+            SCARD_S_SUCCESS);
+  // Attempt connecting via a different protocol ("ANY" denotes "either T0 or
+  // T1").
+  LONG return_code = SimulateConnectCallFromJsClient(
+      kFakeHandlerId, scard_context(), kGemaltoPcTwinReaderPcscName0,
+      SCARD_SHARE_SHARED, SCARD_PROTOCOL_ANY, scard_handle, active_protocol);
+
+  // Assert:
```
emaxx-google, line 1115: Let's also check the protocol is T1, so that the test doesn't pass accidentally if the RAW protocol sticks by mistake.

third_party/pcsc-lite/naclport/server_clients_management/src/admin-policy-service.js
```diff
@@ -0,0 +1,207 @@
...
+AdminPolicyService.prototype.loadManagedStorage_ = function() {
+  chrome.storage.managed.get(this.managedStorageLoadedCallback_.bind(this));
+};
+
+/**
+ * @param {!Object} items
+ * @private
+ */
+AdminPolicyService.prototype.managedStorageLoadedCallback_ = function(items) {
+  goog.log.fine(
```
emaxx-google, line 133: nit: The "fine" log level is off in Release builds IIRC. As I think the logs are useful, let's use something like the INFO log level.

third_party/pcsc-lite/naclport/server_clients_management/src/admin-policy-service.js
```diff
@@ -0,0 +1,207 @@
...
+  chrome.storage.managed.onChanged.addListener(
+      this.storageChangedListener_.bind(this));
+};
+
+/**
+ * @param {!Object} changes
+ * @private
+ */
+AdminPolicyService.prototype.storageChangedListener_ = function(changes) {
+  if (this.isDisposed()) return;
```
emaxx-google, line 155: nit: Is this the style that clang-format uses? If not, then please reformat the code by running "./format-code.sh".

third_party/pcsc-lite/naclport/server_clients_management/src/admin_policy_getter.cc
```diff
@@ -0,0 +1,98 @@
...
+  return kUpdateAdminPolicyMessageType;
+}
+
+bool AdminPolicyGetter::OnTypedMessageReceived(Value data) {
+  AdminPolicy message_data;
+  std::string error_message;
+  if (!ConvertFromValue(std::move(data), &message_data, &error_message)) {
+    GOOGLE_SMART_CARD_LOG_WARNING << "Failed to parse admin policy message: "
+                                  << error_message;
+    return false;
```
emaxx-google, line 76: IIRC this will lead to a crash. I'm not sure it's the appropriate action, given that the format of the policy is received from Chrome, and we shouldn't blindly rely on the latter to do all the checks correctly.

Let's log the error, but still avoid the crash. Also makes sense to pretend that we received an (empty) policy value, so that `WaitAndGet()` callers are unblocked.

emaxx-google (changes requested):
Added a few nits too.

common/cpp/src/google_smart_card_common/admin_policy_getter.h
emaxx-google: Good for now, thanks.

emaxx-google:
The tests hang on bots. It seems the existing "SCardConnectProtocolMismatch" test should also simulate an (empty) policy receipt, so that the tweak in SCardConnect doesn't wait for the policy infinitely long.

BTW we might want to implement some kind of timeout, so that this doesn't happen in production... But let's not try tackling this in the current PR, since it's already big.

emaxx-google:
Hi, any help needed with this PR? I think you made great progress and we're close to completing it now.

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
```diff
@@ -410,33 +437,118 @@ GenericRequestResult PcscLiteClientRequestProcessor::SCardConnect(
   tracer.LogEntrance();
 
   LONG return_code = SCARD_S_SUCCESS;
-  if (!s_card_handles_registry_.ContainsContext(s_card_context))
+  if (update_card_handles_registry &&
```
emaxx-google, line 440: Is it correct that we guard this check by `update_card_handles_registry`? I think `SCARDCONTEXT` should always be valid.

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
```diff
@@ -400,6 +405,28 @@ GenericRequestResult PcscLiteClientRequestProcessor::SCardConnect(
...
+LONG PcscLiteClientRequestProcessor::SCardConnect(
+    SCARDCONTEXT s_card_context,
+    const std::string& reader_name,
+    DWORD share_mode,
+    DWORD preferred_protocols,
+    LPSCARDHANDLE s_card_handle,
+    LPDWORD active_protocol,
+    bool update_card_handles_registry,
+    bool handle_protocol_mismatch_error) {
   FunctionCallTracer tracer("SCardConnect", logging_prefix_,
```
emaxx-google, line 430: Not sure if it's the best place for this logging. Currently we log all calls exactly as they come, without any modifications. If we start logging "SCardConnect" in cases when the caller didn't actually call it, it'll make the investigation harder.

How about moving this `FunctionCallTracer` back to the original (public) `SCardConnect()` method?

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
```diff
@@ -410,33 +437,118 @@ GenericRequestResult PcscLiteClientRequestProcessor::SCardConnect(
...
+
+  optional<AdminPolicy> admin_policy = admin_policy_getter_->WaitAndGet();
+  std::vector<std::string> client_ids;
+  if (admin_policy &&
+      admin_policy.value().scard_disconnect_fallback_client_app_ids) {
+    client_ids =
+        admin_policy.value().scard_disconnect_fallback_client_app_ids.value();
+  }
+
+  if (std::find(client_ids.begin(), client_ids.end(),
```
emaxx-google, line 497: nit: Let's simplify this code to reduce the nesting level. Also we don't need to declare `return_code` that early:

```
  if (std::find(...) == client_ids.end()) {
    // The fallback is not enabled via policy, so return the original error.
    return SCARD_E_PROTO_MISMATCH;
  }
  ...
```

emaxx-google, line 497: nit: Let's simplify this code to reduce the nesting level. Also we don't need to declare `return_code` that early:

```
  if (std::find(...) == client_ids.end()) {
    // The fallback is not enabled via policy, so return the original error.
    return SCARD_E_PROTO_MISMATCH;
  }
  ...
```

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
```diff
@@ -410,33 +437,118 @@ GenericRequestResult PcscLiteClientRequestProcessor::SCardConnect(
...
+                this->client_name_for_log_) != client_ids.end()) {
+    GOOGLE_SMART_CARD_LOG_INFO
+        << logging_prefix_ << "SCardConnect failed with a protocol mismatch "
+        << "error. Attempting SCardDisconnect fallback: disconnecting and "
+        << "resetting any previous connections for context "
+        << DebugDumpSCardContext(s_card_context);
+
+    bool reset_successful =
+        SCardConnectAttemptReset(s_card_context, reader_name, share_mode);
```
emaxx-google, line 506: nit: Same here - we can simplify the code to:

```
  if (!SCardConnectAttemptReset()) {
    // The fallback failed, so return the original error.
    return SCARD_E_PROTO_MISMATCH;
  }
```

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
```diff
@@ -410,33 +437,118 @@ GenericRequestResult PcscLiteClientRequestProcessor::SCardConnect(
...
+    SCARDCONTEXT s_card_context,
+    const std::string& reader_name,
+    DWORD share_mode) {
+  // Try to get an s_card_handle by performing SCardConnect using any protocol
+  // and without checking or updating the card handle registry. Do not handle
+  // potential protocol mismatch errors since at this point the reset attempt is
+  // already doing that.
+  DWORD preferred_protocols =
+      SCARD_PROTOCOL_ANY | SCARD_PROTOCOL_RAW | SCARD_PROTOCOL_T15;
+  SCARDHANDLE s_card_handle;
```
emaxx-google, line 536: nit: Initialize this to `0`, so that the comparison in the `if` below is safer.

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
```diff
@@ -410,33 +437,118 @@ GenericRequestResult PcscLiteClientRequestProcessor::SCardConnect(
...
+  DWORD active_protocol;
+  LONG return_code_connect =
+      SCardConnect(s_card_context, reader_name, share_mode, preferred_protocols,
+                   &s_card_handle, &active_protocol,
+                   /*update_card_handles_registry=*/false,
```
emaxx-google, line 541: Is it really necessary to specify `update_card_handles_registry=false` here? IIUC if we don't do it then the handle added here would normally immediately removed in the call below.

If it's all a pure optimization, I'd rather simplify the code and delete the `update_card_handles_registry` parameter altogether.

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
```diff
@@ -410,33 +437,118 @@ GenericRequestResult PcscLiteClientRequestProcessor::SCardConnect(
...
+                   /*handle_protocol_mismatch_error=*/false);
```
emaxx-google, line 542: This recursive call makes the code a bit hard to follow. IIUC, the `handle_protocol_mismatch_error` parameter is also there to prevent infinite recursion here. I also don't like that we lost some checks/calls (like `OnSCardContextRevoked()`) in the new workaround's code path. What if we restructure the code a bit and streamline it?

```
LONG SCardConnect(...) {
  FunctionCallTracer ...
  LONG return_code = ObtainCardHandleWithFallback(...);
  tracer.LogExit();
  return return_code;
}

LONG ObtainCardHandleWithFallback(...) {
  LONG return_code = ObtainCardHandle(...);
  if (return_code == SCARD_E_PROTO_MISMATCH && IsDisconnectFallbackPolicyEnabled()) {
    if (ResetCard(...) && ObtainCardHandle(...)) {
      return SCARD_S_SUCCESS;
    }
  }
  return return_code;
}

LONG ObtainCardHandle(...) {
  LONG return_code = ::SCardConnect(...);
  if (return_code == SCARD_S_SUCCESS)
    s_card_handles_registry_.AddHandle(...);

  // Catch when PC/SC-Lite core, unlike us, thinks the context doesn't exist.
  if (return_code == SCARD_E_INVALID_HANDLE)
    OnSCardContextRevoked(s_card_context);

  return return_code;
}

bool ResetCard(...) {
  if (ObtainCardHandle(...) != SCARD_S_SUCCESS) {
    return false;
  }
  return SCardDisconnect(...) == SCARD_S_SUCCESS;
}
```

Would something like this work?

emaxx-google (changes requested):

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
```diff
@@ -435,10 +430,100 @@ GenericRequestResult PcscLiteClientRequestProcessor::SCardConnect(
...
+  optional<AdminPolicy> admin_policy = admin_policy_getter_->WaitAndGet();
+  std::vector<std::string> client_ids;
+  if (admin_policy &&
+      admin_policy.value().scard_disconnect_fallback_client_app_ids) {
+    client_ids =
+        admin_policy.value().scard_disconnect_fallback_client_app_ids.value();
+  }
+
+  return std::find(client_ids.begin(), client_ids.end(),
+                   this->client_name_for_log_) != client_ids.end();
```
emaxx-google, line 507: nit: Can `this->` be omitted here?

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
```diff
@@ -491,26 +576,31 @@ GenericRequestResult PcscLiteClientRequestProcessor::SCardDisconnect(
...
 
   tracer.AddReturnValue(DebugDumpSCardReturnCode(return_code));
   tracer.LogExit();
 
-  if (return_code == SCARD_S_SUCCESS)
-    s_card_handles_registry_.RemoveHandle(s_card_handle);
   return ReturnValues(return_code);
 }
 
+LONG PcscLiteClientRequestProcessor::DisconnectCard(LPSCARDHANDLE s_card_handle,
```
emaxx-google, line 587: nit: No need to pass the handle as a pointer here; let's use `SCARDHANDLE` instead.

emaxx-google (approved):
LGTM % nits. Thanks.

annklr:
Thanks for the reviews and all the improvements!

PR #706, opened 2022-06-24
Unit-test Transmit with a fake PIV card

emaxx-google:
Add a unit test for the SCardTransmit call from the JS side. To make the
testing closer to real-world use cases, we add a basic simulation of a
PIV card applet, and simulate the SELECT command in the test.

### DIFF
--- a/smart_card_connector_app/src/application_unittest.cc
+++ b/smart_card_connector_app/src/application_unittest.cc
@@ -382,6 +382,39 @@ class SmartCardConnectorApplicationTest : public ::testing::Test {
         ArrayValueBuilder().Add(scard_handle).Add(disposition).Get()));
   }
 
+  LONG SimulateTransmitCallFromJsClient(
+      int handler_id,
+      SCARDHANDLE scard_handle,
+      DWORD send_protocol,
+      const std::vector<uint8_t>& data_to_send,
+      optional<DWORD> receive_protocol,
+      DWORD& out_response_protocol,
+      std::vector<uint8_t>& out_response) {
+    Value receive_protocol_arg;
+    if (receive_protocol) {
+      receive_protocol_arg =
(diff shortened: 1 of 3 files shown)


smart_card_connector_app/src/application_unittest.cc
```diff
@@ -1175,4 +1208,58 @@ TEST_F(SmartCardConnectorApplicationSingleClientTest, DisconnectAfterRemoving) {
...
+    DWORD response_protocol = 0;
+    EXPECT_EQ(
+        SimulateTransmitCallFromJsClient(
+            kFakeHandlerId, scard_handle, SCARD_PROTOCOL_T1, kSelectCommand,
+            /*receive_protocol=*/{}, response_protocol, response),
+        SCARD_S_SUCCESS);
+    EXPECT_EQ(response_protocol, static_cast<DWORD>(SCARD_PROTOCOL_T1));
+    // The expected result should contain the application identifier followed by
+    // 0x90 0x00 (denoting a successful operation).
+    std::vector<uint8_t> expected_response =
```
Fabian-Sommer, line 1251: In combination with the simulation which also calls `TestingSmartCardSimulation::GetCardProfileApplicationIdentifier` this feels a bit like doing `EXPECT_EQ(f(), f())`, but I don't have a idea how to test this in a better way either

emaxx-google: Yeah, I'm on the fence as well. At least it checks that the response correctly got through all the layers of the driver (which contains a lot of low-level handling).

I remember in some of the previous PRs there was a similar case where I initially hardcoded the same value in two places and we decided to share the constant. Maybe here in the current PR a duplicated constant would be more warranted.

emaxx-google:
Let me land it - if we figure out a better approach later we can update the code anytime.

PR #714, opened 2022-08-06
USB transient error workaround for claimInterface

emaxx-google:
Extend the workaround for transient USB errors to also cover the
failures in claimInterface. This fixes flaky issues with the Smart Card
Connector on the ChromeOS Lock Screen, in case when the Connector is
also installed in-session and happens to be slow to shutdown USB
connections when the session is being locked.

For the background, this workaround existed for errors in the openDevice
function since a long time. By default, PC/SC-Lite doesn't retry the USB
connections, and adding them would introduce a complex patch that's hard
to maintain. So we implemented a trick-based workaround: each time our
Libusb wrapper sees an error, it increments the "bus_number" field for
the next time this device is listed (this field's value is anyway a
complete fake as Chrome USB APIs never exposed the real bus numbers).
This makes PC/SC-Lite retry the connection next time it polls the list
of devices (which happens every couple of seconds) as it thinks it's a
new device that appeared.

Internal tracking bug: [b/241549122](b/241549122).

### DIFF
--- a/third_party/libusb/webport/src/libusb_js_proxy_unittest.cc
+++ b/third_party/libusb/webport/src/libusb_js_proxy_unittest.cc
@@ -463,6 +463,104 @@ TEST_P(LibusbJsProxyTest, DeviceRefUnref) {
   FreeLibusbDevices(devices);
 }
 
+// Test that `LibusbGetBusNumber()` initially returns the default bus number.
+TEST_P(LibusbJsProxyTest, BusNumber) {
+  // Arrange.
+  global_context()->WillReplyToRequestWith(
+      "libusb", "listDevices",
+      /*arguments=*/Value(Value::Type::kArray),
+      /*result_to_reply_with=*/
+      ArrayValueBuilder()
+          .Add(DictValueBuilder()
+                   .Add("deviceId", 123)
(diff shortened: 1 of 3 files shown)


third_party/libusb/webport/src/libusb_js_proxy_unittest.cc
```diff
@@ -616,6 +716,75 @@ TEST_P(LibusbJsProxyWithDeviceTest, DeviceResettingFailure) {
...
+
+  // Act.
+  EXPECT_EQ(libusb()->LibusbClaimInterface(device_handle_, kInterfaceNumber),
+            LIBUSB_ERROR_OTHER);
+  // Obtain the list of devices again.
+  std::vector<libusb_device*> devices = GetLibusbDevices();
+  ASSERT_EQ(devices.size(), 1U);
+
+  // Assert. The bus number should be incremented now.
+  EXPECT_EQ(libusb()->LibusbGetBusNumber(devices[0]), 2);
```
Fabian-Sommer, line 783: nit: I would add an assertion before the call that the BusNumber is 1, to illustrate the assertion

Fabian-Sommer (approved):
LGTM

PR #716, opened 2022-08-08
Log USB transfer errors

emaxx-google:
Print a log message when a USB transfer function
(control/bulk/interrupt) returns an error. This will simplify
investigating bug reports, at the cost of extra verbosity as some errors
(e.g., when a device is unplugged) are expected.


Fabian-Sommer (approved):
LGTM

PR #718, opened 2022-08-09
Fix spurious openct warnings

emaxx-google:
Fix spurious error logs about attempting to read a non-existing
"/var/run/openct/status" file.

This file was read by the PC/SC-Lite as an attempt to troubleshoot USB
errors, because the OpenCT daemon on Linux grabs access to USB devices
and therefore conflicts with the PC/SC-Lite daemon. These checks don't
make sense in ChromeOS, and while harmless they were triggering file
access error logs in the JavaScript console. The commit suppresses all
this by substituting the original CheckForOpenCT() function with our
own stub.


Fabian-Sommer (approved):
LGTM

PR #720, opened 2022-08-10
Declare Alcor 95x0 readers supported

emaxx-google:
Declare Alcor 9540/9560 readerrs as supported, so that the Connector app
attempts to connect to them automatically, without an extra user gesture
("add reader").

Meanwhile the CCID driver's website and some internet sources claim this
reader has issues, we haven't observed them. With the current state of
things, it's better to allow such a reader to reduce user hurdles (and
if the user does experience issues with it they can always uninstall the
Connector app).

We call the reader as "95x0" in the config, because both 9540 and 9560
share the same (vendorID; productID) tuples, which makes the CCID driver
unable to distinguish between them; hence we use a string that would
work in the UI for either of them.

Chromium tracking bug: https://crbug.com/1337303.


Fabian-Sommer (approved):
LGTM

PR #722, opened 2022-08-10
Avoid warning logs on USB device unplugging

emaxx-google:
Downgrade a bunch of logs to the INFO level, so that when a smart card
reader is unplugged there are no warnings/errors appearing in the
application log. This is mainly to avoid confusing users who see a red
"Errors" button in chrome://extensions whenever we log a warning/error
and might interpret it as a real problem (which in this case it's not).


smart_card_connector_app/src/manifest.json.template
```diff
@@ -1,7 +1,5 @@
 {
-${if CONFIG=Debug}
```
Fabian-Sommer, line 2: Is this change intentional?

Fabian-Sommer (approved):
LGTM

smart_card_connector_app/src/manifest.json.template
emaxx-google: Oops. Thanks.

PR #723, opened 2022-08-10
Fix application_unittest setup under PNaCl

emaxx-google:
Fix crashes in smart_card_connector_app's application unittests due to
attempts to repeatedly remount / and /tmp directories. Before this
commit, only one test was passing and the subsequent ones were crashing
due to the leftover mounts.

Now we're unmounting the directories in each test's teardown. Also we
ignore the unmount() errors because they're fine in these test
scenarios, and in general there's little value to crash on these.

### DIFF
--- a/common/cpp/src/google_smart_card_common/nacl_io_utils.cc
+++ b/common/cpp/src/google_smart_card_common/nacl_io_utils.cc
@@ -50,7 +50,13 @@ void MountNaclIoFolders() {
 }
 
 bool UnmountNaclIoFolders() {
-  return ::umount("/tmp") == 0 && ::umount("/") == 0;
+  bool success = true;
+  // Try unmounting both even if one failed.
+  if (::umount("/") != 0)
+    success = false;
+  if (::umount("/tmp") != 0)
+    success = false;
+  return success;
 }
 
 }  // namespace google_smart_card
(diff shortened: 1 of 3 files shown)


common/cpp/src/google_smart_card_common/nacl_io_utils.cc
```diff
@@ -40,12 +40,19 @@ void InitializeNaclIo(const pp::Instance& pp_instance) {
...
+  ::umount("/tmp");
 
   GOOGLE_SMART_CARD_CHECK(
       ::mount("/", "/", "httpfs", 0, "manifest=/nacl_io_manifest.txt") == 0);
 
   GOOGLE_SMART_CARD_CHECK(::mount("", "/tmp", "memfs", 0, "") == 0);
 }
 
+void UnmountNaclIoFolders() {
+  GOOGLE_SMART_CARD_CHECK(::umount("/tmp") == 0);
```
Fabian-Sommer, line 54: Since this is still test only, does crashing make sense?

Fabian-Sommer (approved):
LGTM

common/cpp/src/google_smart_card_common/nacl_io_utils.cc
emaxx-google: Done - Changed this to return a `bool` and to `EXPECT_TRUE` its result in the test teardown.

PR #724, opened 2022-08-24
Clean up PC/SC request receiver JS logs

emaxx-google:
Shorten some log messages printed by the JavaScript client-handler and
request-receiver code. Downgrade some of them to "info" or "fine" levels
to avoid spamming the log dumps with spurious errors and stop confusing
users who see "errors" displayed at chrome://extensions.

This commit contributes to the log clutter reduction tracked by #146.


Fabian-Sommer (approved):
LGTM

PR #726, opened 2022-08-25
Refactor test-all.sh to use make-all.sh

emaxx-google:
There's no need to duplicate the functionality across these two scripts,
and there were subtle differences that caused errors.

Testing is the same as compiling, except that we specify the "test"
target.


Fabian-Sommer (approved):
LGTM

PR #729, opened 2022-09-02
Unit-test concurrent card connections

emaxx-google:
Add unit test coverage for the case when two clients concurrently
connect to the same reader and card. One test checks successful scenario
(when sharing is allowed), and another checks failure scenario (when one
client holds exclusive access).


Fabian-Sommer (approved):
LGTM

PR #732, opened 2022-09-02
Skip compiling strlcpy.c

emaxx-google:
This upstream file is not really needed in our builds.


emaxx-google:
Apparently ASan builds need it. Getting rid of this dependency should be possible, but I don't want to make changes in third-party code just because of that.

PR #735, opened 2022-09-05
Fix NaN timestamp in logs from window.html

emaxx-google:
This commit removes the passing of goog.log.LogRecord objects between
windows, replacing it with passing their fields separately (message,
timestamp, etc.). The direct passing of objects can never work reliably
in Release mode due to Closure Compiler's optimizations that can rename
object fields differently for different targets.

The symptom of the problem addressed by this commit is that the log
messages from the Smart Card Connector's UI (window.html) had "NaN"
timestamps.


Fabian-Sommer (approved):
LGTM

PR #738, opened 2022-09-05
Add new paths to env/clean.sh

emaxx-google:
There are a few new paths that we initialize in env/initialize.sh these
days, and they should also be removed by env/clean.sh.


Fabian-Sommer (approved):
LGTM

PR #740, opened 2022-09-26
Add missing emscripten/version.h include

emaxx-google:
This fixes the compilation issue with Emscripten 3.1.23: the
__EMSCRIPTEN_major__ macro is undefined by default now, and
our code wrongly assumed that the Emscripten version is too old.


Fabian-Sommer (approved):
LGTM

PR #745, opened 2022-09-26
Unit-test SCard calls with wrong handles

emaxx-google:
Add test coverage for calling various PC/SC-Lite functions with wrong contexts or handles.


Fabian-Sommer (approved):
LGTM, but is deletion of the BeginTransaction test intentional?

emaxx-google:
> LGTM, but is deletion of the BeginTransaction test intentional?

Oops, good catch! Fixed.

PR #747, opened 2022-09-27
Unit-test SCardGetStatusChange with card removal

emaxx-google:
Add unit test coverage for the SCardGetStatusChange PC/SC-Lite function in the scenario when the initially present card gets removed.


Fabian-Sommer (approved):
LGTM

PR #751, opened 2022-09-30
Enable ACS ACR122U PICC reader

### LINKED ISSUE #739: Manifest.json Missing { "vendorId": 1839, "productId": 8704 }
We have a reader that works in a chromebook, namely the acr122u. We can extract custom data from nfc cards via the chromeos_smart_card_connector (https://chrome.google.com/webstore/detail/smart-card-connector/khpfeaanjngmcnplbdlpegiifgpfgdco?hl=nl) app in combination with the library here provided, implemented in our custom chrome app. 

But we have trouble using it on chrome enterprise kiosk. It should detect the reader immideatly, but somehow it doesn't, it doesn't even burn the light on the reader. It works outside the chrome enterprise kiosk though.

We think this is because the vendor id product id is not in the manifest.json. { "vendorId": 1839, "productId": 8704 } is missing, are we correct?

emaxx-google:
Allowlist the ACS ACR122U PICC reader (vendorID=0x072F productID=0x2200). This allows the Smart Card Connector to talk to this reader without any extra steps (neither the user having clicking "add reader" nor the admin configuring this via policy). This is especially useful when used in constrained environments like Kiosk secondary apps.

Note that the reader has been previously explicitly marked as unsupported by the CCID Free Driver:
https://ccid.apdu.fr/ccid/unsupported.html#0x072F0x2200 . However, we got aware that there are customer deployments that won't be affected by the issues described on that page, and which want to have this reader be enabled by default.

This fixes #739.


Fabian-Sommer (approved):
LGTM (A comment to explain this would be nice if comments are possible in this file)

emaxx-google:
> LGTM (A comment to explain this would be nice if comments are possible in this file)

Good point. Published #752 for adding the support of it.

emaxx-google:
Added a comment.

PR #752, opened 2022-09-30
Add comments to manifest.json source template

emaxx-google:
Allow writing comments in the manifest.json.template file which is used to generate the Smart Card Connector's manifest.json.

Comments will help documenting how and why various things are configured in the manifest.


Fabian-Sommer (approved):
LGTM

PR #753, opened 2022-10-04
Allowlist Ujambo Kiosk apps

emaxx-google:
Add the Kiosk apps from Ujambo (in its different flavors: production, development, "acceptance") to the list of trusted client IDs, and autoallow the permission to talk to smart cards.

This allows, in particular, to use the Smart Card Connector together with them as a secondary kiosk app (in this mode it has neither any UI nor access to admin policies).


Fabian-Sommer (approved):
LGTM

PR #756, opened 2022-10-04
Fix client lib when there's no bg page

### LINKED ISSUE #755: getBackgroundPage error isn't caught
trying to convert Secure Shell to MV3 and i'm seeing a crash.  i'm using v1.3.9.3 here.
```
00:40:00.842 google-smart-card-client-library.js:80
[221005 00:40:00.83] [] Failure
Gc @ google-smart-card-client-library.js:80

00:40:00.848 nassh.html:1
Error handling response: Error: Failure
    at Mc (chrome-extension://algkcnfjnajfhgimadimbjhmpaeohhln/third_party/google-smart-card/google-smart-card-client-library.js:82:215)
    at chrome-extension://algkcnfjnajfhgimadimbjhmpaeohhln/third_party/google-smart-card/google-smart-card-client-library.js:80:1018

00:40:00.848 nassh.html:1
Unchecked runtime.lastError: You do not have a background page.
```

afaik, Chrome requires you to read `chrome.runtime.lastError` otherwise it throws an exception like this.
```
chrome.runtime.getBackgroundPage((bg) => {
  if (bg === undefined) {
    const error = chrome.runtime.lastError?.message;
    ...
    return;
  }
  ...
});
```

emaxx-google:
Fix the following crash when running our JavaScript client library in a context that has chrome.runtime.* APIs but there's no background page:

  Unchecked runtime.lastError: You do not have a background page.

We should just gracefully exit the callback in that case: this code is trying to set up log forwarding from foreground to background pages, which is not needed in this case.

This fixes #755.

### DIFF
--- a/common/js/src/logging/logging.js
+++ b/common/js/src/logging/logging.js
@@ -353,9 +353,14 @@ function setupLogBuffer() {
   goog.global[GLOBAL_LOG_BUFFER_VARIABLE_NAME] = logBuffer;
 
   chrome.runtime.getBackgroundPage(function(backgroundPage) {
-    GSC.Logging.check(backgroundPage);
-    goog.asserts.assert(backgroundPage);
-
+    if (!backgroundPage) {
+      // There's no background page - it's expected when we're running inside an
+      // extension without a background page. No action needed for log buffers.
+      goog.log.fine(
+          rootLogger,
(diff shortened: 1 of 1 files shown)


Fabian-Sommer (approved):
LGTM

PR #760, opened 2022-10-25
Migrate LogBuffer to ES6 class

emaxx-google:
Migrate the //common/js/src/logging/log-buffer.js file and child classes that depend on it from legacy Closure Compiler classes to ES6 classes.

This is part of the effort to modernize the code base, as recent Closure Compiler versions stopped supporting the legacy class syntax.


Fabian-Sommer (approved):
LGTM

emaxx-google:
Thanks. I've pushed another commit that applies clang-tidy (I intentionally didn't do it initially, to make the reviewed diff smaller).

PR #762, opened 2022-10-25
Migrate MessageChannelPair to ES6 class

emaxx-google:
Migrate the //common/js/src/messaging/message-channel-pair.js file from legacy Closure Compiler classes to ES6 classes.

This is a pure refactoring change. It's is part of the effort to modernize the code base, as recent Closure Compiler versions stopped supporting the legacy class syntax.


emaxx-google:
The first commit is the actual change, and the second commit is reapplying clang-format.

Fabian-Sommer (approved):
LGTM

Issue #136, opened 2020-07-03
Catch and abort illegal parallel PC/SC requests early


emaxx-google:
The PC/SC API requires that every thread of the application uses its own context:

> Each thread of an application shall use its own SCARDCONTEXT, unless calling SCardCancel(), which MUST be called with the same context as the context used to call SCardGetStatusChange().

(https://pcsclite.apdu.fr/api/group__API.html#gaa1b8970169fd4883a6dc4a8f43f19b67)

Our implementation in the Smart Card Connector app retains this requirement. However, it's not currently enforced by the client-side libraries that we provide (neither JS nor C++ ones), which means that it can be violated without the developer being notified about that.

Practically, violating this requirement - e.g., sending a new PC/SC call before the previous one with the same hContext finished - can cause bad effects, like the calls being hung. Also - although not confirmed yet - it might lead to the hang of the Smart Card Connector app.

The proposal is to add debug assertions into our C++ and/or JS libraries that would catch these kinds of violations. Perhaps, crashing the caller is the only reasonable and effective measure in that case (note that typical clients would automatically restart after the crash, which hopefully will recover the state and let the user continue using the smart card).

As a stretch goal, we might want to investigate whether/why the Smart Card Connector app hangs for *all* clients in this scenario, and when that's the case try to fix that behavior.

PR #782, opened 2022-12-08
Connector warns on concurrent PC/SC calls

emaxx-google:
Add warning logs into Smart Card Connector whenever a client application emits illegal concurrent PC/SC requests: an SCARDCONTEXT must only be used from a single thread (with the only exception of SCardCancel that can be called concurrently).

This is preparation for disallowing concurrent PC/SC calls in the future as planned in #136. Concurrent calls are not allowed by the PC/SC API specification, and the PC/SC-Lite's implementation breaks in subtle ways when this is violated.

### DIFF
--- a/third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
+++ b/third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
@@ -213,8 +213,6 @@ void PcscLiteClientRequestProcessor::ProcessRequest(
         << "\"";
   }
 
-  LeaveConcurrencyCheckScope(request.function_name);
-
   result_callback(std::move(result));
 }
 
@@ -223,15 +221,62 @@ void PcscLiteClientRequestProcessor::AsyncProcessRequest(
     std::shared_ptr<PcscLiteClientRequestProcessor> request_processor,
     RemoteCallRequestPayload request,
     RequestReceiver::ResultCallback result_callback) {
-  // This is paired with the call to `LeaveConcurrencyCheckScope()` in
-  // `ProcessRequest()`. We enter the scope here in order to be able to catch
-  // concurrency violations even when they occur so fast that the thread created
-  // below didn't have chance to start and do any work.
-  request_processor->EnterConcurrencyCheckScope(request.function_name);
-
   std::thread(&PcscLiteClientRequestProcessor::ProcessRequest,
               request_processor, std::move(request), result_callback)
       .detach();
+}
+
+PcscLiteClientRequestProcessor::ScopedConcurrencyGuard::ScopedConcurrencyGuard(
+    const std::string& function_name,
+    SCARDCONTEXT s_card_context,
+    SCARDHANDLE s_card_handle,
+    PcscLiteClientRequestProcessor& owner)
+    : owner_(owner),
+      function_name_(function_name),
+      s_card_context_(s_card_context) {
+  if (IsConcurrentCallAllowed(function_name_)) {
+    return;
+  }
+  s_card_context_ =
+      s_card_handle
+          ? owner_.s_card_handles_registry_.FindContextByHandle(s_card_handle)
+          : s_card_context;
+  if (!s_card_context_) {
+    return;
+  }
+
+  std::string concurrent_functions_dump;
+  {
+    const std::unique_lock<std::mutex> lock(owner_.mutex_);
+    std::multiset<std::string>& running_functions =
+        owner_.context_to_running_functions_[s_card_context_];
+    concurrent_functions_dump =
+        JoinStrings(running_functions, /*separator=*/", ");
+    running_functions.insert(function_name_);
+  }
+
+  if (!concurrent_functions_dump.empty()) {
+    GOOGLE_SMART_CARD_LOG_WARNING
+        << owner_.logging_prefix_
+        << "Client violates threading: concurrent calls of " << function_name_
+        << ", " << concurrent_functions_dump
+        << ". Future releases of Smart Card Connector will forbid this.";
+  }
+}
+
+PcscLiteClientRequestProcessor::ScopedConcurrencyGuard::
+    ~ScopedConcurrencyGuard() {
+  if (IsConcurrentCallAllowed(function_name_) || !s_card_context_) {
+    return;
+  }
+  const std::unique_lock<std::mutex> lock(owner_.mutex_);
+  auto iter = owner_.context_to_running_functions_.find(s_card_context_);
+  GOOGLE_SMART_CARD_CHECK(iter != owner_.context_to_running_functions_.end());
+  std::multiset<std::string>& running_functions = iter->second;
+  running_functions.erase(running_functions.find(function_name_));
+  if (running_functions.empty()) {
+    owner_.context_to_running_functions_.erase(iter);
+  }
 }
 
 void PcscLiteClientRequestProcessor::BuildHandlerMap() {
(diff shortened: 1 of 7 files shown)


Fabian-Sommer (approved):
LGTM - intuitively it feels wrong to copy a vector of strings just to concatenate them later, but I couldn't see a better way either (especially since the vector is supposed to be empty in general)

emaxx-google:
> LGTM - intuitively it feels wrong to copy a vector of strings just to concatenate them later, but I couldn't see a better way either (especially since the vector is supposed to be empty in general)

Fair point. It was indeed pretty wasteful, because we needed to copy it and because of a costly `std::ostringstream`.

I've changed the code to directly concatenate into a single string.

emaxx-google:
> LGTM - intuitively it feels wrong to copy a vector of strings just to concatenate them later, but I couldn't see a better way either (especially since the vector is supposed to be empty in general)

Fair point. It was indeed pretty wasteful, because we needed to copy it and because of a costly `std::ostringstream`.

I've changed the code to directly concatenate into a single string.

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
```diff
@@ -1142,25 +1142,21 @@ void PcscLiteClientRequestProcessor::EnterConcurrencyCheckScope(
...
   }
 
-  std::ostringstream str;
-  for (const auto& name : concurrent_functions) {
-    str << ", " << name;
-  }
   GOOGLE_SMART_CARD_LOG_WARNING
       << logging_prefix_ << "Client violates threading: concurrent calls of "
-      << starting_function_name << str.str()
+      << starting_function_name << concurrent_functions
```
Fabian-Sommer, line 1159: nit: We need a separator between these two since concurrent_functions no longer starts with one

Fabian-Sommer (approved):
LGTM

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
emaxx-google: Done - good catch!

emaxx-google:
Oops - I realized that a single `ClientRequestProcessor` can be used with multiple `SCARDCONTEXT`s, and that's fine. We should only complain about concurrent calls that share the same `SCARDCONTEXT`. This resulted in having to redo a big chunk of the implementation.

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
```diff
@@ -223,17 +213,63 @@ void PcscLiteClientRequestProcessor::AsyncProcessRequest(
...
 
+PcscLiteClientRequestProcessor::ScopedConcurrencyGuard::ScopedConcurrencyGuard(
+    const std::string& function_name,
+    SCARDCONTEXT s_card_context,
+    SCARDHANDLE s_card_handle,
+    PcscLiteClientRequestProcessor& owner)
+    : owner_(owner),
+      function_name_(function_name),
+      s_card_context_(s_card_context) {
+  s_card_context_ =
```
Fabian-Sommer, line 229: I think s_card_context_ should only be set in one place (unless there is a requirement that the default initializers are always used)

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
```diff
@@ -223,17 +213,63 @@ void PcscLiteClientRequestProcessor::AsyncProcessRequest(
...
+  {
+    const std::unique_lock<std::mutex> lock(owner_.mutex_);
+    std::multiset<std::string>& running_functions =
+        owner_.context_to_running_functions_[s_card_context_];
+    concurrent_functions_dump =
+        JoinStrings(running_functions, /*separator=*/", ");
+    running_functions.insert(function_name_);
+  }
+
+  if (!concurrent_functions_dump.empty()) {
```
Fabian-Sommer, line 247: This behaves wrong if we ever have a function with an empty string as a name, and it is the only running function. We should either check for empty names, or just use a separate bool to ascertain there are no running functions.

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.h
```diff
@@ -145,6 +145,26 @@ class PcscLiteClientRequestProcessor final
       std::function<GenericRequestResult(std::vector<Value> arguments)>;
   using HandlerMap = std::unordered_map<std::string, Handler>;
 
+  class ScopedConcurrencyGuard final {
+   public:
+    ScopedConcurrencyGuard(const std::string& function_name,
```
Fabian-Sommer, line 150: There should be some comment about this class/the constructor - e.g. I was surprised to read that s_card_context is no more than the default for s_card_context_, but usually it is set in a different manner.

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
```diff
@@ -223,17 +213,63 @@ void PcscLiteClientRequestProcessor::AsyncProcessRequest(
...
+PcscLiteClientRequestProcessor::ScopedConcurrencyGuard::
+    ~ScopedConcurrencyGuard() {
+  if (!s_card_context_) {
+    return;
+  }
+  const std::unique_lock<std::mutex> lock(owner_.mutex_);
+  auto iter = owner_.context_to_running_functions_.find(s_card_context_);
+  GOOGLE_SMART_CARD_CHECK(iter != owner_.context_to_running_functions_.end());
+  std::multiset<std::string>& running_functions = iter->second;
+  running_functions.erase(running_functions.find(function_name_));
```
Fabian-Sommer, line 267: This would erase all entries with the given name, even though only one should be erased (disregarding that we are already in a non-supported state at this point)

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
```diff
@@ -215,6 +218,58 @@ void PcscLiteClientRequestProcessor::AsyncProcessRequest(
...
+  if (running_functions.empty()) {
+    owner_.context_to_running_functions_.erase(iter);
```
Fabian-Sommer, line 269: Where is this actually initialized? I can't find it

Fabian-Sommer (changes requested):

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
emaxx-google: Done. To reduce confusion, I removed the `s_card_handle` parameter: now the callsites translate it into `s_card_context`.

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.h
emaxx-google: Done added a comment.

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
emaxx-google: Done - Used an `optional<>` here. (We couldn't have empty names today, but I agree it was fragile.)

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
emaxx-google: Hmm I think it only delete exactly one item because we pass an iterator. WDYT?

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
emaxx-google: `iter` is above if that's what you mean.

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
Fabian-Sommer: You are right, I did not see the find() call in there

third_party/pcsc-lite/naclport/server_clients_management/src/client_request_processor.cc
Fabian-Sommer: I was referring to the multiset referenced by context_to_running_functions_, but on further review initialization happens in line 236

Fabian-Sommer (approved):
LGTM

PR #783, opened 2022-12-08
Update CCID to 1.5.1

emaxx-google:


emaxx-google:
Only the second commit is to be reviewed - the first one is a pure copy of the upstream.

Fabian-Sommer (approved):
LGTM

PR #803, opened 2023-06-20
Write Connector logs to systemLog

emaxx-google:
Use chrome.systemLog, when it's available (on ChromeOS >=111), in order to include Smart Card Connector's logs into the system logs.

The main benefit is that it makes the smart card logs being present in Feedback Reports that users send from Chromebooks, simplifying interaction with affected users and making it possible to investigate login-screen issues.


Fabian-Sommer (approved):
LGTM % nit

PR #804, opened 2023-06-21
Downgrade GetActiveConfigDescriptor warning log

emaxx-google:
Make following log debug-only instead of a WARNING:

  "LibusbGetActiveConfigDescriptor request failed: No active config
  descriptors were returned by JS API"

This log can be triggered when the USB device disappears suddenly (e.g., when transitioning to the ChromeOS Lock Screen). This is normal and doesn't deserve a warning log (which, e.g., results in showing a red "Errors" button at chrome://extensions).


Fabian-Sommer (approved):
LGTM

PR #805, opened 2023-06-23
Change typings for atr parameter in API. 

ViktoriiaKovalova:


ViktoriiaKovalova:
I didn't check what is actually returned for other functions but looking at #222, I guess in other places `!Array.<number>` is also used incorrectly.

emaxx-google (approved):
I've spent some time looking how we can write a test for this change, but seems like a nontrivial investment in the moment (as we'd need an integration-like C++-to-JS test, which we do have but only for a different component, and this only for NaCl which we cannot test automatically on bots).

But I've performed a manual test by adding test assertions, and the change seems to be correct. Hence LGTM.

PR #808, opened 2023-06-26
Bulletproof C++-to-JS test's tearDown

emaxx-google:
Avoid crashing in the integration test's tearDown in case setUp didn't succeed and the binary module wasn't loaded successfully.

We shouldn't attempt to make C++ calls in case the module isn't loaded.


Fabian-Sommer (approved):
LGTM

PR #809, opened 2023-06-26
Bail out loading NaCl module if unsupported env

emaxx-google:
Detect the case when the NaCl module loading was silently ignored, and return an explicit error. This allows to clearly indicate error, e.g., when running tests in "pnacl" mode on desktop Chrome, as opposed to behaving as if the initialization takes infinitely long time.

We detect the error case by checking whether the `<embed>` that we create actually has NaCl-specific properties populated by the browser, i.e., "postMessage". That seems to be the simplest programmatic detection.


Fabian-Sommer (approved):
LGTM

PR #817, opened 2023-06-30
Use RUNTIME_DEBUG=0 in Emscripten+Debug

emaxx-google:
Pass "-s RUNTIME_DEBUG=0" when linking Emscripten
code in Debug mode.

This is to suppress the very verbose logging that
recent versions of Emscripten runtime emit on some internal events (e.g., each time stack size changes). The RUNTIME_DEBUG flag, if not explicitly specified, gets auto-enabled when other *_DEBUG flags are,
which happened in our case as we use EXCEPTION_DEBUG.


ViktoriiaKovalova (approved):
Thanks 👍

PR #819, opened 2023-07-04
Don't rely on DEPS and dependency.mk for build

emaxx-google:
Delete the vast majority of DEPS and all dependency.mk files. Originally, when this project started, these served the purpose of allowing to build each subdirectory in isolation.

However, this proved to be error-prone and cause a lot of overhead in build times, because transitive dependencies got inspected by Make multiple times. Hence we migrated the project to use a top-level Makefile that always serves as an entry point and is the source of truth for build dependencies between subdirectories. This past change effectively made most of DEPS directives useless.

The commit removes them, resulting not only in simpler build scripts but also in better build times.


Fabian-Sommer (approved):
LGTM

PR #824, opened 2023-07-18
Fix dispose called on null during ClientHandler disposal

### LINKED ISSUE #823: JS error when NaCl crashes and there's active client
Just observed this on a test device after the Connector's NaCl module crashed (for an unrelated reason):

```
Uncaught TypeError: Cannot read properties of null (reading 'g')
    at I (background.js:101:481)
    at kl.A (background.js:456:49)
    at I (background.js:101:494)
    at background.js:457:296
    at H.A (background.js:101:911)
    at fl.A (background.js:451:41)
    at I (background.js:101:494)
    at background.js:450:325
    at H.A (background.js:101:911)
    at h.A (background.js:178:175)
```

The stack trace is obfuscated due to minification, but after comparing the minified and original code I think it's crashing in `ClientHandler.disposeInternal()`: https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/blob/f4eb32264c2c2b7afe57db0cf8ee280c98ce9ae3/third_party/pcsc-lite/naclport/server_clients_management/src/client-handler.js#L211

ViktoriiaKovalova:
This fixes #823.

### DIFF
--- a/third_party/pcsc-lite/naclport/server_clients_management/src/client-handler-unittest.js
+++ b/third_party/pcsc-lite/naclport/server_clients_management/src/client-handler-unittest.js
@@ -19,6 +19,7 @@ goog.require('GoogleSmartCard.PcscLiteServerClientsManagement.ClientHandler');
 goog.require('goog.Promise');
 goog.require('goog.reflect');
 goog.require('goog.testing');
+goog.require('goog.testing.asserts');
 goog.require('goog.testing.MockControl');
 goog.require('goog.testing.jsunit');
 goog.require('goog.testing.messaging.MockMessageChannel');
@@ -44,6 +45,18 @@ let mockControl;
 let mockServerMessageChannel;
 let mockClientMessageChannel;
 const stubPermissionsChecker = new StubPermissionsChecker();
+
+/**
+ * @param {!goog.testing.messaging.MockMessageChannel} mockMessageChannel
+ */
+function disposeOfMockMessageChannel(mockMessageChannel) {
+  mockMessageChannel.dispose();
+  // Hack: call the protected disposeInternal() method in order to trigger the
+  // disposal notifications; the MockMessageChannel.dispose() doesn't call
+  // them.
+  mockMessageChannel[goog.reflect.objectProperty(
+      'disposeInternal', mockMessageChannel)]();
+}
 
 goog.exportSymbol('testClientHandler', {
   'setUp': function() {
@@ -75,12 +88,7 @@ goog.exportSymbol('testClientHandler', {
     const handler = new ClientHandler(
         mockServerMessageChannel, goog.Promise.resolve(),
         mockClientMessageChannel, EXTENSION_ID_A);
-    mockClientMessageChannel.dispose();
-    // Hack: call the protected disposeInternal() method in order to trigger the
-    // disposal notifications; the MockMessageChannel.dispose() doesn't call
-    // them.
-    mockClientMessageChannel[goog.reflect.objectProperty(
-        'disposeInternal', mockClientMessageChannel)]();
+    disposeOfMockMessageChannel(mockClientMessageChannel);
     assert(
         'ClientHandler should dispose on client channel disposal',
         handler.isDisposed());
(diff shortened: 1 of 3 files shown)


emaxx-google:
I'm wondering now whether the `this.serverRequester_ = null` assignment is needed in the first place. Can't we just drop it?

third_party/pcsc-lite/naclport/server_clients_management/src/client-handler.js
```diff
@@ -301,11 +302,6 @@ ClientHandler.prototype.serverRequesterDisposedListener_ = function() {
   if (this.isDisposed())
```
emaxx-google, line 302: nit: This check is not needed anymore as well.

third_party/pcsc-lite/naclport/server_clients_management/src/server-request-handler.js
```diff
@@ -183,7 +183,9 @@ Pcsc.ServerRequestHandler = class extends goog.Disposable {
 
   /** @override */
   disposeInternal() {
-    super.disposeInternal();
+    if (this.isDisposed()) {
```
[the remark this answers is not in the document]
ViktoriiaKovalova: thanks!

third_party/pcsc-lite/naclport/server_clients_management/src/client-handler.js
ViktoriiaKovalova: I don't understand this. Why?

emaxx-google: We can just call `this.dispose()`. Same as in the other comment thread, it can be called multiple times, but it does internal bookkeeping in order to call `disposeInternal()` only once.

ViktoriiaKovalova: ah, thanks, i was looking at the wrong part of the code

emaxx-google:
Thanks for the fix. Let me take a quick look at how we could test this kind of regression fix. This piece of code currently doesn't have test coverage, so I'll try to write the first example.

emaxx-google:
Alright, so with #831 landed I think it should be possible to write a regression test. The test would create the client handler (similarly to `testSmoke`), then shut down `mockServerMessageChannel` and verify the client handler got disposed of.

ViktoriiaKovalova:
I added tests and checked that testServerChannelDisposed breaks without this change :)

third_party/pcsc-lite/naclport/server_clients_management/src/client-handler-unittest.js
```diff
@@ -69,5 +70,35 @@ goog.exportSymbol('testClientHandler', {
 
     handler.dispose();
   },
+
+  'testClientChannelDisposed': function() {
+    const handler = new ClientHandler(
+        mockServerMessageChannel, goog.Promise.resolve(),
+        mockClientMessageChannel, EXTENSION_ID_A);
+    mockClientMessageChannel.dispose();
```
emaxx-google, line 78: nit: Maybe extract this into a separate function? This `.dispose()` together with the `disposeInternal` hack.

third_party/pcsc-lite/naclport/server_clients_management/src/client-handler-unittest.js
```diff
@@ -69,5 +70,35 @@ goog.exportSymbol('testClientHandler', {
...
+    // Hack: call the protected disposeInternal() method in order to trigger the
+    // disposal notifications; the MockMessageChannel.dispose() doesn't call
+    // them.
+    mockClientMessageChannel[goog.reflect.objectProperty(
+        'disposeInternal', mockClientMessageChannel)]();
+    assert(
```
emaxx-google, line 84: nit: `goog.require` `goog.asserts`

emaxx-google:
Thank you! Looks good modulo nits.

emaxx-google (changes requested):

emaxx-google (approved):

third_party/pcsc-lite/naclport/server_clients_management/src/client-handler-unittest.js
```diff
@@ -17,7 +17,9 @@
 
 goog.require('GoogleSmartCard.PcscLiteServerClientsManagement.ClientHandler');
 goog.require('goog.Promise');
+goog.require('goog.reflect');
 goog.require('goog.testing');
+goog.require('goog.testing.asserts');
```
emaxx-google, line 22: nit: Swap this with the next line, to keep directives sorted.

ViktoriiaKovalova: oops, fixed

PR #825, opened 2023-07-18
Migrate Checker to ES6 class

emaxx-google:
This is pure refactoring commit. Reimplement the "Checker" JavaScript class using the modern ES6 class instead of the deprecated Closure Compiler classes.


emaxx-google:
(The diff is horrible, but essentially it's changes of class/function definitions, plus a new constant, and one extra indentation for the rest of the file...)

ViktoriiaKovalova (approved):

PR #827, opened 2023-07-18
Add abstract PermissionsChecker interface

emaxx-google:
This is a pure refactoring commit. Introduce an abstract interface that allows to check whether a client (another extension) is allowed to send commands to Smart Card Connector. Change
PolicyOrPromptingPermissionsChecker to inherit from it. The abstraction will be used in follow-up changes to mock out the checker in tests.


ViktoriiaKovalova (approved):
LGTM

PR #829, opened 2023-07-18
Yet another magic sleep tweak for ASan+Debug tests

emaxx-google:
As we're still observing flaky failures from the ASan+Debug test configuration, and as upstream PC/SC still doesn't have a proper solution to avoid this kind of teardown problem, further increase the magic sleep workaround we're using in tests.

The new formula will give x2 multipler when the build is either ASan or Debug, and x4 multiplier when both hold.


ViktoriiaKovalova (approved):
LGTM

PR #830, opened 2023-07-18
Pass Promise instead of ReadinessTracker

emaxx-google:
This is a pure refactoring commit. Change the JS code to pass around a Promise instead of the whole ReadinessTracker object. This is sufficient in order to delay command processing until the server is ready, and is otherwise much simpler to mock out in tests.


ViktoriiaKovalova (approved):
LGTM

PR #831, opened 2023-07-18
Implement smoke test for ClientHandler

emaxx-google:
This unit test verifies that the ClientHandler object can be successfully created and disposed of.


ViktoriiaKovalova (approved):
LGTM

Issue #832, opened 2023-07-20
env/initialize.sh fails to init Chromedriver: 404


emaxx-google:
Result of executing `env/initialize.sh` on the current Chrome version (115.0.5790):

```
Installing chromedriver...                                                                                                                                                                                                         
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current                                                                                                                                                    
                                 Dload  Upload   Total   Spent    Left  Speed                                                                                                                                                      
100   200  100   200    0     0   1937      0 --:--:-- --:--:-- --:--:--  1960                                                                                                                                                     
Fetched chromedriver for 115.0.5790: <?xml version='1.0' encoding='UTF-8'?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Details>No such object: chromedriver/LATEST_RELEASE_115.0.5790</Detail
s></Error>                                                                                                                                                                                                                         
--2023-07-20 14:13:38--  https://chromedriver.storage.googleapis.com/%3C?xml                                                                                                                                                       
Resolving chromedriver.storage.googleapis.com (chromedriver.storage.googleapis.com)... 142.250.153.128, 2a00:1450:4013:c16::80                                                                                                     
Connecting to chromedriver.storage.googleapis.com (chromedriver.storage.googleapis.com)|142.250.153.128|:443... connected.                                                                                                         
HTTP request sent, awaiting response... 404 Not Found                                                                                                                                                                              
2023-07-20 14:13:38 ERROR 404: Not Found.                                                                                                                                                                                          
                                                                                                                                                                                                                                   
--2023-07-20 14:13:38--  http://version='1.0'/                                                                                                                                                                                     
Resolving version='1.0' (version='1.0')... failed: Name or service not known.                                                                                                                                                      
wget: unable to resolve host address ‘version='1.0'’                                                                                                                                                                               
--2023-07-20 14:13:38--  http://encoding='utf-8'/?%3E%3CError%3E%3CCode%3ENoSuchKey%3C/Code%3E%3CMessage%3EThe                                                                                                                     
Resolving encoding='utf-8' (encoding='utf-8')... failed: Name or service not known.                                                                                                                                                
wget: unable to resolve host address ‘encoding='utf-8'’                                                                                                                                                                            
--2023-07-20 14:13:38--  http://specified/                                                                                                                                                                                         
Resolving specified (specified)... failed: Name or service not known.                                                                                                                                                              
[...]

PR #833, opened 2023-07-20
Make Chromedriver error nonfatal in initialize.sh

emaxx-google:
If Chromedriver cannot be downloaded, env/initialize.sh should print the error message but then proceed to initializing other stuff. Chromedriver is only used for automated unit-testing, hence it shouldn't be considered critical for the env initialization.

Also improve the error messages in this part.

This is preparation for fixing #832.


ViktoriiaKovalova (approved):
LGTM

Issue #53, opened 2017-10-23
Get rid of warning logs in the Connector app when a reader is disconnected


emaxx-google:
Today, a disconnection of a device results in a bunch (~6) warning messages, emitted into the Connector's app JS console.

Those log messages are harmless and don't designate any actual error, but they may be confusing for the users / third-party developers.

PR #835, opened 2023-07-21
Downgrade warning logs for ReaderTracker removal

emaxx-google:
When the reader suddenly becomes unknown in the ReaderTracker, we shouldn't emit a warning log, because it's happening legitimately when the reader is unplugged in the middle of the SCardGetStatusChange() call.

Hence log these messages at the "info" level instead.

This is another small improvement for #53.


Fabian-Sommer (approved):
LGTM

PR #836, opened 2023-07-21
Build scripts use Emscripten by default

emaxx-google:
When no TOOLCHAIN environment variable is specified, assume "emscripten" by default.

Previously "pnacl" has been used by default, however it's becoming increasingly cumbersome for developers to use it.

Note that we continue shipping Smart Card Connector's pnacl version into production - this commit doesn't change that.


ViktoriiaKovalova (approved):
Thanks!

Issue #840, opened 2023-07-21
Smart Card Connector crashes on LockScreen with OMNIKEY 5422


emaxx-google:
When unplugging some readers, e.g., OMNIKEY 5422, Smart Card Connector occasionally crashes with:

```
Check "s_card_reader_state.cbAtr <= MAX_ATR_SIZE" failed. File "../src/google_smart_card_pcsc_lite_common/scard_structs_serialization.cc", line 40, function "GetSCardReaderStateAtr" 
Id @ background.js:122
```

More details in https://b.corp.google.com/issues/291892663.

PR #841, opened 2023-07-21
Fixup instead of crashing on incorrect ATR size

emaxx-google:
Under some circumstances PC/SC-Lite can return incorrect cbAtr, exceeding the maximum allowed MAX_ATR_SIZE. Fix this up, by treating it as zero, instead of crashing on an assertion.

The underlying root cause seems to be a race condition between the thread that handles SCardGetStatusChange() calls and the thread that manages global structures with reader information. When these read&write concurrently to the same structure, it can happen that the former reads the ATR length of the already-uninitialized reader. PC/SC-Lite's internal convention is that uninitialized reader structures have cbAtr equal to READER_NOT_INITIALIZED (which is MAX_ATR_SIZE + 1). Meanwhile we don't fix this race condition in this commit, we at least mitigate the consequences of the bug by substituting the invalid integer.

This serves as a temporary mitigation for #840; a proper fix will need to be done separately later.


Fabian-Sommer (approved):
LGTM

PR #844, opened 2023-07-24
Escape test function names to prevent minification

emaxx-google:
Use string literals for important function names in JavaScript unit tests ("setUp", "tearDown", "test...").

This makes sure that the compiler won't minify/rename them in Release builds, as renaming would make these functions unrecognized by the Closure Library test framework and effectively skipped silently.


Fabian-Sommer (approved):
LGTM

emaxx-google:
(Yeah, as you can see, I had to fix 2 tests as well - they were indeed not executed in Release previously. Horrible, but I'm unaware of a better solution with Closure Compiler.)

PR #845, opened 2023-07-25
Expose errorCode in JS API

### LINKED ISSUE #793: ResultOrErrorCode should export errorCode
The `errorCode` property of the JavaScript `ResultOrErrorCode` class is currently not marked as exported. So after the minification, it's not possible to access it directly.

The current workaround is to call `getBase()` (the `opt_onFailed` argument is called back with the error code).

emaxx-google:
JavaScript API library didn't expose the ResultOrErrorCode's error code value. The "errorCode" property wasn't marked as exported, hence it was unaccessible after minification.

This commit fixes this by exporting a new "getErrorCode" method instead. This fixes #793.


ViktoriiaKovalova (approved):
LGTM

PR #847, opened 2023-07-25
Use "ChromeOS" spelling in code and docs

emaxx-google:
This is the official name that's been preferred since a while. I haven't found official announcements, but a non-official one is https://9to5google.com/2022/04/29/google-prepares-chromeos-rebrand-to-drop-the-space/


Fabian-Sommer (approved):
LGTM

PR #848, opened 2023-07-25
Unit-test libusb_get_active_config_descriptor

emaxx-google:
Add a basic unit test for our libusb_get_active_config_descriptor() implementation (a.k.a LibusbJsProxy::LibusbGetActiveConfigDescriptor()).

The input parameters for the test are taken from an arbitrary real-world reader (SCM SCR 3310).

This commit is preparation for regression testing of #849.

### DIFF
--- a/third_party/libusb/webport/src/libusb_js_proxy_unittest.cc
+++ b/third_party/libusb/webport/src/libusb_js_proxy_unittest.cc
@@ -58,6 +58,7 @@ using testing::Invoke;
 using testing::InvokeWithoutArgs;
 using testing::Mock;
 using testing::MockFunction;
+using testing::SizeIs;
 using testing::WithArgs;
 
 namespace google_smart_card {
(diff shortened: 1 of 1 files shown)


third_party/libusb/webport/src/libusb_js_proxy_unittest.cc
```diff
@@ -463,6 +464,134 @@ TEST_P(LibusbJsProxyTest, DeviceRefUnref) {
...
+  EXPECT_EQ(interface_descriptor.bInterfaceSubClass, 0);
+  EXPECT_EQ(interface_descriptor.bInterfaceProtocol, 0);
+  ASSERT_TRUE(interface_descriptor.extra);
+  ASSERT_GT(interface_descriptor.extra_length, 0);
+  EXPECT_EQ(std::vector<uint8_t>(
+                interface_descriptor.extra,
+                interface_descriptor.extra + interface_descriptor.extra_length),
+            kInterfaceExtraData);
+  const auto* endpoint = interface_descriptor.endpoint;
+  ASSERT_TRUE(endpoint);
```
Fabian-Sommer, line 567: optional nit: You could also test the length of this array

Fabian-Sommer (approved):
LGTM

third_party/libusb/webport/src/libusb_js_proxy_unittest.cc
emaxx-google: It's tested above when checking the `bNumEndpoints`, but as it was hard to spot (even for myself, next day after I wrote this code :-) ) I've moved it close to this line.

Fabian-Sommer: No, I did see that. But I thought that buggy code could still produce an array that is larger (or smaller, though this would be caught implicitly in the assertions below). I don't expect that this would lead to incorrect behavior though, so I considered this optional.

emaxx-google: Got it. With C-style dynamic arrays it's almost impossible to check it properly though. At least our ASan tests would catch a too-small array, but a too-big one wouldn't be caught.

emaxx-google:
Thanks!

Issue #849, opened 2023-07-25
Never-ending initialization for composite readers like Rocketek


emaxx-google:
We received a bug report about broken support of a Rocketek reader:

```
[230718 13:51:40.76] [NaclModule.Wrapper.ReaderTracker] [INFO] The reader "Generic USB2.0-CRW" (port 2097152, device "usb:0bda/0169:libusb-1.0:1:6:0") was successfully initialized
<...>
[230718 13:51:41.38] [NaclModule.Wrapper.ReaderTracker] [INFO] Information about readers returned by PC/SC: "Generic USB2.0-CRW 00 00" (with inserted card)
<...>
[230718 13:51:43.35] [NaclModule] [INFO] 00022100 ../../src/src/ccid_usb.c:205:close_libusb_if_needed() libusb_exit
[230718 13:51:43.35] [NaclModule.Wrapper.ReaderTracker] [INFO] The reader "Generic USB2.0-CRW" (port 2097153) was removed
[230718 13:51:43.37] [NaclModule.Wrapper.ReaderTracker] [INFO] Information about readers returned by PC/SC: no readers
```

The investigation showed it's because of behavior change in ChromeOS 105, in which permission_broker started disallowing `openDevice` calls for already-claimed composite devices. In combination with our heuristic "if opening fails, simulate that the device reappears under a different bus number later", we essentially enter a never-ending loop: "initialize smart card USB interface => try opening the device again for another interface => react to the error by aborting opened connections => back to the initial step". 

More details at https://b.corp.google.com/issues/291810618.

emaxx-google:
It looks like we also need to patch the CCID Driver accordingly, because apparently it's doing null dereferences in some cases (which sometimes happen to work, but sometimes cause crashes).

PR #852, opened 2023-07-26
Simulate Dell Smart Card Reader Keyboard for tests

emaxx-google:
Add the second device into our small simulator used for testing Smart Card Connector - Dell Smart Card Reader Keyboard.

This allows us to verify the behavior of CCID and PC/SC-Lite for composite USB devices, in particular for those that have a non-smart-card interface#0 and the smart-interface#1. Follow-up PR will use this for regression-testing #849.


Fabian-Sommer (approved):
LGTM

PR #853, opened 2023-07-26
Don't look at upper bits when converting connection state

ViktoriiaKovalova:
Connection state contains the number of reader insertion/removal events in upper 16 bits.
This is not used by smartCardProviderPrivate API, so we want to only look at lower bits. Otherwise, the equalities we check for connection states won't hold if upper bits are non-zero.

### DIFF
--- a/smart_card_connector_app/src/chrome-api-provider.js
+++ b/smart_card_connector_app/src/chrome-api-provider.js
@@ -33,9 +33,8 @@ const PcscApi = GSC.PcscLiteClient.API;
 
 /** @type {!goog.log.Logger} */
 const logger = GSC.Logging.getScopedLogger('ConnectorApp.ChromeApiProvider');
-/** @type {number} Bitmask of the lower 16 bits. */
-const lowerBits = 0x0000FFFF;
-
+/** @type {number} Bitmask of the bits used to store the state flags. */
+const STATE_FLAGS_MASK = 0x0000FFFF;
 
 /**
  * Returns the result code for chrome.smartCardProviderPrivate API that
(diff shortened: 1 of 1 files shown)


smart_card_connector_app/src/chrome-api-provider.js
```diff
@@ -308,11 +308,14 @@ function logUnexpectedConnectionState(state) {
...
- * @param {number} state
+ * @param {number} pcscState
  * @returns {!chrome.smartCardProviderPrivate.ConnectionState}
  * @throws {Error} Throws error if unknown connection state is encountered.
  */
-function convertConnectionStateToEnum(state) {
+function convertConnectionStateToEnum(pcscState) {
+  // Upper 16 bits contains the number of events for the reader and don't have
+  // state flags.
+  state = pcscState & 0x0000FFFF;
```
emaxx-google, line 318: nit: This bitmask starts being repeated. Maybe introduce a constant for it (or for its negation)?

ViktoriiaKovalova: thanks!

smart_card_connector_app/src/chrome-api-provider.js
```diff
@@ -33,6 +33,9 @@ const PcscApi = GSC.PcscLiteClient.API;
 
 /** @type {!goog.log.Logger} */
 const logger = GSC.Logging.getScopedLogger('ConnectorApp.ChromeApiProvider');
+/** @type {number} Bitmask of the lower 16 bits. */
+const lowerBits = 0x0000FFFF;
```
emaxx-google, line 37: nit: Maybe some more informative name like `EVENT_FLAGS_MASK` or `EVENT_FLAGS_BITS`?

(And in caps, per style guide recommendation for "true constants".)

emaxx-google (approved):
Thanks for catching the issue.

PR #850, opened 2023-07-26
Cherry-pick upstream CCID fix for null-deref

emaxx-google:
Cherry-pick the "Check num_altsetting before dereffing altsetting" commit from the CCID Driver repository.

This is supposed to fix the occasional crashes we observe after applying the mitigation from #849.

PR #854, opened 2023-07-26
Add PC/SC test running with every simulated reader

emaxx-google:
Add a new test for the Smart Card Connector's PC/SC implementation that's parameterized by the simulated reader type. This lets us test some basic scenarios work correctly for a range of different devices.

This is a regression test for #849.


emaxx-google:
I verified that the new test fails on ASan bots if we don't cherry-pick #850.

Fabian-Sommer (approved):
LGTM

PR #855, opened 2023-07-27
Add logs for request processing to ChromeApiProvider

ViktoriiaKovalova:
Adds logs for requests from chrome.smartCardProviderPrivate API. Requests' arguments and results will be logged in Debug mode.

### DIFF
--- a/smart_card_connector_app/src/chrome-api-provider.js
+++ b/smart_card_connector_app/src/chrome-api-provider.js
@@ -548,7 +548,7 @@ GSC.ConnectorApp.ChromeApiProvider = class extends goog.Disposable {
   }
 
   /**
-   * @returns {!Array}
+   * @returns {!Promise<!Array>}
    * @private
    */
   async establishContext_() {
@@ -586,7 +586,7 @@ GSC.ConnectorApp.ChromeApiProvider = class extends goog.Disposable {
 
   /**
    * @param {number} sCardContext
-   * @returns {!Array}
+   * @returns {!Promise<!Array>}
    * @private
    */
   async releaseContext_(sCardContext) {
@@ -622,7 +622,7 @@ GSC.ConnectorApp.ChromeApiProvider = class extends goog.Disposable {
 
   /**
    * @param {number} sCardContext
-   * @returns {!Array}
+   * @returns {!Promise<!Array>}
    * @private
    */
   async listReaders_(sCardContext) {
@@ -664,7 +664,7 @@ GSC.ConnectorApp.ChromeApiProvider = class extends goog.Disposable {
    * @param {!chrome.smartCardProviderPrivate.Timeout} timeout
    * @param {!Array<!chrome.smartCardProviderPrivate.ReaderStateIn>}
    *     readerStates
-   * @returns {!Array}
+   * @returns {!Promise<!Array>}
    * @private
    */
   async getStatusChange_(sCardContext, timeout, readerStates) {
@@ -706,7 +706,7 @@ GSC.ConnectorApp.ChromeApiProvider = class extends goog.Disposable {
 
   /**
    * @param {number} sCardContext
-   * @returns {!Array}
+   * @returns {!Promise<!Array>}
    * @private
    */
   async cancel_(sCardContext) {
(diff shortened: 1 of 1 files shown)


smart_card_connector_app/src/chrome-api-provider.js
```diff
@@ -355,20 +358,48 @@ function convertConnectionStateToEnum(pcscState) {
...
+/**
+ * Generates a debug textual representation of the array of values.
+ *
+ * This function is safe to be used in Release builds, because all potentially
+ * privacy-sensitive data is stripped away from the resulting text.
+ * @param {!Array} arr
+ * @returns {string}
+ */
+function getArrayDebugRepresentation(arr) {
+  return goog.iter.join(goog.iter.map(arr, GSC.DebugDump.debugDump), ', ');
```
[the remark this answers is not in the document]
ViktoriiaKovalova: thanks!

smart_card_connector_app/src/chrome-api-provider.js
```diff
@@ -355,19 +358,56 @@ function convertConnectionStateToEnum(pcscState) {
...
-  constructor(chromeEvent, callback) {
+  constructor(chromeEvent, callback, reportResultFunction, nameForLogs) {
     super();
     this.chromeEvent_ = chromeEvent;
-    this.callback_ = callback;
+    this.callback_ = async (requestId, ...args) => {
+      goog.log.fine(
+          logger,
+          `Processing an event with id ${requestId}: ${nameForLogs}(${
+              getArrayDebugRepresentation(args)})`);
```
emaxx-google, line 400: nit: `getArrayDebugRepresentation(args)` is called twice within the same function, so can be cached in a variable instead.

Also consider doing the same for `getArrayDebugRepresentation(result)`, for readability purposes.

smart_card_connector_app/src/chrome-api-provider.js
```diff
@@ -511,20 +575,19 @@ GSC.ConnectorApp.ChromeApiProvider = class extends goog.Disposable {
       goog.log.log(
```
emaxx-google, line 575: These `try`-`catch` blocks seems like something that can be automatically done by some kind of a wrapper as well? Just a thought - I haven't thought it through, and no need to try it in the current PR.

emaxx-google (approved):
LGTM % nits

smart_card_connector_app/src/chrome-api-provider.js
ViktoriiaKovalova: i wanted to do smth like that, but i'm not sure how can that be done nicely.
The problem is that if the error is thrown, the wrapper will need to know what are the default values it needs to return.

smart_card_connector_app/src/chrome-api-provider.js
ViktoriiaKovalova: done

smart_card_connector_app/src/chrome-api-provider.js
```diff
@@ -483,10 +548,9 @@ GSC.ConnectorApp.ChromeApiProvider = class extends goog.Disposable {
   }
 
   /**
-   * @param {number} requestId
    * @private
    */
```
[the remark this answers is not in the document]
ViktoriiaKovalova: done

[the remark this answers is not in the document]
ViktoriiaKovalova: oh wait i'm supposed to return promise from this 😮 forgot about this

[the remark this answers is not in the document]
ViktoriiaKovalova: fixed

emaxx-google (approved):
LGTM

smart_card_connector_app/src/chrome-api-provider.js
emaxx-google: You can pass the default values as another parameter for that helper (`ChromeEventListener` or another helper method).

Ultimately, looking at the repetitive patterns across these methods, it seems that they can be replaced with a single generic method? It just needs to know the things that differ, something like: (1) Chrome API event name, (2) lambda that "massages" input arguments for sending them as PC/SC calls, (3) same for output arguments, (4) report*Result function, (5) default arguments for the report function.

PR #858, opened 2023-07-27
[CI] Don't try posting comments in PRs in forks

emaxx-google:
Normally our continuous integration scripts post a comment on a Pull Request once all jobs successfully complete. However, this doesn't work when the PR is in a fork, as the "Resource not accessible by integration" permission error occurs
(https://github.com/peter-evans/create-or-update-comment#action-inputs).

This commit makes this job being completely skipped. That way, we avoid spurious errors for now, until we maybe figure out the right way to do it.


ViktoriiaKovalova (approved):
Oh, nice 👍

Issue #816, opened 2023-06-30
Implement C++-to-JS tests in CI


emaxx-google:
We do already have such kinds of tests: https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/tree/main/example_cpp_smart_card_client_app/build/integration_tests

However, they are currently only implemented for NaCl, and run interactive Chrome. The task is to port the tests to WebAssembly/Emscripten as well, and allow running them in silent mode via Selenium+Chromedriver - so that they could be routinely invoked in Continuous Integration.

emaxx-google:
CC @ViktoriiaKovalova

PR #860, opened 2023-07-31
[JS-to-C++ test] Toolchain-agnostic build interface

emaxx-google:
Move the existing NaCl build rules into a common location, and refactor them to be an implementation of a toolchain-agnostic makefile rule.

This is preparation for introducing Emscripten build support for the same tests, and also for new JS-to-C++ integration tests (tracked by #816).

### DIFF
--- a/common/integration_testing/src/build_nacl.mk
+++ b/common/integration_testing/src/build_nacl.mk
@@ -18,15 +18,25 @@
 # Internal constant containing the list of additional C/C++ static libraries to
 # link against.
 #
-# Note: INTEGRATION_TESTING_LIB has to come before CPP_COMMON_LIB, since the
-# former uses symbols from the latter. And DEFAULT_NACL_LIBS is specified twice,
-# because there's a circular dependency between it and INTEGRATION_TESTING_LIB.
+# Notes:
+# * INTEGRATION_TESTING_LIB has to come before CPP_COMMON_LIB, since the
+#   former uses symbols from the latter.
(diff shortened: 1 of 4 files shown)


common/integration_testing/src/build_nacl.mk
```diff
@@ -0,0 +1,78 @@
...
+# This file contains the implementation of the ../include.mk interface that
+# builds the JS-to-C++ test runner using the Native Client toolchain.
+
+# Internal constant containing the list of additional C/C++ static libraries to
+# link against.
+#
+# Notes:
+# * INTEGRATION_TESTING_LIB has to come before CPP_COMMON_LIB, since the
+#   former uses symbols from the latter.
+# * DEFAULT_NACL_LIBS is specified twice, because there's a circular dependency
```
emaxx-google, line 24: Improved the comment to explain this hack, per offline discussion. Thanks for raising the question.

emaxx-google: After thinking more on this, I guess there's a somewhat more straightforward alternative. If I stop putting `integration_test_pp_module.cc` into a static library, and instead link directly against its object file, then its symbols will be included into the result without any tricks.

Also it looks like I need to do the same for the upcoming Emscripten anyway, for an unrelated reason though: the way how Emscripten exposes its guts to JS is via statically initialized global variables, and again the linker will just prune them away if I put it into a static library.

emaxx-google: Done

ViktoriiaKovalova (approved):
LGTM

PR #863, opened 2023-08-01
[JS-to-C++ test] Toolchain-agnostic controller

emaxx-google:
Refactor the IntegrationTestController JS class to be toolchain-agnostic instead of hardcoding NaCl-specific details.

This is preparation for implementing Emscripten integration tests, as tracked by #816.


ViktoriiaKovalova (approved):
thanks!

PR #864, opened 2023-08-01
[JS-to-C++ test] Add Emscripten barebone support

emaxx-google:
Add build scripts and "entry point" C++ code for the TOOLCHAIN=emscripten mode of integrationtests.

The tests are executed via Chromedriver that's pointed to a "localhost" URL that serves the JS and the C++ counterparts of the test.

This commit contributes to #816.


common/integration_testing/src/entry_point_emscripten.cc
```diff
@@ -0,0 +1,87 @@
...
+#include <utility>
+
+#include <emscripten/bind.h>
+#include <emscripten/val.h>
+
+#include <google_smart_card_common/global_context_impl_emscripten.h>
+#include <google_smart_card_common/optional.h>
+#include <google_smart_card_common/messaging/typed_message_router.h>
+#include <google_smart_card_common/value.h>
+#include <google_smart_card_common/value_debug_dumping.h>
```
ViktoriiaKovalova, line 34: is this one used?

emaxx-google: Good catch

ViktoriiaKovalova (approved):
Thanks!

PR #867, opened 2023-08-01
[JS-to-C++ test] Enable SharedArrayBuffer for Emscripten

emaxx-google:
Force-enable Chrome's SharedArrayBuffer feature when running integration tests in the TOOLCHAIN=emscripten mode, because otherwise multi-threading wouldn't work, which is required by the tests.

The underlying reason is that by default Chrome only allows the SharedArrayBuffer on pages that have CORS headers set up properly: see https://developer.chrome.com/blog/enabling-shared-array-buffer/ . Our tests are run a simple local web server which doesn't have them. As setting these up would be too tedious, the workaround here is to tweak Chrome when running the tests instead.

This commit contributes to #816 and is one of the last blockers before actually enabling integration tests in Emscripten mode and in CI.

### DIFF
--- a/common/integration_testing/src/build_emscripten.mk
+++ b/common/integration_testing/src/build_emscripten.mk
@@ -73,9 +73,15 @@ all: $(OUT_DIR_PATH)/index.html
 # --serve-via-web-server: Run the tests as "localhost:<random_port>/index.html"
 #   instead of just navigating to "file://.../index.html", because Chrome
 #   doesn't allow loading additional JavaScript code on file:// URLs.
+# --chrome-arg:
+#   --enable-features=SharedArrayBuffer: Force-enable SharedArrayBuffer that's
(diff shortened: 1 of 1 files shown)


ViktoriiaKovalova (approved):
LGTM

emaxx-google:
As discussed offline, the approach here isn't very future-proof, because it relies on Chrome to keep the `SharedArrayBuffer` feature that can be controlled this way. Apparently this feature flag was already deleted once in the past, and while there's no expiration date attached to it it can probably go away anytime in the future.

A better approach would be to actually implement COOP headers for our test servers, as described on https://developer.chrome.com/blog/enabling-shared-array-buffer/ . this is totally doable - just a bit inconvenient because Python's `http.server` doesn't have a simple way to inject custom headers.

For now let's land the PR as a quick workaround that unblocks us run the tests in the Emscripten mode. Later I'll take another look at the proper solution.

PR #872, opened 2023-08-01
[JS-to-C++ test] Increase test timeout

emaxx-google:
Closure Library's standard test timeout - 1 second - is insufficient for JS-to-C++ tests that spend multiple seconds alone on setup/teardown of the executable module.

Increase the limit to 60 seconds for now.

### DIFF
--- a/common/integration_testing/src/google_smart_card_integration_testing/integration-test-controller.js
+++ b/common/integration_testing/src/google_smart_card_integration_testing/integration-test-controller.js
@@ -26,6 +26,7 @@ goog.require('GoogleSmartCard.Requester');
 goog.require('goog.Promise');
 goog.require('goog.asserts');
 goog.require('goog.testing.PropertyReplacer');
+goog.require('goog.testing.TestCase');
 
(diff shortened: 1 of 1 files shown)


common/integration_testing/src/google_smart_card_integration_testing/integration-test-controller.js
```diff
@@ -68,6 +84,9 @@ GSC.IntegrationTestController = class {
       this.executableModuleRequester_.dispose();
       this.executableModule.dispose();
       this.propertyReplacer.reset();
+
+      goog.testing.TestCase.getActiveTestCase().promiseTimeout =
```
ViktoriiaKovalova, line 88: why do we need to reset it back? this value only affect the current test case, right?

emaxx-google: Done, you're right!

ViktoriiaKovalova (approved):
Thanks, LGTM

Issue #869, opened 2023-08-01
Add C++-to-JS test for Smart Card Connector API


emaxx-google:
With #816 implemented, we have a framework that allows writing JavaScript tests that exercise both JavaScript and C/C++ code.

We should employ this to write basic tests for verifying the PC/SC API exposed by the Connector.

PR #879, opened 2023-08-05
[JS-to-C++ test] Allow developer manual test debug

emaxx-google:
Let the developer pass the MANUAL_TEST_DEBUGGING=1 environment variable to make the test runner's web page visible and also to leave it opened after the tests finish.

We gate this behavior under an environment variable because Continuous Integration and routine test executions shouldn't mess up with the UI or hang on failures.


ViktoriiaKovalova (approved):
LGTM

PR #882, opened 2023-08-07
Uncommit PC/SC aclocal.m4 and configure

emaxx-google:
These files are autogenerated and shouldn't be tracked by the version control system. They're already gitignore'd, however as they were committed by an accident Git continued tracking them, causing spurious diffs to appear after every build.


ViktoriiaKovalova (approved):
LGTM

PR #883, opened 2023-08-07
Refactor LibusbJsProxy constant into a header

emaxx-google:
Create a new header for the string constant in libusb_js_proxy, so that it can be included without pulling the bunch of libusb-specific includes.

This refactoring is preparation for the USB simulation in JS-to-C++ tests, as tracked by #869.


ViktoriiaKovalova (approved):
LGTM

PR #884, opened 2023-08-07
[JS-to-C++ test] Test SCardReleaseContext

emaxx-google:
Verify the JavaScript API's SCardReleaseContext() behavior.

### DIFF
--- a/smart_card_connector_app/src/pcsc-api-jstocxxtest.js
+++ b/smart_card_connector_app/src/pcsc-api-jstocxxtest.js
@@ -38,6 +38,7 @@ goog.scope(function() {
 const GSC = GoogleSmartCard;
 const ClientHandler = GSC.PcscLiteServerClientsManagement.ClientHandler;
 const ReadinessTracker = GSC.PcscLiteServerClientsManagement.ReadinessTracker;
+const API = GSC.PcscLiteClient.API;
 
 /**
  * Stub that approves any client to make PC/SC calls.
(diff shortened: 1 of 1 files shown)


smart_card_connector_app/src/pcsc-api-jstocxxtest.js
```diff
@@ -131,6 +149,58 @@ goog.exportSymbol('testPcscApi', {
...
+          },
+          (errorCode) => {
+            called = true;
+            assertEquals(errorCode, API.SCARD_E_INVALID_HANDLE);
+          });
+      assert(called);
+    },
+
+    // Test `testSCardReleaseContext()` fails on a handle different from
+    // currently obtained one.
```
ViktoriiaKovalova, line 187: nit: `from currenly obtained one` sounds a bit weird

smart_card_connector_app/src/pcsc-api-jstocxxtest.js
```diff
@@ -131,6 +149,58 @@ goog.exportSymbol('testPcscApi', {
...
+      assert(called);
+    },
+
+    // Test `testSCardReleaseContext()` fails on a non-existing handle.
+    'testSCardReleaseContext_nonExisting': async function() {
+      const BAD_CONTEXT = 123;
+
+      const result = await api.SCardReleaseContext(BAD_CONTEXT);
+      let called = false;
+      result.get(
```
ViktoriiaKovalova, line 175: you can also use `getErrorCode` instead of `get` here 🤔

ViktoriiaKovalova (approved):
LGTM

smart_card_connector_app/src/pcsc-api-jstocxxtest.js
emaxx-google: Rephrased.

smart_card_connector_app/src/pcsc-api-jstocxxtest.js
emaxx-google: Done - good point, but let's actually check both? Admittedly `get()` is clumsy, but as it's part of the API we have to test it...

Also added `getErrorCode()` assertions into the `SCardEstablishContext` tests above.

ViktoriiaKovalova: ack

I don't see the point in adding both though, I suggestedi using `getErrorCode` just to avoid writing all that code :)

emaxx-google: Ack - I think I'll keep both here, since both methods are parts of the Smart Card Connector API, and if we don't check them here they'll remain completely untested. But in tests of other methods - sure, let's use the shorter form.

Issue #885, opened 2023-08-07
Include headers by absolute path


emaxx-google:
Currently we have a full mix of different ways of `#include`'ing C++ headers: (a) current-directory, (b) relative-path, (c) absolute-path from the repo root, (d) implicit search path includes.

The idea is to get rid of `(a)` and `(b)`, and leave `(c)` with `(d)`. The `(c)` approach is widely used in projects, like Chromium, and proved to be robust. The `(d)` approach is for rare cases when it's really unavoidable (mainly in third-party code that expects other libraries to be installed into the search path).

emaxx-google:
As a bonus, we can get rid of some annoyingly long file paths, like `common/cpp/src/google_smart_card_common/logging/logging.h`. Something like `common/cpp/src/public/logging/logging.h` would be better.

PR #886, opened 2023-08-07
[headers] Add root dir to include search path

emaxx-google:
Add the repository's root directory into the C/C++ search path.

This is preparation for the refactoring to use absolute include paths consistently, as tracked by #885.


ViktoriiaKovalova (approved):
LGTM

PR #888, opened 2023-08-08
Helper for rerouting requests in tests

emaxx-google:
Add a primitive for "rerouting" C++-to-JS requests in tests.

This will be used in follow-ups to refactor how application_unittests work today, i.e., how they stub out Libusb's requests with the TestingSmartCardSimulation class. We need this to make the latter easier to reuse for JS-to-C++ tests, as tracked by #869.


ViktoriiaKovalova (approved):
lgtm

PR #891, opened 2023-08-08
[JS-to-C++ test] Create TestingSmartCardSimulation

emaxx-google:
Instantiate TestingSmartCardSimulation in JS-to-C++ tests that use SmartCardConnectorApplicationTestHelper.

To note, it's still disfunctional because we need to wire it up with Libusb requests. This will be addressed in follow-ups.


ViktoriiaKovalova (approved):
LGTM

PR #893, opened 2023-08-08
[headers] Move files to common/cpp/src/public/

emaxx-google:
Move all files from //common/cpp/src/google_smart_card_common/
to //common/cpp/src/public/.

It's a pure refactoring commit that gets rid of unnecessarily long file
names. It's possible now that we include these headers by their
absolute path, as tracked by #885.


emaxx-google:
Mostly mechanical change.

PR #895, opened 2023-08-08
[headers] Include by abs path from libusb/webport/

emaxx-google:
Change `#include "foo.h"` (which implicitly relied on manipulated include search paths) to `#include "third_party/libusb/webport/src/"`.


ViktoriiaKovalova (approved):
LGTM

PR #897, opened 2023-08-08
[headers] Include by abs path from pcsc-lite/

emaxx-google:
Change `#include <google_smart_card_pcsc_...>` and same-directory `#include "foo.h"` includes to `#include
"third_party/pcsc-lite/naclport/..."`.


ViktoriiaKovalova (approved):
LGTM

PR #898, opened 2023-08-08
[headers] Include by abs path from smart_card_connector_app/

emaxx-google:
Change same-directory `#include "foo.h"` includes to `#include "smart_card_connector_app/src/..."`.


ViktoriiaKovalova (approved):
LGTM

PR #901, opened 2023-08-09
[JS-to-C++ test] Test SCardListReaders

emaxx-google:
Verify the JavaScript API's SCardListReaders() behavior.


emaxx-google:
Thank you for all the reviews! This PR reaches the milestone "we can write tests for JS code with simulated PC/SC smart card readers".

Issue #902, opened 2023-08-14
Fix smart_card_connector_app js_unittests


emaxx-google:
1. These tests aren't built or executed by default.
2. When called manually, they fail: `pcsc-api-jstocxxtest.js:23:0: ERROR - [JSC_MISSING_MODULE_OR_PROVIDE] Required namespace "GoogleSmartCard.IntegrationTestController" never defined.`.

PR #903, opened 2023-08-14
[JS unit tests] Fix connector_app test building

emaxx-google:
Exclude JS-to-C++ sources from compilation of js_unittests, because these integration-like tests need special setup and are executed in a separate target.

This commit contributes to #902.

### DIFF
--- a/smart_card_connector_app/build/js_unittests/Makefile
+++ b/smart_card_connector_app/build/js_unittests/Makefile
@@ -28,12 +28,20 @@ include $(ROOT_PATH)/third_party/pcsc-lite/naclport/js_client/include.mk
 include $(ROOT_PATH)/third_party/pcsc-lite/naclport/common/include.mk
 include $(ROOT_PATH)/third_party/pcsc-lite/naclport/server_clients_management/include.mk
 
+# The list of source paths that Closure Compiler should compile for these unit
+# tests.
+# At high level, it's:
+# * Code-under-test and its tests under smart_card_connector_app/.
+# * All libraries it uses (including transitive dependencies).
+# * Exclusion pattern for C++-to-JS tests (they're run in a separate target with
+#   special setup).
 JS_COMPILER_INPUT_PATHS := \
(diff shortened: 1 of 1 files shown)


smart_card_connector_app/build/js_unittests/Makefile
```diff
@@ -28,12 +28,20 @@ include $(ROOT_PATH)/third_party/pcsc-lite/naclport/js_client/include.mk
...
 JS_COMPILER_INPUT_PATHS := \
   $(JS_COMMON_JS_COMPILER_INPUT_DIR_PATHS) \
   $(PCSC_LITE_COMMON_JS_COMPILER_INPUT_DIR_PATHS) \
   $(PCSC_LITE_JS_CLIENT_COMPILER_INPUT_DIR_PATHS) \
   $(PCSC_LITE_SERVER_CLIENTS_MANAGEMENT_JS_COMPILER_INPUT_DIR_PATHS) \
-  ../../../third_party/libusb/webport/src \
-  ../../src \
+  $(ROOT_PATH)/third_party/libusb/webport/src \
+  $(ROOT_PATH)/smart_card_connector_app/src \
+  !$(ROOT_PATH)/smart_card_connector_app/src/!**jstocxxtest.js \
```
emaxx-google, line 45: I was hoping initially I could instead do a more generic solution like:

```
  ... various paths ... \
  !$(ROOT_PATH)/**test.js \
  $(ROOT_PATH)/smart_card_connector_app/src/**unittest.js \
```

However, unfortunately, Closure Compiler always gives higher priority to `!` negation patterns. I.e., a file excluded by some `!` pattern cannot be re-included by some other pattern. Which means we have to continue this tedious inclusion–exclusion-principle-like approach...

ViktoriiaKovalova (approved):
Thanks, Maksim!

smart_card_connector_app/build/js_unittests/Makefile
ViktoriiaKovalova: what is the purpose of the second `!` here?

PR #905, opened 2023-08-16
Continue clean for other toolchains if it breaks for pnacl

ViktoriiaKovalova:
Currently, in case the virtual environment wasn't initialised, e.g., `initialize.sh` was called with `-t emscripten` flag, clean-all breaks after an attempt to clean for toolchain='pnacl'.

This CL changes it to continue with other clean tasks.


emaxx-google (approved):
thanks

PR #907, opened 2023-08-16
Build scripts use Emscripten by default

ViktoriiaKovalova:
Change default toolchain in ./Makefile as well. This is a follow-up for #836.


emaxx-google (approved):
thank you for spotting this inconsistency

PR #910, opened 2023-08-16
[wasm] Optimize Emscripten code via closure

emaxx-google:
Enable Closure Compiler based optimizations of the code generated by Emscripten.

Effect on the file sizes for Smart Card Connector:
* executable_module.js: 149K => 70K
* executable_module.worker.js: 2.9K => 3.0K
* resulting .zip archive: 447K => 435K

Potentially this should also improve performance, but we don't have measurements to compare against.


Fabian-Sommer (approved):
LGTM

PR #911, opened 2023-08-16
[wasm] Prune unneeded ENVIRONMENT support code

emaxx-google:
Tell Emscripten to only generate code for supporting "web"/"worker" environment by default, and only "node" when building C++ unit test runner.

This commit optimizes away unneeded support code from the resulting artifacts. The exact improvement is around a couple kilobytes, still there's no reason to avoid doing this.

For reference, the default value of the "ENVIRONMENT" Emscripten setting is "web,webview,worker,node".

### DIFF
--- a/common/cpp_unit_test_runner/src/build_emscripten.mk
+++ b/common/cpp_unit_test_runner/src/build_emscripten.mk
@@ -29,6 +29,8 @@ TEST_ADDITIONAL_CXXFLAGS := \
 # Documented in ../include.mk.
 #
 # Explanation:
+# ENVIRONMENT: Generate support code for running in the Node.js environment.
+#   (This overrides the default setting in executable_building_emscripten.mk.)
 # EXIT_RUNTIME: Quit the program when the main() exits - otherwise the execution
 #   will hang.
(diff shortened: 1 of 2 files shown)


common/cpp_unit_test_runner/src/build_emscripten.mk
```diff
@@ -42,11 +42,14 @@ TEST_ADDITIONAL_CXXFLAGS := \
...
 #   separate thread. Otherwise, multi-threaded tests will hang as background
 #   tests aren't created until the main thread yields.
+# ENVIRONMENT: Generate support code for running in the Node.js environment.
+#   (This overrides the default setting in executable_building_emscripten.mk.)
 TEST_ADDITIONAL_LDFLAGS := \
 	-s EXIT_RUNTIME \
 	-s EXPORT_NAME=Module \
 	-s MODULARIZE=0 \
 	-s PROXY_TO_PTHREAD \
+  -s ENVIRONMENT=node \
```
Fabian-Sommer, line 52: nit: This does not align with the lines above whitespace-wise

Fabian-Sommer (approved):
LGTM % nit

PR #912, opened 2023-08-16
[wasm] Prune code for unneeded Module attributes

emaxx-google:
Tell Emscripten to only generate code for the attributes of the Module class that we're actually using at runtime. This allows to optimize away code for unneeded ones.

In practice, this reduces the resulting files by a couple kilobytes.


Fabian-Sommer (approved):
LGTM

PR #915, opened 2023-08-16
[wasm] Drop support for ancient Chrome

emaxx-google:
Skip generating Emscripten support code for Chrome versions <96. There are multiple technical reasons why we won't ship WebAssembly builds to such ancient client versions, so let's benefit from it by letting Emscripten prune workarounds for them.

Note: the binary size reduction is almost negligible in practice.


Fabian-Sommer (approved):
LGTM

PR #916, opened 2023-08-16
[usb] Move chrome-login-state-hook.js

emaxx-google:
Move chrome-login-state-hook.js and its unit tests from //third_party/libusb/webport/src/ to //smart_card_connector_app/.

The hook isn't a core part of the libusb functionality, but rather a tweak needed for Smart Card Connector, hence the new location makes more sense logically.


Fabian-Sommer (approved):
LGTM

PR #917, opened 2023-08-16
[wasm] Suppress warning logs about worker pool

emaxx-google:
Disable Emscripten logs every time a new thread is started and there's no free worker.

These warnings are useful when the web porting works just started, because they hint the developer about potential deadlocks caused by deviations of Emscripten semantics from POSIX (namely, that a thread spawned from the main thread won't actually start until the main thread returns to the event loop). Now that we did the basic migration work for Emscripten, these warnings are mostly red herring.


Fabian-Sommer (approved):
LGTM

PR #919, opened 2023-08-17
Create a mock of chrome.smartCardProviderPrivate API

ViktoriiaKovalova:


smart_card_connector_app/src/mock-chrome-api.js
```diff
@@ -0,0 +1,205 @@
...
+ */
+GoogleSmartCard.ConnectorApp.MockChromeApi = class extends goog.Disposable {
+  /**
+   * @param {!goog.testing.PropertyReplacer} propertyReplacer
+   */
+  constructor(propertyReplacer) {
+    super();
+    /** @type {Object.<string, Function>} */
+    this.listeners_ = {};
+    this.listenersMockControl_ = new goog.testing.MockControl();
```
emaxx-google, line 127: Usually there's one `MockControl` for the whole test. It makes sure there's one central place that verifies the mock expectations and nothing is forgotten.

So can we keep creating `mockControl` in the test file and pass it here as a parameter?

smart_card_connector_app/src/mock-chrome-api.js
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  /**
+   * Dispatches the specified chrome.smartCardProviderPrivate API event with
+   * `args`.
+   * @param {string} eventName
+   * @param  {...*} args
+   */
+  async dispatchEvent(eventName, ...args) {
+    goog.asserts.assertFunction(this.listeners_[eventName]);
```
emaxx-google, line 201: This is stripped out in Release builds. A more reliable alternative is either `GSC.Logging.check` or `assert...()` from `goog.testing.asserts`.

smart_card_connector_app/src/mock-chrome-api.js
ViktoriiaKovalova: done

smart_card_connector_app/src/mock-chrome-api.js
ViktoriiaKovalova: thanks

smart_card_connector_app/src/mock-chrome-api.js
```diff
@@ -0,0 +1,190 @@
...
+   * Dispatches the specified chrome.smartCardProviderPrivate API event with
+   * `args`.
+   * @param {string} eventName
+   * @param  {...*} args
+   */
+  async dispatchEvent(eventName, ...args) {
+    assertEquals(
+        'No listener to dispatch event to', 'function',
+        typeof this.listeners_[eventName]);
+    await this.listeners_[eventName](...args);
```
emaxx-google, line 187: One thought about this `await`... Are we sure it's waiting for the correct thing? IIUC it waits for the function that's added via `addListener`, but is it guaranteed to be async function or to resolve after completing all its work?

I think your real intention here is to wait until the corresponding `report...()` method is called on the API, but that's probably better to be waited explicitly.

smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -71,20 +73,15 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
...
-    chromeApiProvider = new ChromeApiProvider(
-        testController.executableModule.getMessageChannel(),
-        pcscReadinessTracker.promise);
+    mockChromeApi =
+        new MockChromeApi(mockControl, testController.propertyReplacer);
   },
 
   'tearDown': async function() {
     try {
+      chromeApiProvider.dispose();
```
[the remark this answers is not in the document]

emaxx-google (approved):

smart_card_connector_app/src/mock-chrome-api.js
emaxx-google: you're welcome

smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -71,20 +73,14 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
...
-        pcscReadinessTracker.promise);
+    mockChromeApi =
+        new MockChromeApi(mockControl, testController.propertyReplacer);
   },
 
   'tearDown': async function() {
     try {
       await testController.disposeAsync();
       pcscReadinessTracker.dispose();
-      assertTrue(chromeApiProvider.isDisposed());
```
emaxx-google, line 87: BTW this assert would be still useful now?

ViktoriiaKovalova: added back in https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/pull/946

PR #920, opened 2023-08-17
Add EstablishContext test to chrome-api-jstocxxtest.js

ViktoriiaKovalova:


smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -93,6 +94,19 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
   'testSmoke': async function() {
     launchPcscServer(/*initialDevices=*/[]);
     await pcscReadinessTracker.promise;
+  },
+
+  'testEstablishContext': async function() {
+    launchPcscServer(/*initialDevices=*/[]);
+    await pcscReadinessTracker.promise;
```
emaxx-google, line 101: Is this necessary? Ideally we test without this, to verify that even early API events are handled correctly (by being delayed until the PC/SC server initialization completes).

smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -93,6 +94,19 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
...
+    chrome
+        .smartCardProviderPrivate['reportEstablishContextResult'](
+            /*requestId=*/ 1,
```
emaxx-google, line 104: Optional, but maybe some more distinguished number like 42 or 123? So that the tests don't accidentally pass if some kind of counter is used accidentally.

emaxx-google (commented):
Nice!

smart_card_connector_app/src/chrome-api-jstocxxtest.js
ViktoriiaKovalova: good catch!

smart_card_connector_app/src/chrome-api-jstocxxtest.js

smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -97,6 +98,24 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
         testController.executableModule.getMessageChannel(),
         pcscReadinessTracker.promise);
     await pcscReadinessTracker.promise;
+  },
+
+  'testEstablishContext': async function() {
+    chrome
+        .smartCardProviderPrivate['reportEstablishContextResult'](
+            /*requestId=*/ 123,
+            /*scardContext=*/ goog.testing.mockmatchers.isNumber, 'SUCCESS')
```
emaxx-google, line 107: That's good enough for the initial PR that introduces this test, but eventually it's probably better to double-check the received `scardContext` against `SCardIsValidContext()`.

emaxx-google (approved):

PR #921, opened 2023-08-17
[pcsc] Fix crash on shutdown with active client

emaxx-google:
Fix the crash during the PC/SC server shutdown in case there's at least one connected JS client.

The underlying problem was that we didn't tear down the C++ client handling subsystem, and hence kept open pipes beyond the shutdown of all other systems. It was a test-only issue since the extensions never shut down in production (unless killed by the browser).


ViktoriiaKovalova (approved):
LGTM

PR #923, opened 2023-08-17
[JS-to-C++ test] Rename example app's tests

emaxx-google:
Rename example_cpp_smart_card_client_app's /integration_tests/ to /js_to_cxx_tests/. We concluded on this new name while adding such tests to smart_card_connector_app recently.

This is a pure refactoring commit. The related effort is tracked by #869.


emaxx-google:
> can you also update `example_cpp_smart_card_client_app/build/Makefile`?

good catch, done

ViktoriiaKovalova (approved):

PR #924, opened 2023-08-17
[headers] Move files to third_party/pcsc-lite/naclport/common/src/public

emaxx-google:
Move all files from
//third_party/pcsc-lite/naclport/common/src/google_smart_card_pcsc_lite_common/ to //third_party/pcsc-lite/naclport/common/src/public/.

It's a pure refactoring commit that gets rid of unnecessarily long file names. It's possible now that we include these headers by their absolute path, as tracked by #885.


Fabian-Sommer (approved):
LGTM

Issue #926, opened 2023-08-17
Warn about crashes in the UI


emaxx-google:
If the Smart Card Connector crashes repeatedly, we stop auto-reloading ourselves but enter a dysfunctional state.

The task is to display a warning in the UI when this happens, so that the user can get the initial idea what went wrong.

PR #927, opened 2023-08-17
[ui] Refactor OS warning UI into separate file

emaxx-google:
Create a new window-status-displaying.js and move the logic that displays the "ChromeOS only" warning into it.

This is a pure refactoring change. It's a preparation for implementing crash warning UI as tracked by #926.


ViktoriiaKovalova (approved):
LGTM

PR #928, opened 2023-08-17
[chrome-api-jstocxxtest] Add tests for releasing context

ViktoriiaKovalova:


ViktoriiaKovalova:
There are also some more fancy release context test cases in pcsc-js2cxxtest. Do you think we should duplicate them here as well?

emaxx-google (approved):

emaxx-google:
> There are also some more fancy release context test cases in pcsc-js2cxxtest. Do you think we should duplicate them here as well?

Covering just all main outcomes (`SUCCESS`/`INVALID_HANDLE` in this case) sounds sufficient to me.

PR #931, opened 2023-08-17
[ui] Show warning in UI if application crashed

### LINKED ISSUE #926: Warn about crashes in the UI

emaxx-google:
If the executable module crashes (and cannot recover by self-reloading), notify the user by displaying a warning in the Smart Card Connector's main window UI.

This fixes #926.


Fabian-Sommer (approved):
LGTM

PR #932, opened 2023-08-18
[policy] Fix crash if policy loaded after disposal

emaxx-google:
Fix exception thrown in AdminPolicyService in case the managed storage gets loaded after the object is already disposed of.

This doesn't happen normally, but in case the executable module crashed due to some issue this AdminPolicyService exception would be a red herring that confuses the developer.


annklr (approved):
Thank you for fixing this!

PR #933, opened 2023-08-18
[wasm] Improve handling of crashes

emaxx-google:
If the Emscripten module crashed, we should avoid making new operations with it that can trigger new exceptions (which would be confusing and useless). Hence we do two improvements in this commit:

1. Don't try calling the C++ object destructor after the crash.
2. Discard messages that the module tries to send to the JS side after the module's disposal.


Fabian-Sommer (approved):
LGTM

PR #934, opened 2023-08-18
[js style] Fix extra semicolons

emaxx-google:
Fix a few occurrences of extra semicolons in JS code. This is found via ESLint.

### DIFF
--- a/common/js/src/logging/log-buffer.js
+++ b/common/js/src/logging/log-buffer.js
@@ -73,7 +73,7 @@ class State {
     result += suffix;
     return result;
   }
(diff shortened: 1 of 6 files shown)


ViktoriiaKovalova (approved):
thanks!
I'm wondering why it doesn't include `mock-chrome-api.js` 🤔

emaxx-google:
> thanks! I'm wondering why it doesn't include `mock-chrome-api.js` 🤔

looks like semicolons are fine in that file currently. unless there are more bad surprises in that file!

Issue #937, opened 2023-08-21
Set up ESLint for JavaScript code


emaxx-google:

emaxx-google:
For now I'm running ESLint manually, and also it produces tons of false positives because it doesn't understand Closure-style imports. I'll figure out how to fix those and fully automate it later.

PR #939, opened 2023-08-21
[js style] Fix no-prototype-builtins

emaxx-google:
Instead of calling hasOwnProperty() on the target object, use the Object.prototype.hasOwnProperty.call() notation instead. This makes sure there's no shadowing possible.

This is found via ESLint. It's part of the effort tracked by #937.


Fabian-Sommer (approved):
LGTM

PR #940, opened 2023-08-21
[js] Fix forgotten arg inside LogBufferForwarder

emaxx-google:
We forgot to pass POSTPONING_BUFFER_CAPACITY to the underlying CircularBuffer. Luckily, this has been a no-op bug since the constant happened to equal the buffer's default size.

This is found via ESLint. It's part of the effort tracked by #937.


ViktoriiaKovalova (approved):
LGTM, nice finding :)

PR #944, opened 2023-08-21
[JS-to-C++ test] Tear down BridgeBackend

emaxx-google:
Make sure to dispose of BridgeBackend in the bridge-integrationtest's teardown.

This was spotted while looking through the ESLint reports per #937.


emaxx-google:
> Since testController is already being disposed, shouldn't that trigger bridgeBackend disposal anyway? SmartCardClientApp.CertificateProviderBridge.Backend has the `executableModuleDisposedListener_` listener for that.

You're right. So two better alternatives would be: either (a) assert it's automatically been disposed of, or (b) dispose of it earlier. I've changed the PR to do "b", since logically the bridge depends on the executable module, and hence the test shouldn't keep the bridge alive past the module's lifetime. LMKWYT.

ViktoriiaKovalova (approved):

PR #945, opened 2023-08-21
[wip ]create chrome api provider if policy enables

ViktoriiaKovalova:


smart_card_connector_app/src/background.js
```diff
@@ -17,6 +17,7 @@
 
 goog.provide('GoogleSmartCard.ConnectorApp.BackgroundMain');
 
+goog.require('GoogleSmartCard.ConnectorApp.ChromeApiProvider');
```
emaxx-google, line 20: nit: Sort the requires.

smart_card_connector_app/src/managed_storage_schema.json
```diff
@@ -11,6 +11,11 @@
         "maxLength": 32
       }
     },
+    "handle_requests_from_chrome_smart_card_api": {
```
emaxx-google, line 14: Maybe `chrome_smart_card_api_enabled` or `expose_chrome_smart_card_api`?

smart_card_connector_app/src/background.js
```diff
@@ -165,6 +166,28 @@ if (GSC.ExecutableModule.TOOLCHAIN ===
   GSC.BackgroundPageUnloadPreventing.enable();
 }
 
+if (chrome.smartCardProviderPrivate !== undefined) {
+  const HANDLE_CHROME_REQUESTS = 'handle_requests_from_chrome_smart_card_api';
+  chrome.storage.managed.get(HANDLE_CHROME_REQUESTS, (policies) => {
```
emaxx-google, line 171: Does Chrome expect the event listeners to be subscribed immediately on the app's startup?

I'm asking because this `chrome.storage.managed.get()` call introduces some asynchronous delay. I haven't checked how Chrome side is implemented - whether it waits for the app's installation, startup or the event listener subscription.

smart_card_connector_app/src/background.js
```diff
@@ -165,6 +166,28 @@ if (GSC.ExecutableModule.TOOLCHAIN ===
...
+    if (policies[HANDLE_CHROME_REQUESTS] === true) {
+      new GSC.ConnectorApp.ChromeApiProvider(
```
emaxx-google, line 173: nit: Add a comment explaining why this variable isn't stored anywhere.

smart_card_connector_app/src/background.js
emaxx-google, line 171: The policy can also change at runtime (it's especially possible during the first-time login). So it's better to set up an observer that'd react to the changes of the policy value.

ViktoriiaKovalova: if no one is subscribed to the events yet, it will return related error to all requests coming from the browser.

emaxx-google: Ack. Might be fine, as long as the application that consumes the Web API is aware of this behavior.

smart_card_connector_app/src/background.js
ViktoriiaKovalova: done

smart_card_connector_app/src/managed_storage_schema.json
ViktoriiaKovalova: done

smart_card_connector_app/src/background.js
ViktoriiaKovalova: fixed

smart_card_connector_app/src/background.js
```diff
@@ -165,6 +166,42 @@ if (GSC.ExecutableModule.TOOLCHAIN ===
...
+          `Requests from Web Smart Card API will not be handled.`);
+}
+
+/**
+ * Updates the Chrome API provider availability based on extension policy.
+ */
+function updateChromeApiProviderAvailability() {
+  const EXPOSE_CHROME_REQUESTS = 'expose_chrome_smart_card_api';
+  chrome.storage.managed.get(EXPOSE_CHROME_REQUESTS, (policies) => {
+    if (policies[EXPOSE_CHROME_REQUESTS] === true) {
```
emaxx-google, line 188: nit: I think it'd be a bit easier to read if we reduce the nesting level, like:

```
const policy_enabled = policies[EXPOSE_CHROME_REQUESTS] === true;
if (policy_enabled && !chromeApiProvider) {
  chromeApiProvider = ...;
} else if (!policy_enabled && chromeApiProvider) {
  ...
  chromeApiProvider.dispose();
  ...
}
```

emaxx-google (approved):
LGTM % suggestion

smart_card_connector_app/src/background.js
ViktoriiaKovalova: thanks, changed to this version

PR #946, opened 2023-08-21
[chrome-api-jstocxxtest] Check that ChromeApiProvider is disposed

ViktoriiaKovalova:


smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -91,6 +89,7 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
     } finally {
       // Check all mock expectations are satisfied.
       mockControl.$verifyAll();
+      assertTrue(chromeApiProvider.isDisposed());
```
emaxx-google, line 92: Let's do it inside the `try {` block. `$verifyAll()` belongs there as well I think.

My reasoning is: the `try-finally` block is intended to make sure no globals leak between test cases. Hence let's try to make `finally` as bulletproof as possible, with everything else sitting in the `try` block.

emaxx-google, line 92: Also assign `chromeApiProvider` to `null`, so that the next test starts from a clear global state.

ViktoriiaKovalova: thanks, fixed

smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -46,7 +46,8 @@ let pcscReadinessTracker;
 let mockChromeApi;
 /** @type {!goog.testing.MockControl|undefined} */
 let mockControl;
-
+/** @type {ChromeApiProvider?} */
+let chromeApiProvider;
```
emaxx-google, line 50: nit: Restore the blank line before the next function.

emaxx-google (approved):
LGTM % nit

smart_card_connector_app/src/chrome-api-jstocxxtest.js
ViktoriiaKovalova: thanks :)

PR #949, opened 2023-08-21
[pcsc] Fix flaky shutdown

emaxx-google:
Fix flaky memory leaks and drop magic sleep in the PC/SC shutdown logic, in favor of deterministic signaling and waiting.

### DIFF
--- a/third_party/pcsc-lite/src/src/hotplug_libusb.c
+++ b/third_party/pcsc-lite/src/src/hotplug_libusb.c
@@ -447,30 +447,6 @@ static void HPRescanUsbBus(void)
 
 	/* free the libusb allocated list & devices */
 	libusb_free_device_list(devs, 1);
-
-	if (AraKiriHotPlug)
-	{
-		int retval;
-
-		for (i=0; i<PCSCLITE_MAX_READERS_CONTEXTS; i++)
-		{
-			if (readerTracker[i].fullName != NULL)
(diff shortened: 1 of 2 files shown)


ViktoriiaKovalova (approved):

third_party/pcsc-lite/src/src/hotplug_libusb.c
```diff
@@ -519,24 +495,23 @@ static void * HPEstablishUSBNotifications(int pipefd[2])
 		do_polling = true;
 	}
 
+	char dummy;
```
ViktoriiaKovalova, line 498: this can be moved to `else` clause?

emaxx-google: We also use it below, in `write()`... But maybe I should try refactoring this into smaller functions.

ViktoriiaKovalova: ack

emaxx-google:
Didn't manage to repro it, but I have one theory... Let's see if commit#2 passes the bots.

emaxx-google:
After stress-testing this for hours, I managed to catch some crashes, but they seem to be orthogonal. So, given the huge benefit of this PR for the developer experience, and given that it should only affect behavior in tests, let's land it for now.

PR #950, opened 2023-08-21
[JS-to-C++ test] Exclude Connector unit tests

emaxx-google:
Exclude JS unit tests ("...-unittest.js") from the js_to_cxx test target. These unit tests are executed by their dedicated targets anyway, so it was duplicate execution.

### DIFF
--- a/smart_card_connector_app/build/js_to_cxx_tests/Makefile
+++ b/smart_card_connector_app/build/js_to_cxx_tests/Makefile
@@ -64,10 +64,10 @@ LIBS := \
   $(CPP_COMMON_LIB) \
   $(DEFAULT_NACL_LIBS) \
 
+# Include *-jstocxxtest.js and non-test files needed for them.
 JS_SOURCES_PATHS := \
   $(SOURCE_DIR) \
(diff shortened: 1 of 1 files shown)


smart_card_connector_app/build/js_to_cxx_tests/Makefile
```diff
@@ -66,6 +66,7 @@ LIBS := \
 
 JS_SOURCES_PATHS := \
   $(SOURCE_DIR) \
+  !$(SOURCE_DIR)/**-unittest.js \
   $(SOURCE_DIR)/**-jstocxxtest.js \
```
ViktoriiaKovalova, line 70: isn't this one already added by $(SOURCE_DIR) in line 68?

ViktoriiaKovalova (approved):

smart_card_connector_app/build/js_to_cxx_tests/Makefile
emaxx-google: You're right, it's not necessary. I guess I originally added it to make it more explicit, but let's remove it.

Issue #951, opened 2023-08-21
ASan test flakiness


emaxx-google:
We observe flakiness in CI runs, especially on asan-debug bots. Example error:

```
=================================================================
==47338==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 15904 byte(s) in 1 object(s) allocated from:
    #0 0x56871beb in __interceptor_realloc (/home/runner/work/chromeos_smart_card_connector/chromeos_smart_card_connector/smart_card_connector_app/build/executable_module/cpp_unittests/out/cpp_unit_test_runner/cpp_unit_test_runner+0x305beb) (BuildId: 6c3e6b3b3c3df359c5f049ce23192c4e560be175)
    #1 0x56b523af in HPReadBundleValues /home/runner/work/chromeos_smart_card_connector/chromeos_smart_card_connector/third_party/pcsc-lite/naclport/server/build/../../../../../third_party/pcsc-lite/src/src/hotplug_libusb.c:253:18
    #2 0x56b50f0f in HPSearchHotPluggables /home/runner/work/chromeos_smart_card_connector/chromeos_smart_card_connector/third_party/pcsc-lite/naclport/server/build/../../../../../third_party/pcsc-lite/src/src/hotplug_libusb.c:568:6
    #3 0x56b12c54 in google_smart_card::PcscLiteServerWebPortService::InitializeAndRunDaemonThread() /home/runner/work/chromeos_smart_card_connector/chromeos_smart_card_connector/third_party/pcsc-lite/naclport/server/build/../src/public/pcsc_lite_server_web_port_service.cc:259:17
    #4 0x568b1eaa in google_smart_card::Application::InitializeServicesOnBackgroundThread() /home/runner/work/chromeos_smart_card_connector/chromeos_smart_card_connector/smart_card_connector_app/build/executable_module/cpp_unittests/../../../src/application.cc:69:39
    #5 0x568b5e57 in void std::__invoke_impl<void, void (google_smart_card::Application::*)(), google_smart_card::Application*>(std::__invoke_memfun_deref, void (google_smart_card::Application::*&&)(), google_smart_card::Application*&&) /usr/bin/../lib/gcc-cross/i686-linux-gnu/12/../../../../i686-linux-gnu/include/c++/12/bits/invoke.h:74:14
[...]

PR #952, opened 2023-08-22
[JS-to-C++ test] Shut down request receiver

emaxx-google:
Make sure to shut down the IntegrationTestService's request receiver before dropping a reference on it, so that use-after-free isn't possible in case there's a background thread still using it.

This is part of the test flakiness fixes tracked by #951.


Fabian-Sommer (approved):
LGTM

PR #953, opened 2023-08-22
[JS-to-C++ test] Print full JS logs on failure

emaxx-google:
Collect and dump all logs from the JS console in case the test fails, for easier investigation for the developer.


Fabian-Sommer (approved):
LGTM

PR #954, opened 2023-08-22
[JS unit tests] Not pull libusb tests in Connector

emaxx-google:
Exclude libusb's JS unit tests when compiling Connector's JS unit tests, so that the former aren't executed twice in both targets.

As a drive-by, fix accidentally added excessive exclamation mark in one of the file patterns.


ViktoriiaKovalova (approved):
Thank you!

emaxx-google:
> Thank you!

yeah unfortunately this tedious include-exclude story continues :unamused:

PR #958, opened 2023-08-22
[js style] Fix excessively long lines

emaxx-google:
Reformat JS code to fit into 80-column limit whenever possible (e.g., we leave long URLs as-is).

This is found via ESLint. This refactoring is part of the effort tracked by #937.


Fabian-Sommer (approved):
LGTM

PR #961, opened 2023-08-23
[js style] Fix invalid typeof comparison

emaxx-google:
Compare values using just "!==" as opposed to "typeof ... !==", since the latter uses a string and hence would never pass.

This seems like a real bug, although it's probably unlikely to hit any user since it requires ChromeOS <=50.

This was found by ESLint. This is part of the cleanup tracked by #937.


Fabian-Sommer (approved):
LGTM

PR #964, opened 2023-08-24
Update Emscripten to 3.1.45

emaxx-google:
We don't require any particular improvements, however it's good to keep up with upstream to learn about possible issues earlier.


Fabian-Sommer (approved):
LGTM

PR #969, opened 2023-08-28
[js style] Execute ESLint in CI runs

emaxx-google:
Lint JavaScript code via ESLint in each Continuous Integration run.

This is part of the linting effort tracked by #937.

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -141,24 +141,24 @@ jobs:
     timeout-minutes: 10
 
     steps:
-    # Check out with additionally pulling the "main" branch, so that the script
-    # can build diff against it.
+    # Check out with additionally pulling the parent commit. (In case of Pull
+    # Requests, it's pointing to the base branch's head.)
     - name: Check out sources
       uses: actions/checkout@v3
       with:
-        fetch-depth: 0
-    - name: Create extra checkout branches
-      run: >
-        git branch main origin/main
+        fetch-depth: 2
 
     - name: Install dependencies
       uses: ./.github/actions/build-dependencies-installation
       with:
         toolchain: eslint
 
+    # Run the linter on suitable files. For PR's, only look at the diff against
+    # the parent commit (which, as Github executes the action on the merge
+    # commit, is the base branch). Otherwise, all files are linted.
     - name: Run ESLint
       run: >
-        scripts/eslint.sh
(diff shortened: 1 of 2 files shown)


.github/workflows/makefile.yml
```diff
@@ -135,6 +135,31 @@ jobs:
...
+    steps:
+    # Check out with additionally pulling the "main" branch, so that the script
+    # can build diff against it.
+    - name: Check out sources
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: Create extra checkout branches
+      run: >
+        git branch main origin/main
```
ViktoriiaKovalova, line 152: should we get a diff with the base branch here instead?

ViktoriiaKovalova: we can just put ```git branch main HEAD^1``` here instead

emaxx-google: Good question. Did I get it correctly you're thinking about the case of CL chains: whether an bug in CL1 should trigger linter errors in all CL1/CL2/etc., or should the linter only complain in CL1?

Currently the PR implements the former, because the linter in each CL looks at the diff from `main`.

If we are to implement the latter, I think we should look at things that Github provides natively, like `github.base_ref`/`GITHUB_BASE_REF`. The `HEAD^1` trick is interesting, but I'm not sure it works correctly in all cases, e.g. when the CL contains multiple merge commits.

ViktoriiaKovalova (approved):
LGTM; thanks!

.github/workflows/makefile.yml
```diff
@@ -135,6 +135,31 @@ jobs:
...
+      uses: ./.github/actions/build-dependencies-installation
+      with:
+        toolchain: eslint
+
+    # Run the linter on suitable files. For PR's, only look at the diff against
+    # the parent commit (which, as Github executes the action on the merge
+    # commit, is the base branch). Otherwise, all files are linted.
+    - name: Run ESLint
+      run: >
+        scripts/eslint.py --base="${{ github.event_name == 'pull_request' && 'HEAD^1' || 'none' }}"
```
emaxx-google, line 161: Hmm, it just occurred to me that this means that when a PR modifies the linter config, we won't essentially check anything in it. If there are existing JS files that violate newly enabled checks, the PR author won't learn about them (until the PR is merged and the post-merge CI waterfall runs the linter on all files).

Either we need to detect when a PR affects the linting stuff or maybe simply run the linter on all files?..

PR #970, opened 2023-08-28
Refactor logic for finding sources-to-be-linted

emaxx-google:
Create a separate helper script for finding manually maintained source files. Currently it's only used for running clang-format from format-code.sh, but we're planning to add a similar script for JS linting.


Fabian-Sommer (approved):
LGTM

PR #972, opened 2023-08-28
[headers] Drop excessive C/C++ search paths

emaxx-google:
Now that most of C/C++ headers are included using their absolute path (starting from the repository's root directory), we can drop most of header search paths.

This is a pure refactoring commit that contributes to #885.


Fabian-Sommer (approved):
LGTM

PR #973, opened 2023-08-28
[JS-to-C++ test] Test SCardListReaders on 2 readers

emaxx-google:
Extend the JavaScript API SCardListReaders()'s tests to cover the case of two devices attached.


ViktoriiaKovalova (approved):
LGTM

PR #978, opened 2023-08-29
[scripts] Rewrite find-files-for-linting in Python

emaxx-google:
Reimplement this script in Python, to allow for easier customization for CI or local execution.

This contributes to the effort tracked by https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/issues/937.


emaxx-google:
(After discussion in other PRs, I realized we want a couple more command-line parameters, which would be very tedious with pure Shell. So let it be Python.)

PR #979, opened 2023-08-29
[scripts] Pass base gitref to find-files-for-linting

emaxx-google:
Let the caller specify whether all files need to be linted, or a diff to a particular git ref needs to be considered.


Fabian-Sommer (approved):
LGTM

PR #980, opened 2023-08-29
[scripts] Ignore third-party code from linting

emaxx-google:
find-files-for-linting.py ignores third-party code that we don't maintain manually within this repository.


Fabian-Sommer (approved):
LGTM

PR #984, opened 2023-08-29
[ci] Skip JS targets in sanitizer/coverage builds

emaxx-google:
JavaScript code building and testing is only useful in relevant toolchains (currently emscripten and pnacl). It should be skipped in analysis builds that are currently focused on C/C++ code (asan_testing and coverage).

This would improve CI speed a bit by shaving off unnecessary operations in those jobs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,16 +71,6 @@ third_party/libusb/webport/build/tests: third_party/libusb/webport/build
 
 TOOLCHAIN ?= emscripten
 
-# Applications are only built in non-sanitizer builds.
-ifneq ($(TOOLCHAIN),asan_testing)
-
-APPLICATION_TARGETS += \
-	example_cpp_smart_card_client_app/build \
-	example_js_smart_card_client_app/build \
-	smart_card_connector_app/build \
-
-endif
-
 # Use our checkout of Googletest for tests in all toolchains except NaCl (whose
 # SDK provides its own version).
 ifneq ($(TOOLCHAIN),pnacl)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -99,15 +94,20 @@ third_party/libusb/webport/build/tests: third_party/googletest/webport/build
 
 endif
 
-# Enable JS-to-C++ tests in relevant configurations.
+# Enable JS targets in relevant configurations.
 ifneq (,$(findstring $(TOOLCHAIN),pnacl emscripten))
```
ViktoriiaKovalova, line 98: Should this part (line 74)
```
# Applications are only built in non-sanitizer builds.
ifneq ($(TOOLCHAIN),asan_testing)
```
also be put here under (pnacl || emscripten) buildflag? I assume that applications don't need to be build for `coverage`.

ViktoriiaKovalova (approved):

Makefile
emaxx-google: Good idea! Unfortunately the effect will be partially neglected by the fact that some test targets depend on the application targets (which btw doesn't look really necessary). But some overhead will go away already now. Thanks.

emaxx-google: Actually I forgot that there was an actual reason why we build the applications: the coverage script looks at the application binaries as well when calculating the resulting percentage. I've added a comment about that now.

But we can still do this small optimization for JS-only applications.

ViktoriiaKovalova: Thanks!

PR #985, opened 2023-08-29
[ci] Skip python3_venv and chromedriver in analysis

emaxx-google:
Don't initialize env's python3_venv and chromedriver when building for analysis (asan_testing and coverage).

This should reduce the analysis build job times a bit.


ViktoriiaKovalova (approved):
Nice, thanks!

PR #986, opened 2023-08-29
[C++ test] Test SCardReleaseContext edge cases

emaxx-google:
Add unit tests for the cases when SCardReleaseContext() is called for wrong values.


Fabian-Sommer (approved):
LGTM

PR #990, opened 2023-08-29
[JS-to-C++] Test malformed requests

emaxx-google:
Test various edge cases of receiving incorrect PC/SC requests. The only thing to verify is that we don't crash.


ViktoriiaKovalova (approved):

smart_card_connector_app/src/pcsc-api-jstocxxtest.js
```diff
@@ -141,7 +146,9 @@ function disposeFakeClient(fakeClient) {
  * @return {!Promise}
  */
 async function sleep(timeoutMillis) {
-  return new Promise(resolve => setTimeout(resolve, timeoutMillis));
```
emaxx-google, line 144: BTW this change is necessary because the about-to-be-enabled linter would complain on this...

```
  148:33  error  Return values from promise executor functions cannot be read  no-promise-executor-return
```

Luckily we're ready to land #969 now, which would prevent these from sneaking into the tree.

PR #992, opened 2023-08-30
[headers] Move cpp_demo to public

emaxx-google:
Rename "google_smart_card_pcsc_lite_cpp_demo" to "public", now that includes from this folder use absolute paths and there's no possible conflict anymore.


Fabian-Sommer (approved):
LGTM

PR #995, opened 2023-08-30
[js style] Enable more ESLint suggestion checks

emaxx-google:
Turn on a bunch of ESLint checks that are documented as "suggestions" ("these rules suggest alternate ways of doing things"). We skip those that trigger too many false positives or require fixing the code (better done in separate PRs).

This is part of the linting effort tracked by #937.


Fabian-Sommer (approved):
LGTM

PR #997, opened 2023-08-30
[js style] PRs touching linting are checked on all

emaxx-google:
When a PR is published that touches ESLint stuff (config, scripts, etc.),
the PR actions should run the linter on all files in the repository. That
way, the developer gets immediate feedback about the change, as
opposed to learning from errors after the PR gets merged into main.


ViktoriiaKovalova (approved):
Thanks!

PR #998, opened 2023-08-31
[js style] Enable ESLint camelcase

emaxx-google:
The check should prevent non-camel-case variable names which are against Google Style Guide. The exceptions are optional arguments ("opt_..."), constant names and property names.


Fabian-Sommer (approved):
LGTM

PR #999, opened 2023-08-31
[js style] Enable ESLint guard-for-in

emaxx-google:
And fix the existing violation of this rule (in tests).


Fabian-Sommer (approved):
LGTM

PR #1001, opened 2023-09-04
[chrome-api-jstocxxtest] Add more GetStatusChange tests

ViktoriiaKovalova:
Add tests to check that events with various `eventState`s and `eventCount`s are handled correctly.
Add test with optional timeout parameter passed.

### DIFF
--- a/smart_card_connector_app/src/chrome-api-jstocxxtest.js
+++ b/smart_card_connector_app/src/chrome-api-jstocxxtest.js
@@ -369,5 +369,106 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
     setSimulatedDevices([{'id': 123, 'type': 'gemaltoPcTwinReader'}]);
     await verifyResult;
   },
+
+  // Test GetStatusChange returns the reader and card information.
+  'testGetStatusChange_withCardInitially': async function() {
+    const establishContextResult = EMPTY_CONTEXT_RESULT;
+    expectReportEstablishContext(
+        /*requestId=*/ 123, 'SUCCESS', establishContextResult);
+    expectReportGetStatusChange(
+        /*requestId=*/ 124, [{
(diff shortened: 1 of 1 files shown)


ViktoriiaKovalova:
Looks like CI is breaking because of the github bug 
https://github.com/actions/checkout/issues/1448

smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -369,5 +369,106 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
...
+  'testGetStatusChange_withTimeout': async function() {
+    const establishContextResult = EMPTY_CONTEXT_RESULT;
+    expectReportEstablishContext(
+        /*requestId=*/ 123, 'SUCCESS', establishContextResult);
+    expectReportGetStatusChange(
+        /*requestId=*/ 124, [], 'TIMEOUT');
+    mockControl.$replayAll();
+
+    launchPcscServer(/*initialDevices=*/[
+      {'id': 123, 'type': 'gemaltoPcTwinReader', 'cardType': 'cosmoId70'}
```
Fabian-Sommer, line 456: I think considering how often the strings 'gemaltoPcTwinReader' and 'cosmoId70' are used, they deserve to be constants (throughout this file, maybe in another PR) . Especially considering that at least the Gemalto Reader already has some associated constants, and I suspect the exact string might be important.

Fabian-Sommer (approved):
LGTM

smart_card_connector_app/src/chrome-api-jstocxxtest.js
ViktoriiaKovalova: good idea! I'll make a follow-up PR :)

PR #1003, opened 2023-09-04
Create constants for simulation values used by js-to-cxx tests

ViktoriiaKovalova:
Create constants for TestingSmartCardSimulation values used by js-to-cxx tests. This is a pure refactoring commit.

### DIFF
--- a/smart_card_connector_app/src/testing-smart-card-simulation-constants.js
+++ b/smart_card_connector_app/src/testing-smart-card-simulation-constants.js
@@ -27,6 +27,22 @@ const GSC = GoogleSmartCard;
 GSC.TestingLibusbSmartCardSimulationConstants.CPP_HELPER_NAME =
     'SmartCardConnectorApplicationTestHelper';
 
+// Names of the device types defined by TestingSmartCardSimulation.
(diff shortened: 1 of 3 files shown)


smart_card_connector_app/src/testing-smart-card-simulation-constants.js
```diff
@@ -27,6 +27,18 @@ const GSC = GoogleSmartCard;
 GSC.TestingLibusbSmartCardSimulationConstants.CPP_HELPER_NAME =
     'SmartCardConnectorApplicationTestHelper';
 
+// Names of the device types defined by TestingSmartCardSimulation.
```
Fabian-Sommer, line 30: nit: I suspect that these strings need to match the ones in testing_smart_card_simulation.cc - if so, it would be good to add that to the comment.

Fabian-Sommer (approved):
LGTM % nit

smart_card_connector_app/src/testing-smart-card-simulation-constants.js
ViktoriiaKovalova: done, thanks!

PR #1005, opened 2023-09-04
Fix clean-all.sh for pnacl

ViktoriiaKovalova:
Fix the following error `[-f: Command not found]` that was happening during clean, because of missing whitespaces.


Fabian-Sommer (approved):
LGTM

PR #1006, opened 2023-09-05
[Web API] Log on ChromeApiProvider creation/disposal

ViktoriiaKovalova:
Add log on ChromeApiProvider successful creation. Log if provider needs to be created/destroyed due to policy.

### DIFF
--- a/smart_card_connector_app/src/background.js
+++ b/smart_card_connector_app/src/background.js
@@ -173,7 +173,7 @@ if (chrome.smartCardProviderPrivate !== undefined) {
   updateChromeApiProviderAvailability();
   chrome.storage.onChanged.addListener(updateChromeApiProviderAvailability);
 } else {
-  goog.log.fine(
+  goog.log.info(
       logger,
       `chrome.smartCardProviderPrivate is not available. ` +
           `Requests from Web Smart Card API will not be handled.`);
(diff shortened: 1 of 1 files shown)


Fabian-Sommer (commented):
I think we should defer this one until emaxx is back - my feeling is that we don't need to log at info level if the object was created successfully, but I am not the person who will need to look at the logs later on :)

smart_card_connector_app/src/chrome-api-provider.js
```diff
@@ -445,6 +445,7 @@ GSC.ConnectorApp.ChromeApiProvider = class extends goog.Disposable {
     this.subscribeToChromeApiEvents_();
 
     this.serverRequester_.addOnDisposeCallback(() => this.dispose());
+    goog.log.info(logger, 'ChromeApiProvider was intialized successfully.');
```
emaxx-google, line 448: nit: Typo in "initialized"

emaxx-google, line 448: This log is emitted in addition to `Creating ChromeApiProvider <...>`. Seems excessive; maybe leave only one of the two?

emaxx-google:
> I think we should defer this one until emaxx is back - my feeling is that we don't need to log at info level if the object was created successfully, but I am not the person who will need to look at the logs later on :)

I think an extra log line is tolerable in this case, given the nontrivial conditions that determine whether Web Smart Card API is available in Chrome.

smart_card_connector_app/src/chrome-api-provider.js
ViktoriiaKovalova: done

emaxx-google (approved):

PR #1015, opened 2023-09-07
Skip setup-python for toolchains that don't need python

### LINKED ISSUE #1014: CI breaks for non-emscripten/pnacl toolchains
CI breaks >50% of the time for asan-testing/collect-coverage/lint-js with the following error:
```
Post job cleanup.
Post job cleanup.
Error: Cache folder path is retrieved for pip but doesn't exist on disk: /home/runner/.cache/pip
```
Example:
https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/actions/runs/6111626282/job/16587278119?pr=1013

This seems to happen only for those jobs, because they don't use python after installation.

Related issue in setup-python repo:
https://github.com/actions/setup-python/issues/436

I tried checking if it still happens with setup-python@v4 and it still broke ([example link](https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/actions/runs/6111626282/job/16587278119?pr=1013)).

Probably we can disable setup-python for those toolchains as a workaround.

ViktoriiaKovalova:
This fixes #1014.

### DIFF
--- a/.github/actions/build-dependencies-installation/action.yml
+++ b/.github/actions/build-dependencies-installation/action.yml
@@ -25,6 +25,7 @@ runs:
       pkg-config
 
   - name: Set up Python 3
+    if: ${{ inputs.toolchain == 'emscripten' || inputs.toolchain == 'pnacl' }}
     uses: actions/setup-python@v3.1.2
     with:
       python-version: 3.9


.github/actions/build-dependencies-installation/action.yml
```diff
@@ -25,6 +25,7 @@ runs:
       pkg-config
 
   - name: Set up Python 3
+    if: ${{ inputs.toolchain == 'emscripten' || inputs.toolchain == 'pnacl' }}
```
emaxx-google, line 28: ESLint also needs Python

emaxx-google: Also `coverage` as well, because we build the whole `smart_card_connector_app` in all modes except ASan:

https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/blob/0ba4103b135d3a622a2550917507d1852ba63a3c/Makefile#L82-L92

and the latter triggers:

https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/blob/0ba4103b135d3a622a2550917507d1852ba63a3c/smart_card_connector_app/build/Makefile#L115

(I might fix this some day, because conceptually `coverage` only needs the C++ code to be built, but right now we just build everything.)

So I guess right now we can only really exclude `asan_testing`?..

ViktoriiaKovalova: i'll close it, since we decided to go with https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/pull/1021 instead.

PR #1016, opened 2023-09-07
[chrome-api-jstocxxtest] Test Status

ViktoriiaKovalova:


smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -620,7 +639,7 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
         /*preferredProtocols=*/ {'t1': true},
         /*expectedProtocol=*/ chrome.smartCardProviderPrivate.Protocol.T1);
     expectReportPlainResult(/*requestId=*/ 113, 'SUCCESS');
-    mockChromeApi.dispatchEvent(
+    await mockChromeApi.dispatchEvent(
```
emaxx-google, line 642: Stuff from another PR. A rebase would help.

smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -640,6 +659,139 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
...
+                chrome.smartCardProviderPrivate.ConnectionState.NEGOTIABLE,
+                chrome.smartCardProviderPrivate.Protocol.T1,
+                SimulationConstants.COSMO_ID_70_ATR, 'SUCCESS'
+              ],
+              saveArgs.allArgs);
+          continue;
+        }
+
+        if (resultCode != 'READER_UNAVAILABLE') {
+          goog.testing.asserts.raiseException(
```
emaxx-google, line 731: nit: `fail('...')` is more concise.

smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -640,6 +659,139 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
...
+          .dispatchEvent('onStatusRequested', /*requestId=*/ 111, readerHandle)
+          .$waitAndVerify();
+    },
+
+    // Test that Status returns READER_UNAVAILABLE after reader disappears.
+    'testStatus_readerUnavailable': async function() {
+      setSimulatedDevices(/*devices=*/[]);
+
+      // Status should eventually return a READER_UNAVAILABLE error.
+      for (let requestId = 111; ; requestId += 1) {
```
emaxx-google, line 708: Having such a busy loop is unfortunate. It's unclear how many times it spins in a test run (if any), whether it's crucial for test coverage that it spins at least once. Also if it happens to work sufficiently long, `requestId` will start coinciding with those used in the setUp/tearDown.

Can we have a more deterministic test here? E.g., will we lose on test coverage if we launch the PC/SC server without any devices, do a single `onStatusRequested` and verify it returns `READER_UNAVAILABLE`?
Alternatively, if you really want to test the transition, you can use `SCardGetStatusChange()` or analogs to wait for it explicitly.

smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -640,6 +659,139 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
...
+        const saveArgs = new goog.testing.mockmatchers.SaveArgument();
+        expectReportStatusResult(
+            requestId, saveArgs, saveArgs, saveArgs, saveArgs, saveArgs);
```
emaxx-google, line 711: Does anything guarantee the left-to-right ordering of the argument processing? I'm worried `saveArgs` might turn out to be populated in unexpected order.

smart_card_connector_app/src/chrome-api-jstocxxtest.js
ViktoriiaKovalova: > It's unclear how many times it spins in a test run

around 50
> E.g., will we lose on test coverage if we launch the PC/SC server without any devices, do a single onStatusRequested and verify it returns READER_UNAVAILABLE

it won't work like this, since you first need to make a `Connect` call.

>Alternatively, if you really want to test the transition, you can use SCardGetStatusChange() or analogs to wait for it explicitly.

I'll check if smth like this works. Initially I just made a similar test to the one in `application_unittest.cc` and there it's claimed that `(we have to do a polling loop since there's no simple way to observe this asynchronous event).`

emaxx-google: OK indeed I forgot about that C++ test that I wrote :D Alright, in that case I'd say - it's fine to keep the polling in the JS as well, but please fix the request id overlap.

ViktoriiaKovalova: fixed the request id overlap

smart_card_connector_app/src/chrome-api-jstocxxtest.js
ViktoriiaKovalova: thanks!

smart_card_connector_app/src/chrome-api-jstocxxtest.js
ViktoriiaKovalova: fixed

smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -236,9 +237,27 @@ function expectReportPlainResult(requestId, resultCode) {
...
+      resultCode)
+      .$once()
+      .$replay();
+}
+
+/**
+ * Sets expectation that reportStatusResult will be called with given
+ * parameters.
+ * @param {number} requestId
+ * @param {string|ArgMatcher} readerName
```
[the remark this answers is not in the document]

smart_card_connector_app/src/chrome-api-jstocxxtest.js
ViktoriiaKovalova: changed this to saving with `$does`.

emaxx-google (approved):
thank you

PR #1017, opened 2023-09-07
Get chromedriver version from CfT API endpoint

### LINKED ISSUE #1009: Downloading chromedriver breaks for certain versions
Example: https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/actions/runs/6097245953/job/16544492257

Currently we are downloading chromedriver with the same version as Chrome currently installed on device. However, sometimes it breaks (see example).

It's probably better to use some versions provided by these [endpoints](https://github.com/GoogleChromeLabs/chrome-for-testing#json-api-endpoints) for chromedriver.
Maybe this one: `https://googlechromelabs.github.io/chrome-for-testing/LATEST_RELEASE_STABLE`

ViktoriiaKovalova:
Get chromedriver version from Chrome for Testing API endpoint, instead of using the same version as currently installed Chrome. This fixes #1009.

### DIFF
--- a/env/initialize.sh
+++ b/env/initialize.sh
@@ -213,11 +213,15 @@ initialize_chromedriver() {
   fi
   log_message "Installing chromedriver..."
   rm -rf ./chromedriver
-  # Determine the currently installed version of Chrome. Leave only the numbers,
-  # e.g., 72.0.3626.0.
-  local chrome_version=$(google-chrome --version | cut -f 3 -d ' ')
-  # Download Chromedriver. As of now, the maintainers guarantee that there's a
-  # matching Chromedriver artifact for every publicly released Chrome version.
+  # Determine the version of the latest available stable chromedriver.
+  # The link is taken from here https://github.com/GoogleChromeLabs/chrome-for-testing#other-api-endpoints.
+  local chromedriver_version_url="https://googlechromelabs.github.io/chrome-for-testing/LATEST_RELEASE_STABLE"
+  local chrome_version=$(wget -qO- "${chromedriver_version_url}")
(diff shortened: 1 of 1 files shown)


env/initialize.sh
```diff
@@ -213,11 +213,15 @@ initialize_chromedriver() {
...
   log_message "Installing chromedriver..."
   rm -rf ./chromedriver
-  # Determine the currently installed version of Chrome. Leave only the numbers,
-  # e.g., 72.0.3626.0.
-  local chrome_version=$(google-chrome --version | cut -f 3 -d ' ')
-  # Download Chromedriver. As of now, the maintainers guarantee that there's a
-  # matching Chromedriver artifact for every publicly released Chrome version.
+  # Determine the version of the latest available stable chromedriver.
+  # The link is taken from here https://github.com/GoogleChromeLabs/chrome-for-testing#other-api-endpoints.
+  local chromedriver_version_url="https://googlechromelabs.github.io/chrome-for-testing/LATEST_RELEASE_STABLE"
```
emaxx-google, line 218: If the version fetched here mismatches the installed version of Chrome, I think Chromedriver will fail.

emaxx-google: So if we're to follow this approach, we'd need to install this `Chrome for Testing`. Might be doable - I haven't checked the details.

I've also commented on the Issue with some thoughts about an alternative solution.

ViktoriiaKovalova:
closing, since https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/pull/1023 was merged

emaxx-google:
From long-term perspective, this still might be a good way to go, as chatted offline. But we can reopen it anytime.

PR #1018, opened 2023-09-07
Update set-output to GITHUB_OUTPUT

ViktoriiaKovalova:
`set-output` is [deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).


emaxx-google (approved):
Thanks!

PR #1020, opened 2023-09-07
[JS-to-C++ test] Test SCardConnect success

emaxx-google:
Test basic successful scenarios for the SCardConnect PC/SC function.

### DIFF
--- a/smart_card_connector_app/src/pcsc-api-jstocxxtest.js
+++ b/smart_card_connector_app/src/pcsc-api-jstocxxtest.js
@@ -832,6 +832,64 @@ goog.exportSymbol('testPcscApi', {
       assertTrue(called);
       assertEquals(statusResult.getErrorCode(), API.SCARD_E_CANCELLED);
     },
+
+    // Test `SCardConnect()` succeeds for dwShareMode `SCARD_SHARE_DIRECT` even
+    // when there's no card inserted.
+    'testSCardConnect_successDirectNoCard': async function() {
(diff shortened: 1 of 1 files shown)


smart_card_connector_app/src/pcsc-api-jstocxxtest.js
```diff
@@ -832,6 +832,64 @@ goog.exportSymbol('testPcscApi', {
...
+      const result = await client.api.SCardConnect(
+          context, SimulationConstants.GEMALTO_PC_TWIN_READER_PCSC_NAME0,
+          API.SCARD_SHARE_DIRECT, /*preferred_protocols=*/ 0);
+
+      let called = false;
+      result.get(
+          (sCardHandle, activeProtocol) => {
+            called = true;
+            assert(Number.isInteger(sCardHandle));
+            assertNotEquals(sCardHandle, 0);
```
ViktoriiaKovalova, line 854: why can't it be 0?

ViktoriiaKovalova (approved):
LGTM

smart_card_connector_app/src/pcsc-api-jstocxxtest.js
emaxx-google: ok, let's drop it, there's no strict API guarantee indeed and the implementation can emit it with 2^{-31} probability (although the value "zero" also is returned by C API on errors, and some applications may in theory assume "zero => error")

PR #1021, opened 2023-09-07
Disable pip cache in CI

### LINKED ISSUE #1014: CI breaks for non-emscripten/pnacl toolchains

ViktoriiaKovalova:
This fixes #1014.

### DIFF
--- a/.github/actions/build-dependencies-installation/action.yml
+++ b/.github/actions/build-dependencies-installation/action.yml
@@ -28,7 +28,6 @@ runs:
     uses: actions/setup-python@v3.1.2
     with:
       python-version: 3.9
-      cache: pip
 
   - name: Initialize env
     shell: bash


emaxx-google (approved):
lgtm, seems to work - thanks!

emaxx-google:
"Chumping" this PR, as the CI is broken to this and other reasons.

PR #1022, opened 2023-09-07
[PC/SC test] Don't assert handle/context non-zero

emaxx-google:
Remove test assertions that PC/SC handles and contexts must be nonzero.

This isn't guaranteed by the specs or the implementation, and while in theory frequent zeroes might be an indication of broken random generators, it's better to verify it in some other way.


emaxx-google:
This is a follow-up to https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/pull/1020#discussion_r1319099577.

ViktoriiaKovalova (approved):

PR #834, opened 2023-07-20
Switch to new Chromedriver download URLs

emaxx-google:
Per [1], Chromedriver shouldn't be downloaded anymore by a two-stage process, which was first obtaining the closest Chromedriver version to the installed Chrome version, and then actually downloading the latter's package. Now, starting from Chrome 115, it's guaranteed that every official release of Chrome has a corresponding version of Chromedriver available.

[1] https://groups.google.com/g/chromedriver-users/c/clpipqvOGjE

This also fixes #832, since the old way of downloading stopped working for Chrome >=115.

PR #1023, opened 2023-09-07
Choose latest Chromedriver per Chrome milestone

### LINKED ISSUE #1009: Downloading chromedriver breaks for certain versions

emaxx-google:
As in practice it's not guaranteed that every public Chrome release gets a corresponding Chromedriver release with the exact same version, we change env/initialize.sh to fetch using just milestone instead.

This is a different attempt to fix #1009. It's also a partial revert of #834 (as Chromedriver maintainers seem to have revived the convenient LATEST_RELEASE_ URLs).

### DIFF
--- a/env/initialize.sh
+++ b/env/initialize.sh
@@ -214,7 +214,7 @@ initialize_chromedriver() {
   log_message "Installing chromedriver..."
   rm -rf ./chromedriver
   # Determine the currently installed version of Chrome, in the
-  # major-minor-build format, e.g., 72.0.3626. (# The program's output format is
+  # major-minor-build format, e.g., 72.0.3626. (The program's output format is
   # "Google Chrome 72.0.3626.0").
   local chrome_version=$(google-chrome --version | cut -f 3 -d ' ' |
     cut -d '.' -f 1-3)
@@ -236,7 +236,6 @@ initialize_chromedriver() {
     log_error_message "Failed to unpack chromedriver executable; skipping chromedriver installation..."
     return
   fi
-  rm -f latest-versions-per-milestone-with-downloads.json
   rm -f chromedriver-linux64.zip
   log_message "chromedriver ${chromedriver_version} was installed successfully."
 }
--- a/env/initialize.sh
+++ b/env/initialize.sh
@@ -213,23 +213,20 @@ initialize_chromedriver() {
   fi
   log_message "Installing chromedriver..."
   rm -rf ./chromedriver
-  # Determine the currently installed version of Chrome, e.g., "72".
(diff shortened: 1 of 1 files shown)


env/initialize.sh
```diff
@@ -213,12 +213,23 @@ initialize_chromedriver() {
...
+    cut -f 1 -d '.')
+  # Download Chromedriver's releases information.
+  local chromedriver_releases_url="https://googlechromelabs.github.io/chrome-for-testing/latest-versions-per-milestone-with-downloads.json"
+  if ! wget "${chromedriver_releases_url}" ; then
+    log_error_message "Failed to fetch chromedriver releases at ${chromedriver_releases_url} ; skipping chromedriver installation..."
+    return
+  fi
+  # Parse the JSON to get the needed milestone's Chromedriver download URL.
+  local chromedriver_url=$(jq -r ".milestones[] | \
+    select(.milestone == \"${chrome_milestone}\") | .downloads | \
```
ViktoriiaKovalova, line 229: you can also use `https://googlechromelabs.github.io/chrome-for-testing/LATEST_RELEASE_{milestone}` or with first three numbers like `72.0.3626` instead of milestone.
is `jq` supposed to be installed by default?

emaxx-google: Good point, let me try it out. I think originally they wanted to kill these convenience URLs together with the switch to the new servers, but seems like they brought it back 2 weeks ago: https://github.com/GoogleChromeLabs/chrome-for-testing/pull/44

emaxx-google: So, lol, write-scratch-write-again strategy seems to be actual again! (context: #834 which stopped using `LATEST_RELEASE_`)

ViktoriiaKovalova: 🤯

emaxx-google: Done `https://googlechromelabs.github.io/chrome-for-testing/LATEST_RELEASE_`. So essentially it's a revert of #834, with the exception of different domains and modulo comments.

ViktoriiaKovalova (approved):
Thanks!

PR #1024, opened 2023-09-07
[style] format-code in Python, supports --base

emaxx-google:
Rewrite the format-code Shell script into Python, and make it support "--base=..." argument.

The refactoring is for better maintainability, and the "--base" arg allows reformatting the whole repository instead of the current diff.

### DIFF
--- a/format-code.sh
+++ b/format-code.sh
@@ -14,24 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This script formats the modified C++/JS files via clang-format according to
-# the Chromium Style Guides.
-# Only files that are known to Git and belong to the diff of the "main" branch
-# are considered.
-#
-# Note: Sorting of C/C++ #include's is currently disabled, since the standard
-# algorithm reorders them incorrectly (e.g., it puts includes within our project
-# into the same group as C system headers).
-
 set -eu
 
-# Go to the root directory of the repository.
-cd $(dirname $(realpath ${0}))
+>&2 echo "WARNING: The script has been moved. Use scripts/format-code.py instead."
 
-# Find all relevant files to be checked. Bail out if nothing is found.
-FILES=$(scripts/internal/find-files-for-linting.py "*.cc" "*.h" "*.js")
-[ "${FILES}" ] || exit 0
-
-# Run clang-format on every found file.
-echo "${FILES}" | \
-  xargs clang-format -i --style=Chromium --sort-includes=false
+$(dirname $(realpath ${0}))/scripts/format-code.py
--- a/scripts/format-code.py
+++ b/scripts/format-code.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python3
+
+# Copyright 2023 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


Fabian-Sommer (approved):
This seems to work, though the way that these two python scripts interact with each other is very sad

emaxx-google:
> This seems to work, though the way that these two python scripts interact with each other is very sad

Yeah it's slightly awkward, although one can argue it's Unix-philosophy :) smaller programs, doing separate items of work, and their combinations.

A more Python-friendly way would probably be to move the core stuff into a separate "lib" `.py` file that's included from "CLI" `.py` files, but I felt it's too heavy-weight for 2 use cases. Also the headache with Python's import story. I'll keep this option in mind though.

Fabian-Sommer:
> > This seems to work, though the way that these two python scripts interact with each other is very sad
> 
> Yeah it's slightly awkward, although one can argue it's Unix-philosophy :) smaller programs, doing separate items of work, and their combinations.
> 
> A more Python-friendly way would probably be to move the core stuff into a separate "lib" `.py` file that's included from "CLI" `.py` files, but I felt it's too heavy-weight for 2 use cases. Also the headache with Python's import story. I'll keep this option in mind though.

I think I would just have the lib part be a function in find-files-for-linting. The main() of find-files-for-linting would call it and print the output to CLI, while this file could call it directly.
I have not had headaches with python imports before - especially with two scripts that are co-located. Could you elaborate?

emaxx-google:
> > > This seems to work, though the way that these two python scripts interact with each other is very sad
> > 
> > 
> > Yeah it's slightly awkward, although one can argue it's Unix-philosophy :) smaller programs, doing separate items of work, and their combinations.
> > A more Python-friendly way would probably be to move the core stuff into a separate "lib" `.py` file that's included from "CLI" `.py` files, but I felt it's too heavy-weight for 2 use cases. Also the headache with Python's import story. I'll keep this option in mind though.
> 
> I think I would just have the lib part be a function in find-files-for-linting. The main() of find-files-for-linting would call it and print the output to CLI, while this file could call it directly. I have not had headaches with python imports before - especially with two scripts that are co-located. Could you elaborate?

OK, let me try that out!

As for imports - it's usually related to import search paths. A random example from Chromium repository - it's a nice combo of prepending, appending, preloading and reloading modules just to make the interpreter happy: https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/ash/policy/tools/generate_device_policy_remover.py;l=43-58;drc=4a8573cb240df29b0e4d9820303538fb28e31d84

emaxx-google:
> > > > This seems to work, though the way that these two python scripts interact with each other is very sad
> > > 
> > > 
> > > Yeah it's slightly awkward, although one can argue it's Unix-philosophy :) smaller programs, doing separate items of work, and their combinations.
> > > A more Python-friendly way would probably be to move the core stuff into a separate "lib" `.py` file that's included from "CLI" `.py` files, but I felt it's too heavy-weight for 2 use cases. Also the headache with Python's import story. I'll keep this option in mind though.
> > 
> > 
> > I think I would just have the lib part be a function in find-files-for-linting. The main() of find-files-for-linting would call it and print the output to CLI, while this file could call it directly. I have not had headaches with python imports before - especially with two scripts that are co-located. Could you elaborate?
> 
> OK, let me try that out!
> 
> As for imports - it's usually related to import search paths. A random example from Chromium repository - it's a nice combo of prepending, appending, preloading and reloading modules just to make the interpreter happy: https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/ash/policy/tools/generate_device_policy_remover.py;l=43-58;drc=4a8573cb240df29b0e4d9820303538fb28e31d84

Addressing this in #1045.

PR #1027, opened 2023-09-08
[JS-to-C++ test] Test SCardReconnect success

emaxx-google:
Test basic successful scenario for the SCardReconnect PC/SC function.


ViktoriiaKovalova (approved):
LGTM

PR #1028, opened 2023-09-08
[JS-to-C++ test] Fix flaky USB transfer exception

emaxx-google:
One code path in TestingSmartCardSimulation - the USB interrupt transfer that's intended to notify about card insertions/removals - was returning values of incorrect format, causing assertion failures down the stack.

Presumably this never caused observable test failures, since the tests happened to coincide with the CCID driver's internal task timing (which would mean the interrupt transfers are undertested as of now).


ViktoriiaKovalova:
Nice, I actually stopped seeing those consol errors from 'callCpp' with this patch.

ViktoriiaKovalova (approved):

emaxx-google:
> Nice, I actually stopped seeing those consol errors from 'callCpp' with this patch.

Yeah it was my first suspect last week, but I took extra time in attempts to get a reliable regression test for this. Unfortunately still haven't got it, hence landed as is...

PR #1029, opened 2023-09-08
[JS-to-C++ test] Boilerplate for //common/js tests

emaxx-google:
Add build target for //common/js/build/js_to_cxx_tests/, so that we can write tests for this common JS and C++ code.

This is preparation for implementing a regression test for #823.


Fabian-Sommer (approved):
LGTM

PR #1035, opened 2023-09-08
[chrome-api-jstocxxtest] Fix missing await

ViktoriiaKovalova:


emaxx-google (approved):
thanks for the quick turnaround

PR #1036, opened 2023-09-08
[chrome-api-jstocxxtest] Move creation of provider to set-up

ViktoriiaKovalova:
Move `createChromeApiProvider` to set-up.


emaxx-google (approved):
LGTM (seems fine to create it before triggering the PC/SC start, and probably in production we'll have this mode soon as well)

PR #1038, opened 2023-09-08
[chrome-api-jstocxxtest] Test SetAttrib

ViktoriiaKovalova:


smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -632,14 +651,168 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
...
+          /*requestId=*/ 111, SimulationConstants.COSMO_ID_70_ATR, 'SUCCESS');
+
+      await mockChromeApi
+          .dispatchEvent(
+              'onGetAttribRequested', /*requestId=*/ 111, sCardHandle,
+              API.SCARD_ATTR_ATR_STRING)
+          .$waitAndVerify();
+    },
+
+    // Test that SetAttrib fails for unsupported attributes.
```
emaxx-google, line 806: So only failure case in the end?

We stared at `TAG_IFD_DEVICE_REMOVED`, which seems like a settable attribute, but OTOH was added to be used by PC/SC internally.

smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -632,14 +651,168 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
...
+    'testSetAttrib_error': async function() {
+    expectReportPlainResult(/*requestId=*/ 111, 'INVALID_PARAMETER');
```
emaxx-google, line 808: Formatting is off.

(On that topic, I spent half of the day trying to make a CI check on this - #1032 - but it's surprisingly difficult because Ubuntu's default clang-format apparently formats code differently than our regular developer machines' ones.)

smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -261,6 +295,8 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
     mockChromeApi =
         new MockChromeApi(mockControl, testController.propertyReplacer);
     mockControl.$replayAll();
+
+    createChromeApiProvider();
```
emaxx-google, line 299: Many changes, like this, seem to have leaked from other PRs.

smart_card_connector_app/src/chrome-api-jstocxxtest.js
ViktoriiaKovalova: added another testcase with `TAG_IFD_DEVICE_REMOVED`

smart_card_connector_app/src/chrome-api-jstocxxtest.js
ViktoriiaKovalova: fixed

smart_card_connector_app/src/chrome-api-jstocxxtest.js
ViktoriiaKovalova: fixed

smart_card_connector_app/src/chrome-api-jstocxxtest.js
emaxx-google: thank you

smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -826,6 +829,28 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
...
+    },
+
+    // Test that SetAttrib works for supported attribute.
+    'testSetAttrib_success': async function() {
+      expectReportPlainResult(/*requestId=*/ 111, 'SUCCESS');
+
+      await mockChromeApi
+          .dispatchEvent(
+              'onSetAttribRequested', /*requestId=*/ 111, readerHandle,
+              TAG_IFD_DEVICE_REMOVED, /*data=*/ new Uint8Array(['0x0']))
```
emaxx-google, line 851: (Looks like the byte has to be nonzero to actually cause something to happen in the driver, but I guess it's not the point of the test here. It's enough that we verify the "success" here. If you can maybe drop a comment about this for future readers, but I don't mind if you land without it.)

emaxx-google (approved):
lgtm

PR #1039, opened 2023-09-08
[js] Use JS 2020, don't transpile to JS 2012

emaxx-google:
We don't keep backwards compatibility with such old ChromeOS versions that requires JS syntax from 2012. Hence tell Closure Compiler to use the 2020 syntax, corresponding to Chrome 81 that we currently specify as "minimum_chrome_version" for Smart Card Connector production builds.

This slightly reduces the final CRX files (-2% for Smart Card Connector). Also we hope this should reduce the frequency of mysterious Closure Compiler failures (like "Partial alias created for namespace $jscomp.scope, possibly due to await/yield transpilation").

### DIFF
--- a/common/make/js_building_common.mk
+++ b/common/make/js_building_common.mk
@@ -95,6 +95,8 @@ endif
 # Add common Closure Compiler flags.
 #
 # Explanation:
+# browser_featureset_year: Specify the year that's appropriate for the oldest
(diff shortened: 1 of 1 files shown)


emaxx-google:
Let the modernization happen.

ViktoriiaKovalova (approved):

PR #1041, opened 2023-09-08
[wasm] Force-kill threads and workers on disposal

emaxx-google:
Disposing of EmscriptenModule forcibly terminates all Emscripten threads and workers.

This is to avoid resource usage beyond the point when the module should've been logically disposed of. In particular, this should fix flaky OOM failures in the js_to_cxx tests that we run in succession now.


emaxx-google:
> JFYI the Emscripten authors currently believe there should be a simpler way, like calling `exit()`: https://groups.google.com/g/emscripten-discuss/c/f_eN9EW2O5o

But it might take a while to resolve that thread and/or get an alternative fix, meanwhile our tests are timing out already now :)

ViktoriiaKovalova (approved):
Thanks!

PR #1043, opened 2023-09-11
[style] format-code check-only mode

emaxx-google:
Add new parameter "--check-only" to the format-code.py script.

It's primarily intended to be used in automated checks, like CI actions, for checking for improperly-formatted code.


Fabian-Sommer (approved):
LGTM

PR #1045, opened 2023-09-11
[style] Import find-files-for-linting Python code

emaxx-google:
Change the users of the find-files-for-linting.py Python code (eslient.py and format-code.py) to import it as a Python module rather than run as a subprocess.

This should avoid unnecessary overhead and data conversions. The change renames the file to "find_files_for_linting.py" as dashes in file names aren't Python-friendly.


emaxx-google:
PTAL - this is to address your comment in #1024. Thanks.

scripts/eslint.py
```diff
@@ -17,6 +17,7 @@
 """Runs ESLint on all relevant JavaScript files in the repository."""
 
 import argparse
+from internal import find_files_for_linting
```
emaxx-google, line 20: The name "internal" sounds a bit too generic for an import here. But I didn't want to rename the folder just because of that, and another option - relative imports - doesn't work (`attempted relative import with no known parent package`).

Fabian-Sommer (approved):
LGTM - Thanks!

PR #1046, opened 2023-09-12
[JS-to-C++ test] Test SCardGetAttrib

emaxx-google:
Add basic success/failure test cases for SCardGetAttrib() PC/SC function.

### DIFF
--- a/smart_card_connector_app/src/pcsc-api-jstocxxtest.js
+++ b/smart_card_connector_app/src/pcsc-api-jstocxxtest.js
@@ -1082,6 +1082,102 @@ goog.exportSymbol('testPcscApi', {
       assert(called);
       assertEquals(result.getErrorCode(), API.SCARD_S_SUCCESS);
     },
+
+    // Test `SCardGetAttrib()` succeeds and returns information about the card.
+    'testSCardGetAttrib_successAtr': async function() {
+      await launchPcscServer(
+          /*initialDevices=*/[{
+            'id': 123,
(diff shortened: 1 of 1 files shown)


smart_card_connector_app/src/pcsc-api-jstocxxtest.js
```diff
@@ -1072,6 +1072,130 @@ goog.exportSymbol('testPcscApi', {
...
+    // Test `SCardGetAttrib()` succeeds and returns information about the card.
+    'testSCardGetAttrib_successAtr': async function() {
+      await launchPcscServer(
+          /*initialDevices=*/[{
+            'id': 123,
+            'type': SimulationConstants.GEMALTO_DEVICE_TYPE,
+            'cardType': SimulationConstants.COSMO_CARD_TYPE
+          }]);
+      const context = await establishContextOrThrow();
+      const cardHandle = await connectToCardOrThrow(
```
ViktoriiaKovalova, line 1085: from `Connect` doc: `Establishes a connection to the reader specified in * szReader.`. Maybe this function should be called `connectToReader` then(instead of card)?

ViktoriiaKovalova (approved):

smart_card_connector_app/src/pcsc-api-jstocxxtest.js
emaxx-google: Fair point; renamed.

PR #1047, opened 2023-09-12
[JS-to-C++ test] Suppress reader_init warnings

emaxx-google:
Make each PC/SC-server-based test set up listeners for reader_init C++-to-JS messages.

This is to avoid spurious warnings in the tests about unexpected messages.


ViktoriiaKovalova (approved):
Thanks!

PR #1048, opened 2023-09-12
JS Requester handles exceptions correctly

emaxx-google:
If sending a message failed, the Requester.postRequest() message should reject the returned promise instead of failing itself with an exception.

A real example when this can happen is when sending a message that causes a crash in an Emscripten module, because in that case the exception is bubbled up synchronously.

This is preparation for implementing a regression test for https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/issues/823.


Fabian-Sommer (approved):
LGTM

PR #1050, opened 2023-09-12
Update Closure Compiler to 20230802

emaxx-google:
from 20210104.

Nothing specific was required, but given a big number of fixes and general maintainability considerations makes sense to bump this dependency.


ViktoriiaKovalova (approved):
Thanks!

PR #1052, opened 2023-09-13
[js style] Disable semi-spacing ESLint

emaxx-google:
This rule disagrees with clang-format in some cases, e.g. the "for(...; ;...)" kind of loop.


ViktoriiaKovalova (approved):
Thanks!

PR #1053, opened 2023-09-13
[JS-to-C++ test] Test crashes

emaxx-google:
Add a test that crash in a C++ executable module gets correctly handled on the JS side, in particular marks the module as disposed.

This test verifies a slightly simpler scenario than the regression fixed in #823.

### DIFF
--- a/common/js/src/executable-module/executable-module-jstocxxtest.js
+++ b/common/js/src/executable-module/executable-module-jstocxxtest.js
@@ -21,6 +21,7 @@
 
 goog.require('GoogleSmartCard.ExecutableModule');
 goog.require('GoogleSmartCard.IntegrationTestController');
+goog.require('goog.testing.asserts');
 goog.require('goog.testing.jsunit');
 
 goog.setTestOnly();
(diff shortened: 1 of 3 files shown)


Fabian-Sommer (approved):
LGTM

common/js/src/executable-module/executable-module-jstocxxtest.js
```diff
@@ -48,6 +48,38 @@ goog.exportSymbol('testExecutableModule', {
...
+  // Test that after the C++ code crashes via `GOOGLE_SMART_CARD_CHECK()`, the
+  // JS module object gets disposed of.
+  'testCrashViaCheck': async function() {
+    await testController.setUpCppHelper(
+        'LoggingTestHelper', /*helperArgument=*/ {});
+
+    try {
+      await testController.sendMessageToCppHelper(
+          'LoggingTestHelper', 'crash-via-check');
+      fail('Unexpectedly proceeded beyond crash');
```
emaxx-google, line 60: Oh just realized this is wrong. `fail()` tries to abort the test by throwing an exception, but my `try`-`catch` suppresses it immediately :flushed:

emaxx-google: Fixed. That was embarrassing.

PR #1054, opened 2023-09-13
[makefile] Drop static lib constants

emaxx-google:
These constants don't bring much value; it seems more readable to just spell out static lib names in makefiles that need it.

This is the first step in eliminating the include.mk files and simplifying our build scripts.


Fabian-Sommer (approved):
LGTM

PR #1055, opened 2023-09-13
[js style] await instead of returning promise

emaxx-google:
Explicitly do "await" instead of returning a promise obtained from another function.

It's considered best practice, as it makes asynchronous points more explicit and also captures the current frame in the promise's stack trace.

This commit was prepared by enabling the "require-await" ESLint diagnostic, however we don't commit it because it has lots of false positives (on abstract classes, on functions returning "new Promise", on overrides that return result synchronously, etc.).


Fabian-Sommer (approved):
LGTM

PR #1056, opened 2023-09-14
[js promise] Move promisify to common

emaxx-google:
Make the "promisification" of Extension APIs available as a common helper that can be reused in other places beyond
libusb-to-chrome-usb-adaptor.js.


Fabian-Sommer (approved):
LGTM

PR #1057, opened 2023-09-14
JS Requester handles disposal-and-exception-while-sending

emaxx-google:
Improve the JS Requester class to gracefully handle this edge case: the send() call triggers not only an exception, but also the disposal of the message channel. The bug was that we were trying to reject the same request twice (first during disposal, second while handling the exception).

This is follow-up to the crash handling issue fixed in #823.


Fabian-Sommer (commented):
I don't understand how the unit test actually tests the code changes - all I see in the test is that 'If we send a request, some kind of exception happens'

Fabian-Sommer (approved):
LGTM

PR #1058, opened 2023-09-14
[JS-to-C++ test] Test crash with active PC/SC client

emaxx-google:
Test that a C++ crash in the executable module gets correctly handled in case there's an active PC/SC client. The module and the client should get disposed of without any unexpected exceptions.

This is a regression test for #823.

### DIFF
--- a/smart_card_connector_app/src/pcsc-api-jstocxxtest.js
+++ b/smart_card_connector_app/src/pcsc-api-jstocxxtest.js
@@ -20,6 +20,7 @@
  * Card Connector.
  */
 
+goog.require('GoogleSmartCard.AsyncAsserts');
 goog.require('GoogleSmartCard.DebugDump');
 goog.require('GoogleSmartCard.IntegrationTestController');
 goog.require('GoogleSmartCard.LibusbProxyReceiver');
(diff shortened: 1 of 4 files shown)


smart_card_connector_app/src/pcsc-api-jstocxxtest.js
```diff
@@ -1206,6 +1206,28 @@ goog.exportSymbol('testPcscApi', {
...
+      // Make a random call, to make sure all lazily created per-client state
+      // gets initialized.
+      await client.api.SCardIsValidContext(BAD_CONTEXT);
+
+      // Trigger the C++ module crash.
+      try {
+        await testController.sendMessageToCppHelper(
+            'LoggingTestHelper', 'crash-via-check');
+      } catch (e) {
+        // This is expected branch - discard the exception.
```
Fabian-Sommer, line 1226: This catches any exception - how does this then test that no unexpected exception occurred?

emaxx-google: Fixed, thanks for pointing this out.

Fabian-Sommer (approved):
LGTM

PR #1059, opened 2023-09-22
[JS tests] Unify the used Chrome binary path

emaxx-google:
Fix the tests executed via Chromedriver/Selenium to respect the CHOME_PATH make variable that lets specifying custom path to Chrome binary.


ViktoriiaKovalova (approved):
Thanks!

PR #1060, opened 2023-09-22
[refactoring] Rename _nacl to _webport in server

emaxx-google:
These source files are used in Emscripten mode as well, so the more generic "_webport" naming is more suitable.

This is a pure refactoring change.


Fabian-Sommer (approved):
LGTM

PR #1061, opened 2023-09-22
Fix permissions for find_files_for_linting.py

emaxx-google:
Add "+x" to this script as it can be executed standalone.


Fabian-Sommer (approved):
LGTM

PR #1063, opened 2023-09-22
[js promise] Use promisify in UserPromptingChecker

emaxx-google:
Simplify asynchronous code in UserPromptingChecker using the common promisify() helper function.

Also bullet-proof promisify() for environments that don't have full Chrome API definitions like `chrome.runtime` (this is also needed to make unit tests pass).


Fabian-Sommer (approved):
LGTM

PR #1064, opened 2023-09-22
[js promise] Use promisify in CrashLoopDetection

emaxx-google:
Simplify asynchronous code in CrashLoopDetection using the common promisify() helper function.


Fabian-Sommer (approved):
LGTM

PR #1070, opened 2023-09-22
[JS-to-C++ test] Test SCardConnect protocol mismatch

emaxx-google:
Add test cases for the successful/broken scenarios when the SCardConnect() is called multiple times with different protocols requested.

### DIFF
--- a/smart_card_connector_app/src/pcsc-api-jstocxxtest.js
+++ b/smart_card_connector_app/src/pcsc-api-jstocxxtest.js
@@ -1029,7 +1029,7 @@ goog.exportSymbol('testPcscApi', {
             fail(`Unexpected error in first SCardConnect: ${errorCode}`);
           });
       assertTrue(called);
-      assertEquals(secondResult.getErrorCode(), API.SCARD_E_PROTO_MISMATCH);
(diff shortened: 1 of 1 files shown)


smart_card_connector_app/src/pcsc-api-jstocxxtest.js
```diff
@@ -983,6 +983,125 @@ goog.exportSymbol('testPcscApi', {
...
+          (sCardHandleArg, activeProtocol) => {
+            called = true;
+            assert(Number.isInteger(sCardHandle));
+            assertEquals(activeProtocol, API.SCARD_PROTOCOL_T1);
+          },
+          (errorCode) => {
+            fail(`Unexpected error in first SCardConnect: ${errorCode}`);
+          });
+      assertTrue(called);
+      assertEquals(secondResult.getErrorCode(), API.SCARD_E_PROTO_MISMATCH);
```
ViktoriiaKovalova, line 1032: why should it return error code? it looks like you expect success here.

ViktoriiaKovalova (approved):

smart_card_connector_app/src/pcsc-api-jstocxxtest.js
emaxx-google: done, thanks

PR #1072, opened 2023-09-22
[PC/SC test] More EndTransaction errors

emaxx-google:
Add tests for the cases of calling SCardEndTransaction() when there's no actual transaction active.


ViktoriiaKovalova (approved):
thank you!

PR #1079, opened 2023-09-27
Cherry-pick CCID fix for 10-min hang on shutdown

### LINKED ISSUE #1078: Test shutdown hangs for 10min if reader is added shortly before
Pseudocode of a reproducer:

```
  LaunchPcscServer();
  SetSimulatedDevices([{...}]);
  ShutDownPcscServer();  // <-- hangs for 10 min here
```

emaxx-google:
Apply the fix that was submitted upstream:
https://github.com/LudovicRousseau/CCID/pull/116.

This fixes #1078.


ViktoriiaKovalova (approved):
Thanks for fixing this!

emaxx-google:
I suspect this PR might've caused some regressions we've observed... Apparently it's possible for the driver to re-start the interrupt transfer USB communication after stopping it before, however this commit makes it impossible. This makes the CCID code fall back from the interrupt-based detection (with 10-minute timeout) to a quick polling-based detection (with 400 ms timeout), which in turn combined with the old issue #1104 means that we open too many concurrent USB transfers and eventually break after a few hours.

PR #1080, opened 2023-09-27
Test for 10-minute shutdown hang

emaxx-google:
Add unit tests that verify the PC/SC shutdown process doesn't take unnecessarily long, especially in the "reader added shortly before the shutdown" scenarios.

This is regression test for #1078.

### DIFF
--- a/smart_card_connector_app/src/application_unittest.cc
+++ b/smart_card_connector_app/src/application_unittest.cc
@@ -75,6 +75,11 @@ namespace google_smart_card {
 
 namespace {
 
+// The upper boundary on how long the `Application` shutdown may take. Used
+// primarily to catch mistakes when the shutdown is hung completely or is
+// wrongly blocked on multi-minute timeouts.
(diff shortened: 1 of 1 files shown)


smart_card_connector_app/src/application_unittest.cc
```diff
@@ -2609,4 +2626,40 @@ TEST_F(SmartCardConnectorApplicationTwoClientsTest,
...
+  device.id = 123;
+  device.type = TestingSmartCardSimulation::DeviceType::kGemaltoPcTwinReader;
+  SetUsbDevices({device});
+
+  // Nothing to assert explicitly here - the relevant assertion is in the test
+  // fixture destructor that verifies that the shutdown fits into a timeout.
+}
+
+// Same as above, but additionally makes a PC/SC call roundtrip to affect the
+// timings and increase the probabilty of calling bugs if they happen to appear.
```
ViktoriiaKovalova, line 2648: nit: catching bugs*

ViktoriiaKovalova (approved):

PR #1081, opened 2023-09-27
[PC/SC test] SetAttrib success and more error cases

emaxx-google:
Test a few more scenarios in the SCardSetAttrib() method, adding first test cases for successful cases as well as extending the set of tests for error cases.


Fabian-Sommer (approved):
LGTM

PR #1082, opened 2023-09-29
[PC/SC test] Simulate Yubikey 4C

emaxx-google:
Implement simulation of a new device - Yubico Yubikey 4C. There's no particular reason this device is chosen, except that to increase test coverage for various types of devices.

The simulation that's added is minimalistic and makes only the basic interactions (plug in/out, obtain ATR) work. More can be added later if necessary.


ViktoriiaKovalova (approved):
LGTM

PR #1084, opened 2023-10-04
[PC/SC tests] Test Connect on all simulated readers

emaxx-google:
Add tests that execute SCardConnect/SCardDisconnect in pattern tests that are instantiated for every card+reader combination we simulate in tests.

We also slightly change the structure of these templated tests, so that we can have a single test for such common functionality that's the same on both "regular readers" and "readers with built-in cards" (e.g., Yubikeys).


Fabian-Sommer (approved):
LGTM

PR #1085, opened 2023-10-04
[Web API] Remove expose_chrome_smart_card_api policy check

ViktoriiaKovalova:
This CL makes ChromeApiProvider created by default, not guarded by the managed schema policy.

### DIFF
--- a/smart_card_connector_app/src/background.js
+++ b/smart_card_connector_app/src/background.js
@@ -166,42 +166,16 @@ if (GSC.ExecutableModule.TOOLCHAIN ===
   GSC.BackgroundPageUnloadPreventing.enable();
 }
 
-/** @type {GSC.ConnectorApp.ChromeApiProvider?} */
-let chromeApiProvider = null;
-
 if (chrome.smartCardProviderPrivate !== undefined) {
-  updateChromeApiProviderAvailability();
-  chrome.storage.onChanged.addListener(updateChromeApiProviderAvailability);
+  // This object's lifetime is bound to the message channel's one, hence we
+  // don't need to store it in a variable.
+  new GSC.ConnectorApp.ChromeApiProvider(
+      executableModule.getMessageChannel(), pcscLiteReadinessTracker.promise);
 } else {
   goog.log.info(
       logger,
       `chrome.smartCardProviderPrivate is not available. ` +
           `Requests from Web Smart Card API will not be handled.`);
-}
-
-/**
- * Updates the Chrome API provider availability based on extension policy.
- */
-function updateChromeApiProviderAvailability() {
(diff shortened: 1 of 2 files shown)


smart_card_connector_app/src/managed_storage_schema.json
```diff
@@ -3,7 +3,7 @@
   "properties": {
     "expose_chrome_smart_card_api": {
       "title": "Handle requests from chrome.smartCardProviderPrivate API.",
-      "description": "If set to true, the app will subscribe to chrome.smartCardProviderPrivate API events and handle the requests. This is needed for Web Smart Card API.",
+      "description": "DEPRECATED. If set to true, the app will subscribe to chrome.smartCardProviderPrivate API events and handle the requests. This is needed for Web Smart Card API.",
```
emaxx-google, line 6: nit: Let's also edit that old description to say that this functionality is now on by default.

smart_card_connector_app/src/background.js
```diff
@@ -166,44 +166,18 @@ if (GSC.ExecutableModule.TOOLCHAIN ===
...
 
-/** @type {GSC.ConnectorApp.ChromeApiProvider?} */
-let chromeApiProvider = null;
-
 if (chrome.smartCardProviderPrivate !== undefined) {
-  updateChromeApiProviderAvailability();
-  chrome.storage.onChanged.addListener(updateChromeApiProviderAvailability);
+  // This object's lifetime is bound to the message channel's one, hence we
+  // don't need to store it in a variable.
+  new GSC.ConnectorApp.ChromeApiProvider(
```
emaxx-google, line 172: I feel like I already asked this 10 times in the past, but keep forgetting the answer... What's the story with the backwards-incompatible changes in the API that were made by Chrome a few versions ago? Will the rest of the Connector's functionality work fine there?

ViktoriiaKovalova: So the interesting use case is those semi-old versions on the dev channel where smartCardProviderPrivate is already available but not fully implemented.
There the app will subscribe to all the events that are implemented (for the rest it'll just print some warning messages).
And then if someone will enable the SmartCard feature flag and try to send some requests from IWA, I guess there could be some errors for GetStatusChange requests. 
But I don't think someone will ever do this :)

smart_card_connector_app/src/managed_storage_schema.json
ViktoriiaKovalova: done

smart_card_connector_app/src/background.js
emaxx-google: Ack, sounds fine. Thanks for capturing this in the comment thread here.

emaxx-google (approved):
Hooray!

PR #1088, opened 2023-10-05
[libusb] Fix use-after-free in submit_transfer 

emaxx-google:
Fix a use-after-free bug in our libusb_submit_transfer() implementation
(i.e., LibusbJsProxy::LibusbSubmitTransfer()). The bug was that this
function could continue reading from a libusb_transfer* pointer after
it's been placed into a shared data structure, where it can be seen,
populated and destroyed by other threads.

We also add a PC/SC test that triggered this bug with some probability
(visible on the asan_testing bot).

This fixes https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/issues/1090.


emaxx-google:
The test is flaky - apparently at least one of the issues is a genuine bug: #1090.

emaxx-google:
> The test is flaky - apparently at least one of the issues is a genuine bug: #1090.

Now the PR has the fix for it as well.

Fabian-Sommer (approved):
LGTM

PR #1089, opened 2023-10-05
Release Connector 1.4.0

emaxx-google:
We also switch from the previous four-number version to a proper SemVer: MAJOR.MINOR.PATCH.

Not only SemVer is the more standard versioning scheme, but it also
happens to work better for logging, because groups of four
dot-separated numbers are stripped out from ChromeOS logs by
anonymizer tools.


ViktoriiaKovalova (approved):
Thanks!

PR #1093, opened 2023-10-05
[logging] Fix wrong number in the clients_manager log

emaxx-google:
When shutting down clients, we've always been printing zero, instead of the real number of clients. This change fixes this minor logging glitch.


ViktoriiaKovalova (approved):
Nice, thanks for the fix!

PR #1101, opened 2023-10-23
[PC/SC test] Test delayed card events with GetStatusChange

emaxx-google:
Duplicate the existing tests for SCardGetStatusChange() in the card insertion/removal scenarios, with introducing a variant of these tests that delays the card insertion/removal by some (hardcoded) interval.

This adds test coverage for some internal implementation codepaths, because the previously existing tests were only triggering events that are immediately discovered by SCardGetStatusChange before it enters the polling loop, and before it starts relying on asynchronous USB notifications from the reader (via interrupt USB transfers).

### DIFF
--- a/smart_card_connector_app/src/application_unittest.cc
+++ b/smart_card_connector_app/src/application_unittest.cc
@@ -1305,9 +1305,10 @@ TEST_F(SmartCardConnectorApplicationSingleClientTest,
                        TestingSmartCardSimulation::CardType::kCosmoId70)));
 }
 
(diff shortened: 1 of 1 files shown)


Fabian-Sommer (approved):
LGTM - optional nit: I would refer to the tests by their name instead of 'previous test', in case some other test is inserted in between

emaxx-google:
> LGTM - optional nit: I would refer to the tests by their name instead of 'previous test', in case some other test is inserted in between

sure, done. thanks!

Issue #1104, opened 2023-10-27
Avoid duplicate concurrent transfer USB API calls


emaxx-google:
Currently, every Libusb transfer made by the CCID driver results in a chrome.usb/WebUSB API call.

This is unnecessary and wasteful in some cases, because we might already have equivalent in-flight API calls for the previously canceled transfers. Besides consuming Chrome resources, these extra calls can reportedly lead to hitting the OS kernel limits on the number of active transfers, which completely breaks the USB communication.

Instead, we should enqueue the requests and make sure we only make one JS API call at a time (among all input transfers that read from the same endpoint).

Issue #462, opened 2021-11-25
[libusb] Improve transfer results ordering


emaxx-google:
Our current libusb implementation doesn't guarantee any particular ordering in case there are multiple transfers simultaneously in flight. This might or might not be a problem, depending on the needs of the consumer application.

There are several reasons that cause undetermined ordering:
1. The transfer results from the JS API are temporarily put into a map that's ordered by some key (which is not a time nor a counter, but some other data).
2. Pending transfers are stored in a set-inside-map, so they're not ordered either.
3. [already fixed] Until recently, synchronous transfers had special implementation, such that synchronous transfers were fighting with asynchronous event loops (and who wins was determined, basically, on the order in which the POSIX conditional variable notified callers).

emaxx-google:
I believe #1105 and #1107 resolved this.

PR #1105, opened 2023-10-27
[libusb] Handle transfers in FIFO

emaxx-google:
When choosing which of active Libusb transfers should receive a reply from the JS API, use the FIFO order rather than a random order. Also change the internal structures so that FIFO is used uniformly, which we'll use in follow-ups for correctly enqueueing JS API calls.

The FIFO order is more intuitive and better corresponds to Libusb behavior on other platforms. Moreover, the FIFO property will be crucial to avoid the situation when an enqueued JS API call is discarded because the transfer happened to be resolved by the call originally triggered by a different transfer.

This contributes to #462 and #1104.

### DIFF
--- a/third_party/libusb/webport/src/libusb_js_proxy_unittest.cc
+++ b/third_party/libusb/webport/src/libusb_js_proxy_unittest.cc
@@ -1590,6 +1590,70 @@ TEST_P(LibusbJsProxyWithDeviceTest, AsyncInputControlTransferDataRerouting) {
   libusb()->LibusbFreeTransfer(transfer2);
 }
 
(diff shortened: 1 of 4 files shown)


Fabian-Sommer (approved):
LGTM - also, I realized that I was reviewing these CLs in the wrong order, which explains why I didn't know some of the code during my earlier reviews :)

emaxx-google:
> LGTM - also, I realized that I was reviewing these CLs in the wrong order, which explains why I didn't know some of the code during my earlier reviews :)

Oops. Next time I'll explicitly mention the numbers in PR titles or comments. Thanks for the reviews; I've added some tests.

PR #1108, opened 2023-10-27
[wasm] Increase default stack size in Release

emaxx-google:
Set the stack size to 1 MiB in both Debug and Release modes.

Previously we were doing this only in Debug, and left the default 64 KiB otherwise, but with recent changes it turned out that we're besides this (tough) limit. In the future we might want to investigate whether something smaller than 1 MiB would make sense, however for now WebAssembly builds aren't shipped anywhere, and we just want to make the CI bots happy on follow-up changes.

This change contributes to #1104.


Fabian-Sommer (approved):
LGTM

PR #1109, opened 2023-10-27
WIP fix CCID polling transfer restart

emaxx-google:
We should support restarting the interrupt transfer after it's been stopped in the past.

This fixes the breakage of the polling transfer after the "insert card => SCardConnect => SCardDisconnect => remove card" scenario, which led the application to actively poll the reader every 400 ms instead of a 10-minute transfer, which in turn led to excessively big number of inflight chrome.usb transfers and various breakages in production.


emaxx-google:
This was landed in different PRs.

PR #1110, opened 2023-10-30
[ccid] Cherry-pick upstream fix for excessive polling

emaxx-google:
Cherry-pick the fix we landed into the upstream CCID repository (sadly, for the regression that we introduced previously there).

This fixes the polling_transfer functionality to continue working correctly after a card has been connected and disconnected once. The bug was that the polling_transfer was immediately canceled in such case, leading to the driver falling back to an active polling of the reader every 400ms instead of passive reading transfers that are restarted only every 10 minutes. On ChromeOS, this suboptimal behavior, combined with the fact that we make a new infinitely-timed chrome.usb API call for every transfer, led to the system resource exhaustion after a few hours and some other race conditions. (On non-ChromeOS platforms, the bug was less severe as the real Libusb library doesn't have this suboptimal behavior.)


Fabian-Sommer (approved):

emaxx-google:
OK, it seems confirmed that this is the right fix for all the recent regressions we've been observing.

PR #1118, opened 2024-01-08
Update CCID to 1.5.5

emaxx-google:
Also drop four of our local patches that have gone upstream now.


Fabian-Sommer (approved):
LGTM - I was a bit confused why the for some patches the diffs got deleted without changes to the actual files, but I guess those patches just got upstreamed :)

emaxx-google:
> LGTM - I was a bit confused why the for some patches the diffs got deleted without changes to the actual files, but I guess those patches just got upstreamed :)

Exactly! Let me add this to the CL description.

PR #1123, opened 2024-01-24
[JS-to-C++ tests] Bulletproof teardowns

emaxx-google:
Make the code in test cases' tearDown functions resistant to failures in setUps. This avoids spurious errors like "reading the dispose property of a null object" after the real test failures.

This commit is a no-op for the successfully built tree, it's only intended to improve troubleshooting of test failures.

### DIFF
--- a/smart_card_connector_app/src/chrome-api-jstocxxtest.js
+++ b/smart_card_connector_app/src/chrome-api-jstocxxtest.js
@@ -62,8 +62,8 @@ let libusbProxyReceiver;
 let pcscReadinessTracker;
 /** @type {MockChromeApi?} */
 let mockChromeApi;
-/** @type {!goog.testing.MockControl|undefined} */
-let mockControl;
+/** @type {!goog.testing.MockControl|null} */
+let mockControl = null;
 /** @type {ChromeApiProvider?} */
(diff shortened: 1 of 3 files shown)


smart_card_connector_app/src/chrome-api-jstocxxtest.js
```diff
@@ -1086,11 +1086,19 @@ goog.exportSymbol('testChromeApiProviderToCpp', {
...
+      if (chromeApiProvider) {
+        assertTrue(chromeApiProvider.isDisposed());
+      }
+      if (mockControl) {
+        // Check all mock expectations are satisfied.
+        mockControl.$verifyAll();
+      }
     } finally {
       chromeApiProvider = null;
       pcscReadinessTracker = null;
```
ViktoriiaKovalova, line 1104: I guess `mockControl = null` also is needed here?
```suggestion
      mockControl = null;
      pcscReadinessTracker = null;
```

ViktoriiaKovalova (approved):
Thanks, Maksim!

emaxx-google:
OK, I had to fix the suggestion, so that the code compiles.

Thanks for the review.

Issue #1126, opened 2024-01-30
Popup dialogs are created with wrong width


emaxx-google:
When a dialog (e.g., the built-in PIN dialog in example_cpp_smart_card_client_app) is opened, it gets a wrong width instead of the one specified in the code.

emaxx-google:
I think it regressed in #518: this line shouldn't have been deleted:

```
Object.assign(createWindowOptions, opt_createWindowOptionsOverrides);
```

emaxx-google:
And there's another issue introduced in #404, when the format of the "width"/"height" parameters has changed, but the example PIN dialog callsites weren't updated accordingly.

PR #1128, opened 2024-01-30
[ui] Fix width parameter format in example PIN dlg

emaxx-google:
The "width" property should be passed without being wrapped into the "innerBounds" property.

This fixes the regression that was introduced in #404. Tracking issue #1126.


Fabian-Sommer (approved):

emaxx-google:
Thank you for the reviews!

Issue #1129, opened 2024-02-06
Switch Smart Card Connector to Extensions Manifest V3


emaxx-google:
Currently, Smart Card Connector is a Chrome App. A couple of years ago we started the work on migrating it an Extension (https://github.com/GoogleChromeLabs/chromeos_smart_card_connector/projects/3), however originally this was targeting Manifest V2 Extensions.

With the new deprecation timelines (https://developer.chrome.com/docs/extensions/develop/migrate/mv2-deprecation-timeline), we should switch to Extensions Manifest V3 instead.

Test18415:
Is it possible that this could this be an issue for us, considering the fact that we use a chrome kiosk app with manifest version 2 and also using smart card connector reader chrome app, or does this only apply for extension of smart card connector?

Test18415:
Okay, thank you, that is good news.

emaxx-google:
Capturing one thought: probably #1152 should be enhanced in a way similar to the code touched in #1155. In other words, when we `await` on an incoming port, the port should initially freeze incoming messages until it's marked as "ready". Otherwise we risk the following race condition:

1. receiver: creates `PortMessageChannelWaiter`, and starts `await`ing on it
1. sender: `port = chrome.runtime.connect(...)`
1. receiver: `chrome.runtime.onConnect` received by `PortMessageChannelWaiter`, and the `Port` is created
1. receiver: the `await`ed promise starts being resolved
1. sender: `port.postMessage(...)`
1. receiver: the `Port` receives the message, and discards it because there's been no service registered for incoming messages yet
1. receiver: the `await`ed promise gets resolved, and the code registers services, but the message that has been sent is already lost

So the fix would be to move the message handling to happen at a later step ("8") instead of step 6.

bderuijsscher:
Our current Chrome App (deployed via Chrome Enterprise in Kiosk mode) heavily relies on this connector. As Google is pushing forward with Chrome App deprecation, is there a migration path or advice for us to consider, specifically for integration with this connector?

emaxx-google:
The MV3 version of Smart Card Connector is still not production-ready - we're working on that.

The migration path will be fully transparent though: once we're done the MV3 version will be published on WebStore under the same ID, and users will receive it automatically.

PR #1130, opened 2024-02-06
[mv3] Use globalThis instead of window

emaxx-google:
Switch a couple of places that use the "window" object to use "globalThis" instead. The latter is a better equivalent because it works in Manifest V3 Service Workers as well (in a Service Worker, "self" should be used instead of "window"; "globalThis" is an alias to either of those).

This contributes to the effort tracked by #1129.

### DIFF
--- a/common/js/src/random.js
+++ b/common/js/src/random.js
@@ -36,12 +36,7 @@ const GSC = GoogleSmartCard;
  */
 GSC.Random.randomIntegerNumber = function() {
   const randomBytes = new Uint8Array(RANDOM_INTEGER_BYTE_COUNT);
-  if (typeof window !== 'undefined') {
-    window.crypto.getRandomValues(randomBytes);
-  } else {
-    // We're likely inside a Service Worker.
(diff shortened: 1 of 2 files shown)


common/js/src/random.js
```diff
@@ -36,12 +36,7 @@ const GSC = GoogleSmartCard;
  */
 GSC.Random.randomIntegerNumber = function() {
   const randomBytes = new Uint8Array(RANDOM_INTEGER_BYTE_COUNT);
-  if (typeof window !== 'undefined') {
-    window.crypto.getRandomValues(randomBytes);
-  } else {
-    // We're likely inside a Service Worker.
-    self.crypto.getRandomValues(randomBytes);
-  }
+  globalThis.crypto.getRandomValues(randomBytes);
```
emaxx-google, line 39: The change in this file is a pure refactoring: I didn't know about `globalThis` back then.

Fabian-Sommer (approved):

emaxx-google:
Thanks for the reviews!

PR #1132, opened 2024-02-18
[chromedriver] Update chromedriver download URL

emaxx-google:
Google changed the Chromedriver download domain to "storage.googleapis.com".


emaxx-google:
Landing directly as the patch is trivial and to unblock the CI.

PR #1137, opened 2024-04-02
[mv3] Skip BackgroundPageUnloadPreventing

emaxx-google:
Skip executing the "BackgroundPageUnloadPreventing" code when built in the "extension" packaging mode.

This code doesn't work under extensions manifest v3, because DOM operations aren't available for Service Workers. We also don't need this code because Chrome will itself keep the application alive (based on the open USB communication and/or message channels to other extensions).


Fabian-Sommer (approved):
LGTM

emaxx-google:
Also updated the related comment in the example client sources.

emaxx-google:
I've updated the PR to also skip packaging the now-unnecessary files (`background-page-unload-preventing-iframe.js` and `background-page-unload-preventing-iframe.html`) in the corresponding configuration.

PR #1139, opened 2024-04-04
[libusb] Support proxying through message channels

emaxx-google:
Perform data massaging when handling USB requests, so that they can be sent through message pipes. WebUSB uses ArrayBuffer and DataView types for input/output parameters, however for sending via extension messages we have to convert them to/from arrays of integers-bytes.

The underlying motivation is the migration to manifest v3 and WebAssembly, because due to multi-threading issues we have to move the WebAssembly module from the Service Worker (where most API calls, including WebUSB APIs will happen) into a separate Offscreen Document. So each USB request will have to be proxied as a request-response pair of messages between the Offscreen Document and the Service Worker.


Fabian-Sommer (approved):
LGTM

PR #1141, opened 2024-04-05
[mv3] Put Wasm module into Offscreen Document

emaxx-google:
When built in the Emscripten+Extension mode, run the Emscripten module in the Offscreen Document instead of the Service Worker.

This is needed because multi-threaded Emscripten modules require creating Workers, which Chrome doesn't currently support from Service Workers - see https://crbug.com/40772041. Hence this commit moves the module into a separate Offscreen Document which doesn't have this limitation. The rest of the background logic remains in the Service Worker because Offscreen Documents have their own strict limitations (most importantly, only chrome.runtime APIs are exposed to them).

### DIFF
--- a/common/js/src/executable-module/emscripten-in-offscreen-doc-main-script.js
+++ b/common/js/src/executable-module/emscripten-in-offscreen-doc-main-script.js
@@ -43,6 +43,10 @@ const moduleName =
     url.searchParams.get(GSC.OffscreenDocEmscriptenModule.URL_PARAM);
 GSC.Logging.check(moduleName);
 goog.asserts.assert(moduleName);
+// A quick sanity check against malicious URL parameters (in case the user was
+// tricked into opening our page via chrome-extension://). We don't want to
+// allow loading&executing arbitrary URLs.
+GSC.Logging.check(/^[0-9a-zA-Z_]+$/.test(moduleName));
 
 const emscriptenModule = new GSC.EmscriptenModule(moduleName);
 
--- a/common/js/src/executable-module/emscripten-in-offscreen-doc-main-script.js
+++ b/common/js/src/executable-module/emscripten-in-offscreen-doc-main-script.js
@@ -1,0 +1,72 @@
+/**
+ * @license
+ * Copyright 2024 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview This file contains code to be directly executed in the
+ * Offscreen Document in order to load an Emscripten module.
+ *
+ * This is to be treated an internal implementation detail of
+ * offscreen-doc-emscripten-module.js and shouldn't be called by client code
+ * directly. It's expected to be compiled and packaged as the file called
+ * "emscripten-offscreen-doc.js" (see "emscripten-offscreen-doc.html").
+ */
+
+goog.provide('GoogleSmartCard.EmscriptenInOffscreenDocMain');
+
+goog.require('GoogleSmartCard.EmscriptenModule');
+goog.require('GoogleSmartCard.Logging');
+goog.require('GoogleSmartCard.OffscreenDocEmscriptenModule');
+goog.require('GoogleSmartCard.PortMessageChannel');
+goog.require('goog.asserts');
+
+goog.scope(function() {
+
+const GSC = GoogleSmartCard;
+
+// The URL parameters specify the name of the Emscripten module to be loaded.
+const url = new URL(globalThis.location.href);
+const moduleName =
+    url.searchParams.get(GSC.OffscreenDocEmscriptenModule.URL_PARAM);
+GSC.Logging.check(moduleName);
+goog.asserts.assert(moduleName);
+
+const emscriptenModule = new GSC.EmscriptenModule(moduleName);
+
+// Pipe the module's messaging channel to the port to the opener.
+const moduleChannel = emscriptenModule.getMessageChannel();
+const portChannel = new GSC.PortMessageChannel(chrome.runtime.connect(
+    {'name': GSC.OffscreenDocEmscriptenModule.MESSAGING_PORT_NAME}));
+portChannel.registerDefaultService((serviceName, payload) => {
+  moduleChannel.send(serviceName, payload);
+});
+moduleChannel.registerDefaultService((serviceName, payload) => {
+  portChannel.send(serviceName, payload);
+});
+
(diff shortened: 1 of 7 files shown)


common/js/src/executable-module/offscreen-doc-emscripten-module.js
```diff
@@ -0,0 +1,256 @@
...
+    // query parameters.
+    const url = new URL(OFFSCREEN_DOC_URL, globalThis.location.href);
+    url.searchParams.append(URL_PARAM, this.moduleName_);
+    chrome.offscreen.createDocument(
+        {
+          'url': url.toString(),
+          'reasons': OFFSCREEN_DOC_REASONS,
+          'justification': OFFSCREEN_DOC_JUSTIFICATION,
+        },
+        () => this.onOffscreenDocCreated_());
```
Fabian-Sommer, line 109: Most of SCC is relying on callbacks, but since this API is MV3 exclusive anyway, we could consider using promises instead here - IMO they make the code more readable.

common/js/src/executable-module/offscreen-doc-emscripten-module.js
```diff
@@ -0,0 +1,256 @@
...
+      this.offscreenDocActive_ = true;
+    }
+
+    if (this.offscreenDocActive_ && this.isDisposed()) {
+      // Edge case: we got destroyed while waiting for the
+      // `chrome.offscreen.createDocument()` completion; make sure to not leave
+      // the doc in that case as well.
+      chrome.offscreen.closeDocument();
+    } else if (!this.offscreenDocActive_) {
+      this.dispose();
```
Fabian-Sommer, line 152: Unless this function is called multiple times, I think this line can be put after line 141

smart_card_connector_app/src/background.js
```diff
@@ -151,7 +155,6 @@ executableModule.startLoading();
 if (GSC.Packaging.MODE === GSC.Packaging.Mode.APP)
   chrome.app.runtime.onLaunched.addListener(launchedListener);
 
-chrome.runtime.onConnect.addListener(connectionListener);
```
Fabian-Sommer, line 154: This change seems to affect the APP mode, while the new code only runs in the EXTENSION mode - was this not needed for an app?

Fabian-Sommer (changes requested):

emaxx-google:
All fair points! Let me rewrite some of the code to `await`s and split it up.

smart_card_connector_app/src/background.js
emaxx-google: Correct, it was useless code :see_no_evil: Extracted into #1154.

common/js/src/executable-module/offscreen-doc-emscripten-module.js
emaxx-google: Done - the linear logic is definitely more readable; thanks for the suggestion!

The only downside is that we need to handle the scenario "what if we got disposed in the middle of this long async function", but it's not too bad I think.

common/js/src/executable-module/offscreen-doc-emscripten-module.js
emaxx-google: Ack - this chunk of code doesn't exist anymore after the rewrite.

emaxx-google:
Ready for the next review iteration. This PR depends now on a few other ones, but as dependencies are horrible in GitHub I'm not trying to form a chain; hence tests fail here now, but rebasing will fix that later.

common/js/src/executable-module/emscripten-in-offscreen-doc-main-script.js
```diff
@@ -0,0 +1,72 @@
...
+goog.require('goog.asserts');
+
+goog.scope(function() {
+
+const GSC = GoogleSmartCard;
+
+// The URL parameters specify the name of the Emscripten module to be loaded.
+const url = new URL(globalThis.location.href);
+const moduleName =
+    url.searchParams.get(GSC.OffscreenDocEmscriptenModule.URL_PARAM);
```
emaxx-google, line 43: Starting to think I should better sanitize this parameter, to avoid possibility of XSS-style injection. I.e., even if someone tricks the user into opening `chrome-extension://khp.../emscripten-offscreen-doc.html?module_name=https://attacker.com/script` it shouldn't end up actually loading that third-party script.

emaxx-google: done

common/js/src/executable-module/emscripten-in-offscreen-doc-main-script.js
```diff
@@ -0,0 +1,76 @@
...
+GSC.Logging.check(moduleName);
+goog.asserts.assert(moduleName);
+// A quick sanity check against malicious URL parameters (in case the user was
+// tricked into opening our page via chrome-extension://). We don't want to
+// allow loading&executing arbitrary URLs.
+GSC.Logging.check(/^[0-9a-zA-Z_]+$/.test(moduleName));
```
emaxx-google, line 49: It's a poor man's check, but should be secure enough. The name is anyway normally `executable_module` unless this could starts being used for other things in the future. Standard JS doesn't have good built-in methods, and while Closure Library has something this library is deprecated :( , so I don't want to add new dependencies onto it...

common/js/src/executable-module/offscreen-doc-emscripten-module.js
```diff
@@ -0,0 +1,217 @@
...
+    await this.waitMessage_(STATUS_LOADED);
+  }
+
+  /**
+   * Returns a promise that gets rejected on receiving the "disposed" status
+   * message from the Offscreen Document. Also disposes of `this`.
+   * @return {!Promise<void>}
+   * @private
+   */
+  async waitDisposedMessage_() {
```
Fabian-Sommer, line 192: nit: waitForDisposedMessage

common/js/src/executable-module/offscreen-doc-emscripten-module.js
```diff
@@ -0,0 +1,217 @@
...
+    this.messageChannel_.setReady();
+  }
+
+  /**
+   * Returns a promise that gets resolved on receiving the "loaded" status
+   * message from the Offscreen Document.
+   * @return {!Promise<void>}
+   * @private
+   */
+  async waitLoadedMessage_() {
```
Fabian-Sommer, line 182: nit: waitForLoadedMessage

common/js/src/executable-module/offscreen-doc-emscripten-module.js
```diff
@@ -0,0 +1,217 @@
...
+    const statusChannel = await statusChannelWaiter.getPromise();
+    if (this.isDisposed()) {
+      statusChannel.dispose();
+      throw new Error('Disposed');
+    }
+    this.statusChannel_ = statusChannel;
+
+    // Wait until the Offscreen Document signals us about the module's
+    // loaded/disposed status (whichever comes first).
+    await Promise.all([this.waitLoadedMessage_(), this.waitDisposedMessage_()]);
```
Fabian-Sommer, line 165: The way I read the documentation, this wait on all promises to be resolved or one to be rejected. I think this should be Promise.race instead.

Fabian-Sommer (approved):
LGTM % nits

common/js/src/executable-module/offscreen-doc-emscripten-module.js
emaxx-google: Oh, a great catch!

emaxx-google:
Thanks for the reviews!

PR #1156, opened 2024-04-19
Fix flakiness in UserPromptingChecker

emaxx-google:
Fix two errors flakily observed in unit tests:

* "Cannot read properties of undefined" when trying to call into chrome.app.window.get().
  * Solution: we shouldn't call this function in unit tests, hence stub it out.
* "Unhandled promise rejection" inside PolicyOrPromptingPermissionsChecker.
  * Solution: suppress these errors since they aren't useful. This happens in the checkByUserPromptingChecker_() method where we create a child promise that's intentionally not returned anywhere.


annklr (approved):
Thank you for fixing this!

PR #1157, opened 2024-04-19
[mv3] Use Fetcher API in TrustedClientsRegistry

emaxx-google:
Switch the JSON fetching code from goog.net.XhrIo, which essentially is XMLHttpRequest, to the Fetcher API.

Not only is the Fetcher API more concise and promise-friendly. The main reason is that XMLHttpRequest is unsupported in Service Worker contexts, hence this was one of the blockers for the manifestv3 migration.


third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/trusted-clients-registry.js
```diff
@@ -143,163 +135,121 @@ TrustedClientsRegistryImpl.prototype.logger =
     GSC.Logging.getScopedLogger('PcscLiteServer.TrustedClientsRegistry');
 
 /** @override */
-TrustedClientsRegistryImpl.prototype.getByOrigin = function(origin) {
```
emaxx-google, line 146: Here and below is the main change: switching from `then` to `await` and from `xhrio` to `fetch`.

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/policy-or-prompting-permissions-checker-unittest.js
```diff
@@ -58,8 +57,8 @@ function FakeTrustedClientsRegistry() {}
 /** @override */
 FakeTrustedClientsRegistry.prototype.getByOrigin = function(origin) {
   if (origin === CLIENT_ORIGIN)
-    return goog.Promise.resolve(TRUSTED_CLIENT_INFO);
```
emaxx-google, line 61: Note: The PR is unfortunately a bit sizy because of mechanical changes like these - switching from Closure Library's (deprecated) Promises to the standard Ecmascript ones.

This is a viral effect of switching to `fetch` which is standard-promise-based and required the code chunks around it to use the same.

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/trusted-clients-registry-unittest.js
AleksMat, line 1: I think you can now remove some imports, such as:

```
goog.require('goog.Thenable');
goog.require('goog.testing.net.XhrIo');
```

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/policy-or-prompting-permissions-checker-unittest.js
emaxx-google: Yes, eventually I'm planning to switch everything to standard Promises. The current PR is a bit special in that doing both changes simulatenously was much simpler (because the `xhrio` interface used legacy promises), but generally it should be doable gradually.

emaxx-google:
Thanks! PTAL. I've also dropped the `json.js` helper since it's not needed anymore with the Fetch API.

third_party/pcsc-lite/naclport/server_clients_management/src/permissions_checking/trusted-clients-registry.js
```diff
@@ -54,10 +49,10 @@ const AUTOAPPROVE_FIELD = 'autoapprove';
 const GSC = GoogleSmartCard;
 
 /**
- * Test-only override of `goog.net.XhrIo.send()`.
- * @type {function(string, function(this: goog.net.XhrIo))|null}
+ * Test-only override of the `fetch()` API.
+ * @type {function(string):!Promise<!Response>|null}
```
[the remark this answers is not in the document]
emaxx-google: Ah I got confused by this myself as well - this `|null|` is actually referring to the whole `function`. So it's needed, but I've made it clearer by enclosing the `function` part into brackets.

AleksMat (approved):
Thanks, looks good 👍

PR #1160, opened 2024-04-22
[mv3] Fix messages to Emscripten in Offscreen Doc

emaxx-google:
Create the waiter for the messaging channel before opening the Offscreen Document, so that we don't miss the chrome.runtime.onConnect event.


emaxx-google:
A quick bugfix for #1141.

Fabian-Sommer (approved):

PR #1164, opened 2024-05-14
[pcsc] Retry init on more errors, try USB reset

emaxx-google:
Change the init retry mechanism to be triggered on any reader initialization error, as opposed to only doing it on openDevice/claimInterface errors. Additionally, attempt resetting the USB device after a few unsuccessful retries.

This should fix occasional flaky errors we see in the field, when there's a temporary conflict between the in-session and the login-screen instances of Smart Card Connector, which may cause failures at various USB functions and accidentally receiving a reply for a request sent by a different extension ("Invalid frame detected").


emaxx-google:
> One of the tests failed - seems like a real problem with composite devices. Looking into it.

Fixed now.

Fabian-Sommer (approved):

PR #1166, opened 2024-05-15
Refactor message awaiting

emaxx-google:
Create a small library function for for waiting on a specific incoming message. Cover it with unit tests.

This code will be reused for implementing the communication with popup dialogs in the manifest v3 mode.


AleksMat (approved):

emaxx-google:
I realized I need to add `await` when calling `assertRejects` in tests, so that we don't leave the test body before the assertion has been checked.

PR #1167, opened 2024-05-15
[mv3] Support popup windows - via message channels

emaxx-google:
This adds support of manifest v3 to the "popup-window" helpers. This is achieved by reimplementing the popup==>background page communication using message ports.

Previously, we used two other mechanisms for this communication:

* in the Chrome App build mode we injected callbacks into the popup's global state using the chrome.app.window.create() callback.
* in the Chrome manifest v2 Extension build mode we used "setSelfAsOpener" to let the popup call global functions in the opener page itself.

Neither works in manifest v3 Extensions, hence this commit implements a new approach that should work in all build modes.

### DIFF
--- a/common/js/src/popup-window/popup-opener.js
+++ b/common/js/src/popup-window/popup-opener.js
@@ -232,7 +232,7 @@ async function waitForPopupResult(popupId) {
       port, GSC.PopupConstants.RESULT_MESSAGE_TYPE);
 
   // Extract the result: either an error (needs to be thrown) or a value.
-  if (result.hasOwn(GSC.PopupConstants.RESULT_ERROR_KEY)) {
+  if (Object.hasOwn(result, GSC.PopupConstants.RESULT_ERROR_KEY)) {
     throw result[GSC.PopupConstants.RESULT_ERROR_KEY];
   }
   return result[GSC.PopupConstants.RESULT_VALUE_KEY];
--- a/common/js/src/popup-window/popup-opener.js
+++ b/common/js/src/popup-window/popup-opener.js
@@ -24,7 +24,10 @@ goog.provide('GoogleSmartCard.PopupOpener');
 
 goog.require('GoogleSmartCard.DebugDump');
 goog.require('GoogleSmartCard.Logging');
+goog.require('GoogleSmartCard.MessageWaiter');
 goog.require('GoogleSmartCard.Packaging');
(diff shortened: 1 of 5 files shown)


common/js/src/popup-window/popup-opener.js
```diff
@@ -147,50 +145,35 @@ GSC.PopupOpener.createWindow = function(url, windowOptions, opt_data) {
...
-  if (GSC.Packaging.MODE === GSC.Packaging.Mode.EXTENSION) {
-    goog.global[`googleSmartCard_resolveModalDialog${lastUsedPopupId}`] =
-        promiseResolver.resolve;
-    goog.global[`googleSmartCard_rejectModalDialog${lastUsedPopupId}`] =
-        promiseResolver.reject;
-  }
-
+  // Additionally pass the auto-generated popup ID. The popup will use it to
+  // talk back to us when sending the result.
+  ++lastUsedPopupId;
```
AleksMat, line 158: Nit: Isn't `lastUsedPopupId++;` more readable? I also don't see any difference in functionality.

AleksMat (approved):
LGTM % nits

common/js/src/popup-window/popup-opener.js
```diff
@@ -24,7 +24,10 @@ goog.provide('GoogleSmartCard.PopupOpener');
 
 goog.require('GoogleSmartCard.DebugDump');
 goog.require('GoogleSmartCard.Logging');
+goog.require('GoogleSmartCard.MessageWaiter');
 goog.require('GoogleSmartCard.Packaging');
+goog.require('GoogleSmartCard.PopupConstants');
+goog.require('GoogleSmartCard.PortMessageChannelWaiter');
 goog.require('goog.Promise');
 goog.require('goog.Thenable');
```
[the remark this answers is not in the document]
emaxx-google: good catch, done

common/js/src/popup-window/popup-opener.js
emaxx-google: done (I'm personally a bit more inclined to the preincrement version by default as it has simpler semantics than postincrement, but there's obviously no strong argument for/against in simple cases like this)

emaxx-google (commented):
thanks!

emaxx-google:
I've guarded some code in `in-popup-main-script.js` with checks because when the window is opened via `createWindow()` there's no `popup_id` passed and there's no need in one.

PR #1171, opened 2024-05-28
[mv3] Support popup windows - via message channels

emaxx-google:
This adds support of manifest v3 to the "popup-window" helpers. This is
achieved by reimplementing the popup==>background page communication
using message ports.

Previously, we used two other mechanisms for this communication:

* in the Chrome App build mode we injected callbacks into the popup's
  global state using the chrome.app.window.create() callback.
* in the Chrome manifest v2 Extension build mode we used
  "setSelfAsOpener" to let the popup call global functions in the
  opener page itself.

Neither works in manifest v3 Extensions, hence this commit implements a
new approach that should work in all build modes.

### DIFF
--- a/common/js/src/messaging/port-message-channel.js
+++ b/common/js/src/messaging/port-message-channel.js
@@ -74,6 +74,7 @@ GSC.PortMessageChannel = class extends goog.messaging.AbstractChannel {
 
     /**
      * @type {!goog.log.Logger}
+     * @public
      * @const
      */
     this.logger = GSC.Logging.getScopedLogger(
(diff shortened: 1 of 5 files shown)


emaxx-google:
@AleksMat This was already reviewed on #1167, but I screwed up merging and it ended up in a branch instead of `main`. Apologies :) Asking for a re-stamp.

AleksMat (approved):
I confirm the code is the same as in the previous PR.

emaxx-google:
Thank you again :)

PR #1181, opened 2024-07-17
[mv3] Support logs export

emaxx-google:
Make the log export UI mv3-compatible, by switching it from directly inspecting the background page's log buffer to using message passing.


AleksMat (approved):

emaxx-google:
I've pushed another commit that shouldn't make functional changes, but seems to greatly improve RAM consumption in practice (~3 MB of overhead instead of ~30 MB for the typical 2'000 log entries). Seems that V8's garbage collection is very lazy :)

PR #1182, opened 2024-09-10
[wasm] Update Emscripten to 3.1.67

emaxx-google:


Fabian-Sommer (approved):

emaxx-google:
I bumped it further to 3.1.67 - this seems backwards-compatible and they made a couple more bugfixes.

PR #1183, opened 2024-09-11
[wasm] Switch App builds from NaCl to Wasm

emaxx-google:
Stop building via Native Client in CI and in local builds via `make-all.sh`.


Fabian-Sommer (approved):
LGTM

PR #1184, opened 2024-09-11
[js] Update Closure Compiler to 20240317

emaxx-google:
We don't expect any specific improvement, but upstream fixed a few bugs.


Fabian-Sommer (approved):
LGTM

PR #1192, opened 2024-10-15
[libusb] Implement libusb_open_device_with_vid_pid

emaxx-google:
This libusb function will be needed for new drivers we're going to add.


Fabian-Sommer (approved):
LGTM

PR #1201, opened 2024-10-23
[pcsc] Let drivers export their methods themselves

emaxx-google:
Add an abstraction that represents a driver, and change PC/SC-Lite to call it instead of directly referencing CCID's functions.

This is a refactoring that prepares us for introducing multiple drivers in the application. Note that on *nix PC/SC-Lite achieves the same by dynamically loading each driver's .so file - something we cannot do in the webport.

### DIFF
--- a/third_party/ccid/webport/src/ccid_pcsc_driver_adaptor.cc
+++ b/third_party/ccid/webport/src/ccid_pcsc_driver_adaptor.cc
@@ -1,0 +1,70 @@
+// Copyright 2024 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


third_party/ccid/webport/src/ccid_pcsc_driver_adaptor.cc
```diff
@@ -0,0 +1,70 @@
...
+
+// Constructed as the concatenation of:
+// * path where the Info.plist config is installed at
+//   //smart_card_connector_app/build/executable_module/Makefile;
+// * the "Linux" string;
+// * the file name passed via "--target" to create_Info_plist.pl at
+//   ../build/Makefile.
+constexpr char kDriverFilePath[] =
+    "executable-module-filesystem/pcsc/drivers/ifd-ccid.bundle/Contents/Linux/"
+    "libccid.so";
```
Fabian-Sommer, line 36: optional nit: If these depend on values passed from the Makefile, maybe they should actually be passed from the Makefile via definitions to ensure that the values match.

Fabian-Sommer (approved):
LGTM

third_party/ccid/webport/src/ccid_pcsc_driver_adaptor.cc
emaxx-google: Makes sense - I've created #1204 to centralize these constants in the makefiles first of all.

emaxx-google: Landing the current PR, but #1204 will address your request. Thanks.

PR #1202, opened 2024-10-23
[ccid] Rename all exported IFDH functions

emaxx-google:
Append "CCID_" to all names of the IFDH functions in the CCID driver.

This is preparation for us introducing another driver, to avoid symbol clashes: every PC/SC-Lite driver exports the same set of IFDH functions, and we link everything statically in our webport (unlike it's done with separate .so libraries on *nix systems).

### DIFF
--- a/third_party/ccid/webport/build/Makefile
+++ b/third_party/ccid/webport/build/Makefile
@@ -39,6 +39,7 @@ CCID_NACL_SOURCES_PATH := ../src
 
 # Rules for invoking the source files compilation into object files
 
+# * rename all IFDH...() functions to avoid symbol clashes with other drivers;
 # * HAVE_PTHREAD definition enables the support of working with multiple readers
 #   simultaneously;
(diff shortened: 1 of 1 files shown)


third_party/ccid/webport/build/Makefile
```diff
@@ -47,6 +48,16 @@ CCID_NACL_SOURCES_PATH := ../src
 #   for all PC/SC-Lite server drivers; use a relative path, so that it works
 #   both inside the Smart Card Connector app and in unit tests;
 COMMON_CPPFLAGS := \
+	-DIFDHCloseChannel=CCID_IFDHCloseChannel \
```
emaxx-google, line 51: To note, I was looking for a more elegant way of achieving this, and the `objcopy --redefine-sym` seemed promising, however Wasm turned out to be unsupported in it (https://github.com/llvm/llvm-project/issues/50623)...

Fabian-Sommer (approved):

PR #1203, opened 2024-10-24
[ccid] Rename all exported IFDH functions

emaxx-google:
Append "CCID_" to all names of the IFDH functions in the CCID driver.

This is preparation for us introducing another driver, to avoid symbol clashes: every PC/SC-Lite driver exports the same set of IFDH functions, and we link everything statically in our webport (unlike it's done with separate .so libraries on *nix systems).


emaxx-google:
A revive of #1202 because I merged that one into a wrong place...

Fabian-Sommer (approved):

PR #1228, opened 2024-10-30
[c++] Drop std::optional polyfill

emaxx-google:
Now that we switched beyond C++14, we can use the standard std::optional instead of our hardwritten helper.


emaxx-google:
This is mostly a mechanical search-and-replace, except:

* Adding some previously missing includes - otherwise the code stopped compiling after I removed the inclusion of our `optional.h` in a couple of places.
* Use GMock's `Eq(std::nullopt)` instead of our handwritten helper.
* Drop some explicit conversions in a couple of places - `std::optional`s constructor and assignment operator are a bit more flexible that our polyfill's one.
* Drop some unneeded `std::move()` for the same reason.

Fabian-Sommer (approved):

PR #1242, opened 2024-11-06
[libusb] Drop check for pre-ChromeOS-51

emaxx-google:
The Device.version field has been provided by chrome.usb since ChromeOS 51 - we don't need to check for the field presence.

The minimum_chrome_version field in manifest.json guarantees that we'll never run on such ancient versions.


emaxx-google:
Hehe, it's been a while since ChromeOS 51 was released and I added the compatibility code!

PR #1251, opened 2024-11-07
[ccid] Update to 1.6.1

emaxx-google:


emaxx-google:
Only the second (smaller) patch is to be reviewed. The first is an import of 3p code as-is.

Fabian-Sommer (approved):

PR #1264, opened 2024-11-08
[PC/SC test] Test reconnect with reset & unpower

emaxx-google:
Test scenarios for SCardReconnect() for resetting, unpowering the card.

### DIFF
--- a/smart_card_connector_app/src/application_unittest.cc
+++ b/smart_card_connector_app/src/application_unittest.cc
@@ -2466,6 +2466,37 @@ TEST_F(SmartCardConnectorApplicationSingleClientTest,
   EXPECT_EQ(new_active_protocol, active_protocol);
 }
 
+// `SCardReconnect()` call from JS succeeds when using the same parameters as
+// the previous `SCardConnect()` call.
+TEST_F(SmartCardConnectorApplicationConnectedReaderTest, SCardReconnectReset) {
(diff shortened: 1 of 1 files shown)


smart_card_connector_app/src/application_unittest.cc
```diff
@@ -2466,6 +2466,37 @@ TEST_F(SmartCardConnectorApplicationSingleClientTest,
...
+  LONG return_code = SimulateReconnectCallFromJsClient(
+      kFakeHandlerId, scard_handle(), SCARD_SHARE_SHARED,
+      /*preferred_protocols=*/SCARD_PROTOCOL_T1, SCARD_RESET_CARD,
+      new_active_protocol);
+
+  EXPECT_EQ(return_code, SCARD_S_SUCCESS);
+  EXPECT_EQ(new_active_protocol, static_cast<DWORD>(SCARD_PROTOCOL_T1));
+}
+
+// `SCardReconnect()` call from JS succeeds when using the same parameters as
```
Fabian-Sommer, line 2484: I think the two comments of the tests should not be identical, but call out the difference in the tests.

Fabian-Sommer (approved):
LGTM % nit

smart_card_connector_app/src/application_unittest.cc
emaxx-google: done, thanks for catching this copy-paste error

PR #1266, opened 2025-01-16
Add SHA256 tests to example extension

Fabian-Sommer:
Merging some leftover changes for MV3 into main.

Prevent building as app.
Add some more tests.

### DIFF
--- a/example_cpp_smart_card_client_app/build/Makefile
+++ b/example_cpp_smart_card_client_app/build/Makefile
@@ -30,11 +30,6 @@ include $(THIRD_PARTY_DIR_PATH)/pcsc-lite/webport/js_client/include.mk
 include $(THIRD_PARTY_DIR_PATH)/pcsc-lite/webport/cpp_client/include.mk
 
 
-ifeq (app,$(PACKAGING))
-$(error Building as Chrome App is unsupported. Please set the environment variable: PACKAGING=extension)
-endif
-
-
 SOURCES_PATH := ../src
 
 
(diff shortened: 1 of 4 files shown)


example_cpp_smart_card_client_app/build/Makefile
```diff
@@ -30,6 +30,11 @@ include $(THIRD_PARTY_DIR_PATH)/pcsc-lite/webport/js_client/include.mk
 include $(THIRD_PARTY_DIR_PATH)/pcsc-lite/webport/cpp_client/include.mk
 
 
+ifeq (app,$(PACKAGING))
```
emaxx-google, line 33: This has to be reverted I think - or otherwise the build scripts need to be improved to skip building examples in the "app" mode...

example_cpp_smart_card_client_app/src/background.js
```diff
@@ -213,4 +213,8 @@ if (GSC.Packaging.MODE === GSC.Packaging.Mode.APP) {
 
 // Automatically load the App (in the background) with Chrome startup.
 GSC.AppUtils.enableSelfAutoLoading();
+
+// Trigger the PC/SC test.
+executableModule.getMessageChannel().send(
```
emaxx-google, line 218: This change is orthogonal to the PR - it's not about the SHA256. And actually, it's not important to be merged into `main` either. (It was just a quick hack to check whether PC/SC-Lite calls are still possible from the examples built the `PACKAGING=extension` mode.)

emaxx-google (approved):
lgtm % removing a couple of things

PR #1274, opened 2025-07-08
Ensure ArrayBuffer is the type returned to Chrome private API

zgroza:
This remediates the side effect of the extension migration and makes the Web Smart Card API work with the extension version of the Connector.

### DIFF
--- a/smart_card_connector_app/src/chrome-api-provider.js
+++ b/smart_card_connector_app/src/chrome-api-provider.js
@@ -41,6 +41,23 @@ const logger = GSC.Logging.getScopedLogger('ConnectorApp.ChromeApiProvider');
 const STATE_FLAGS_MASK = 0x0000FFFF;
 
 /**
+ * Returns the data in form of ArrayBuffer. This is needed because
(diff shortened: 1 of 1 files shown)


smart_card_connector_app/src/chrome-api-provider.js
```diff
@@ -40,6 +40,52 @@ const logger = GSC.Logging.getScopedLogger('ConnectorApp.ChromeApiProvider');
...
+ *     passed.
+ */
+function convertToArrayBuffer(data) {
+  if (Array.isArray(data)) {
+    return new Uint8Array(data).buffer;
+  } else if (data instanceof ArrayBuffer) {
+    return data;
+  } else {
+    throw new Error(`Unknown data type received from PCSC. " +
+      "Data type: ${typeof data}, data: ${JSON.stringify(data)}.`);
```
Fabian-Sommer, line 85: While we don't expect this to happen, I don't like that we print (possibly sensitive) data here. I think the data type should be good enough to fix this should it happen.

Fabian-Sommer (approved):

PR #1286, opened 2026-02-09
Migrate eslint config

Fabian-Sommer:
With the recent release of eslint v10.0, the automatic linter test on Github no longer accepts the old (deprecated) config file.

Migrate to the new config.mjs standard.


Fabian-Sommer:
The new linter detected some issues in the code, we can fix these in a follow-up CL.

